PR #5, opened 2015-05-29
Some cleanups

rumpelsepp:
This pr contains two patches:

* Cleanup conf.py
* Remove the font size adjustments

The first one cleans up the config file and reorders the config values. It also enables `html_last_updated_fmt`. The second patch removes the font size adjustments. These adjustments cause several problems, because they only change some little parts of the theme. Those little changes break the overal experience. The commit descriptions contains more details.

```
 _static/syncthing.css | 15 ---------------
 conf.py               | 30 +++++++++++++-----------------
```

### DIFF
--- a/_static/syncthing.css
+++ b/_static/syncthing.css
@@ -1,18 +1,3 @@
-.rst-content p, .rst-content p.admonition-title, .rst-content li, .rst-content table.docutils td,
-.rst-content table.docutils th {
-	font-size: 14pt;
-	line-height: 1.4;
-}
-.rst-content code, .rst-content div[class^='highlight'] pre {
-	font-size: 12pt;
-	line-height: 1.4;
-}
-.rst-content li {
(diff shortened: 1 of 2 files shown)


calmh:
I strongly disagree on the font size; many modern web sites have a base size of 18-19 px for better readability. Although I can buy that the problem you mention exist and that this should be solvable in a better way. Probably by modifying the source SASS of the theme.

My problem with the width is mostly that the sudden snap to the left that happens when the window gets wide looks like shit. With a maximized browser on my screen the theme takes up the leftmost 40% or so of the browser; the rest is a featureless greyness. If the whole thing got centered that'd be cool...

https://nym.se/t/wider.mov

calmh:
Merging, will try to med the theme instead.

rumpelsepp:
:+1:

calmh:
OK I hacked it horribly, but I think with better results than previously...

PR #4, opened 2015-05-29
Create syncthing-config(5)

rumpelsepp:
This pr converts users/config.rst to a properly formatted manpage

PR #6, opened 2015-05-29
Create syncthing-stignore(5)

rumpelsepp:
This patch generates syncthing-stignore(5). There is one patch missing, which enables the manpage in the configuration. Due to conflicts I can only include that missing patch when #4 has been merged.


rumpelsepp:
I'll have some more time right now, so I will create one manpage pull request, sorry. :)

PR #9, opened 2015-06-01
Add Gentoo Linux instructions

timboudreau:
I've created a Gentoo ebuild, so syncthing can be installed on Gentoo Linux (builds syncthing from source, which is the Gentoo way).

This patch adds Gentoo instructions to the contrib page.


calmh:
:+1:

PR #11, opened 2015-06-04
Explain CSRF tokens & cookies

Zillode:


users/faq.rst
```diff
@@ -216,6 +216,17 @@ The CSRF tokens are stored using cookies. Therefore, if you get the message
 ``Syncthing seems to be experiencing a problem processing your request``, you
 should verify the cookie settings of your browser.
 
+How to resolve "Syncthing seems to be experiencing a problem processing your request"?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The Syncthing GUI uses cookies to maintain a session. You can't have more than one session open against the same website in the same browser, much like you can't log in as two different users on this forum at the same time from the same browser, so that may cause this error message.
```
calmh: This is true, but lacks context here. When is it that one might have two sessions to the same site? (When browsing both localhost:8384 and localhost:8888 for example, but this needs explaining)

PR #25, opened 2015-07-09
Fix REST discovery API documentation

lfam:
Replace reference to nonexistent "hint" URI for POSTing to the discovery
cache with correct reference:
/rest/system/discovery/hint -> /rest/system/discovery


AudriusButkevicius:
Thanks.

PR #28, opened 2015-07-21
Small fixes to config.rst

Canisaur:
Fixed IPv6 bracket description ([] are square brackets, not angle brackets).  Also added a ref from the versioning object to the page describing versioning options (I'm almost positive the referencing is done correctly... don't have a way to preview it set up).

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -152,7 +152,7 @@ device
 versioning
     Specifies a versioning configuration.
 
-    .. note:: Needs explanation.
+    seealso:: :ref:`versioning`
 
 copiers, pullers, hashers
     The number of copier, puller and hasher routines to use, or zero for the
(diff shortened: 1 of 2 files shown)


users/config.rst
```diff
@@ -150,9 +150,7 @@ device
     the configuration file.
 
 versioning
-    Specifies a versioning configuration.
-
-    .. note:: Needs explanation.
+    See :ref:`versioning`
```
calmh: I'd leave the first sentence and the note in there, while adding this as a `seealso`. The referenced page describes versioning, but not its structure in the config xml

calmh:
Otherwise :+1: :)

Canisaur:
Will this work?  Thanks for the feedback!

calmh:
Thanks!

Canisaur:
Argh sorry, the seealso wasn't formatted properly...  I'm still figuring out this format.  #30 should fix it.

PR #33, opened 2015-07-27
Corrected mistype of "have"

cron410:


AudriusButkevicius:
Thanks

cron410:
Thanks for replying so fast

PR #35, opened 2015-08-04
Updated with field information for /rest/db/file

jorgen-k:
See forum: https://forum.syncthing.net/t/how-to-know-that-a-file-is-in-sync-using-rest-api/5313/2?u=jorgen_k

### DIFF
--- a/rest/db-file-get.rst
+++ b/rest/db-file-get.rst
@@ -33,3 +33,44 @@ availability. Takes ``folder`` and ``file`` parameters.
         ]
       }
     }
+
+Fields
+------
+
+
+The *name* is the file name path relative to the folder root (UTF-8 NFC). The name field uses the slash character ("/") as path separator, regardless of the operating system conventions. The combination of Folder and Name uniquely identifies each file in a cluster.
+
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
So the fallback is not hash of the file but the id of the device (so that everyone agress on the resolution order)

AudriusButkevicius:
Also perhaps worth metioning delete vs non-delete conflict. 

Btw mtime is only a sensible comparison insystems that don't modify it and are synced up. Syncthing only uses mtime for detecting changes and deciding a conflict. Newer mtime does not mean that that file is newer.

jorgen-k:
About the hash, I think I have written that in the pull request, but you porbably just pointed out that my first thought in the forum was incorrect, right?

Do I need to update anything in repsect to the delete vs non-delete conflict? If so, do you have a suggestion on how to phrase it?

I am releived about your comment about mtime :-)

AudriusButkevicius:
Also, numBlocks is number of blocks the file contains, version vector contains a list of device ids and their localVersion of the file, not version of blocks. Blocks don't have versions. Regarding deletes you should just say that non-deletion wins against deletion in case of conflict.

PR #37, opened 2015-08-08
Update Command Line Operation: Synopsis & Options

raferobinson:
Added -no-console to Synopsis
Added -no-console to Options


calmh:
Nice. Please add a note that this is a Windows-only option?

raferobinson:
Added (On Windows only) to -no-console Option.

calmh:
:+1:

PR #45, opened 2015-09-02
Update ignoring.rst

simplypeachy:
Explicitly specify that #include'd files must be within the repo. Note about Mac/Windows always being case-insensitive.

### DIFF
--- a/users/ignoring.rst
+++ b/users/ignoring.rst
@@ -44,7 +44,7 @@ The ``.stignore`` file contains a list of file or path patterns. The
    ``/foo`` matches ``foo`` but not ``subdir/foo``.
 
 -  A pattern beginning with ``#include`` results in loading patterns
-   from the named file. It is an error for a file to not exist or be
+   from the named file which must reside in the repository. It is an error for a file to not exist or be
    included more than once. Note that while this can be used to include
    patterns from a file in a subdirectory, the patterns themselves are
    still relative to the repository *root*. Example:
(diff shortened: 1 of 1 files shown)


users/ignoring.rst
```diff
@@ -44,7 +44,7 @@ The ``.stignore`` file contains a list of file or path patterns. The
    ``/foo`` matches ``foo`` but not ``subdir/foo``.
 
 -  A pattern beginning with ``#include`` results in loading patterns
-   from the named file. It is an error for a file to not exist or be
+   from the named file which must reside in the repository. It is an error for a file to not exist or be
```
calmh: This was actually not the case, it's just that the name is always relative to the folder root, no? So I think you can reach c:\ignores.txt from c:\stuff\sync by `#include ../../ignores.txt`, no?

simplypeachy: Just tested - you're correct. Should this behaviour be considered a bug?

calmh:
Possibly. We mentioned it in passing today and it seems like a nonissue for now...

```
18:50:00 <@Zillode> but you can escape to ../
19:25:34 <@calmh> Zillode: hmm. might that be a security issue?
19:25:59 <@calmh> not sure how though
19:32:01 <@Zillode> calmh: nah, I thought about removing that option but so far I could not see a use case to leak information
19:32:12 <@calmh> me neither
```

i.e. unclear how it could be used for badness, possibility of it actually being useful

calmh:
Thanks. I forgot about this one. :(

PR #46, opened 2015-09-03
Fix StartupComplete event documentation title

majedev:
Event title is StartupCompleted, while actual event type is StartupComplete (without d). Should be consistent with event type.


calmh:
:+1:

PR #48, opened 2015-09-07
Added the newly added -logfile option

cbhushan:


rumpelsepp:
Never mind; I'll fix that :)

PR #51, opened 2015-09-15
Update firewall.rst

MarkSHarrison:
Local Firewall UDP ports should read 22025 and 22026. This has been confirmed across several of my local installs.
Adjö

### DIFF
--- a/users/firewall.rst
+++ b/users/firewall.rst
@@ -30,8 +30,8 @@ incoming traffic:
 
 -  Port **22000/TCP** (or the actual listening port if you have changed
    the *Sync Protocol Listen Address* setting.)
--  Port **21025/UDP** (for discovery broadcasts on IPv4)
--  Port **21026/UDP** (for discovery multicasts on IPv6)
+-  Port **22025/UDP** (for discovery broadcasts on IPv4)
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
https://github.com/syncthing/syncthing/blob/v0.11.25/lib/config/config.go#L219 says otherwise

MarkSHarrison:
I see what you are saying, but then why is the "Global Discovery Server" property under the GUI Settings set to "udp4://announce.syncthing.net:22026, udp6://announce-v6.syncthing.net:22026"? All my syncs stopped when I only opened UDP 21025 and 21026 ports on my install local software firewalls? Once I modified the firewall rule to allow UDP 22025 and 22026 everything worked like a champ.

MarkSHarrison:
After looking at my edge firewall, it seems that UDP:22026 outbound was being blocked. Hence Syncthing installs could not talk to the Global Discovery Server. It seems that the firewall documentation should state that outbound UDP:22026 needs to be opened if outbound traffic is being restricted on a firewall.

AudriusButkevicius:
True, we can add that. Mind ammending your PR or making a new one?

calmh:
(added)

PR #57, opened 2015-09-23
Move dead or dying projects to the bottom.

calmh:
These are things not updated in the last six months and which don't look
like they are still viable. I don't want to kill them outright, but I
don't want to point users to them either. Other opinions?

### DIFF
--- a/users/contrib.rst
+++ b/users/contrib.rst
@@ -17,27 +17,16 @@ Cross-platform
 
 -  https://github.com/syncthing/syncthing-gtk
 -  https://github.com/sieren/QSyncthingTray
--  https://github.com/alex2108/syncthing-tray (Tray icon only)
 
 Android
 ~~~~~~~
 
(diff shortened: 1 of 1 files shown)


users/contrib.rst
```diff
@@ -88,33 +76,11 @@ Docker
...
-   packages coming)
-
--  https://copr.fedoraproject.org/coprs/tune2fs/syncthing/ (RPMs, based
-   on https://github.com/thunderbirdtr/syncthing\_rpm)
 
 Gentoo
 ~~~~~~
 
-Syncthing is available from the [timboudreau overlay](http://github.com/timboudreau/gentoo), or from [fearedbliss's overlay](https://github.com/fearedbliss/bliss-overlay). These overlays contain the ebuild for building syncthing from source,
-installing it, and setting it up as a service that runs on startup.
```
calmh: Gentoo users will have to know how to handle their OS, I don't want a tutorial in here.

PR #62, opened 2015-10-11
Clarify things which confused me my first time trying to set up autostart

terrycloth:
I haven't worked much with reStructuredText before, so hopefully I didn't mess up the markup.

A couple things I was unsure about:

The systemd section of the Autostart article is quite a bit longer than the other sections, partly because it includes extra subsections about things like how to handle the Ignore Permissions option in the Syncthing client's folder settings. Do the other methods of setting up autostart on Linux not need special configs related to umask permissions?

Also, I wonder about the references to file paths in this article, or elsewhere in the documentation. It seems that, as there get to be more ways of installing Syncthing from the distributions' package managers, that there will be some variation

Other than that, I try to explain/justify my changes in the commit message.

### DIFF
--- a/users/autostart.rst
+++ b/users/autostart.rst
@@ -138,21 +138,20 @@ access the GUI type 127.0.0.1:8384 (by default) into Safari.
 Linux
 -----
 
-Ubuntu like systems
-~~~~~~~~~~~~~~~~~~~
+On Ubuntu-like systems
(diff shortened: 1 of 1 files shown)


rumpelsepp:
Since I have created that mess with systemd, I will review it within the next few days. :)

rumpelsepp:
:+1:

PR #64, opened 2015-10-19
Add information about syncing config files safely

burkemw3:
A first pass at docs for me. Writing clearly about a distributed system is hard! Comments welcome!

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -514,3 +514,21 @@ minHomeDiskFreePct
 
 releasesURL
     The URL from which release information is loaded, for automatic upgrades.
+
+Syncing Configuration files
+---------------------------
+
+Syncing configuration files between devices (such that multiple devices are
+using the same configuration files) can cause issues. This is easy to do
+accidentally if you sync your home folder between devices. A common symptom
+of syncing configuration files is two devices ending up with the same Device ID.
+
+If you want to use syncthing to backup your configuration files, it is recommended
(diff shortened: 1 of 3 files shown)


users/config.rst
```diff
@@ -424,3 +424,21 @@ pingTimeoutS
...
+
+If you want to use syncthing to backup your configuration files, it is recommended
+that the files you are backing up are in a :ref:`folder-master` to prevent other
+devices from overwriting the per device configuration. The folder on the remote
+device(s) should not be used as configuration for the remote devices.
+
+If you'd like to sync your home folder in non-master mode, you may add the
+folder that stores the configuration files to the :ref:`ignore list <ignoring-files>`.
+If you'd also like to backup your configuration files, add another folder in
+master mode for just the configuration folder.
```
burkemw3: Will this setup of nesting syncthing folders actually work?

AudriusButkevicius: I don't think nesting is involved here.

burkemw3: I was thinking it'd end up something like this:

- Sync home folders across devices (e.g. /home/burkemw3/) and ignore .config/syncthing/**
- Sync config folders as back (e.g. /home/burkemw3/.config/syncthing/) as folder master on local and to a different path on remove devices

/home/burkemw3/.config/syncthing/ is nested in /home/burkemw3/

burkemw3:
The added section (Syncing Configuration files) is linked to in a warning message in the v0.12 release, so I think this should be merged so the section exists!

PR #68, opened 2015-11-04
REST and config file updates for v0.12

calmh:


users/config.rst
```diff
@@ -351,6 +406,12 @@ localAnnouncePort
 localAnnounceMCAddr
     The group address and port to join and send IPv6 multicast announcements on.
 
+relayServer
+    Lists one or more relay servers, on the format ``relay://hostname:port``.
+    Alternatively, a relay list can be loaded over https by using an URL like
+    ``https://somehost/path``. The default loads the list of relays from the
```
AudriusButkevicius: Dynamic+https

users/config.rst
```diff
@@ -149,6 +169,11 @@ device
     Syncthing will currently add this automatically if it is not present in
     the configuration file.
 
+minDiskFreePct
+    The amount of space that should be available on the disk this folder
```
phil-davis: "The percentage of space that..."
"amount of" implies that it is an amount like 100MB, 1GB... but the units of the name are in percent.
Also on the following line change another "amount" to "percentage"

users/config.rst
```diff
@@ -361,6 +422,17 @@ reconnectionIntervalS
     The number of seconds to wait between each attempt to connect to currently
     unconnected devices.
 
+relaysEnabled
+    When true, relays will be connected to and potentially used for device to device connections.
+
+relayReconnectIntervalM
+    Sets the interval, in minutes, between relay reconnect attempts.
+
+relayWithoutGlobalAnn|
```
phil-davis: Seems to be a bonus vertical bar at the end of this line.

calmh:
Fixd, thanks

PR #75, opened 2015-11-07
/rest/system/connections update for v0.12

alex2108:
Added the output of two remote devices configured: one connected and one disconnected. Maybe the disconnected device could also be left out, but I added it to make clear that disconnected and local device looks the same.


calmh:
:+1:

PR #78, opened 2015-11-14
Add hash algos as flags on the folder in cluster config

calmh:
Reserves a permanent spot among the folder flags for up to 16 different hash algorithms.


Zillode:
Hmm, do we need this per folder? I can imagine a scenario where blocks are pulled using a murmur hash between trusted devices, and a sha hash for untrusted devices (within the same shared folder).

calmh:
That needs a bigger protocol change then as we'd need to track multiple hashes per file. Note though that if an implementation wanted to do this, it could simply represent the two aspects of the folder as two separate folder ids.

PR #89, opened 2015-11-29
Add to the Relays page

canton7:
- Minimum recommended specifications
 - More detail on how to run a relay on port 443

### DIFF
--- a/users/relaying.rst
+++ b/users/relaying.rst
@@ -27,7 +27,30 @@ devices.
 Running Your Own Relay
 ----------------------
 
-To run a relay of your own, download the latest release of the `relay server <https://github.com/syncthing/relaysrv/releases>`__
+Syncthing relies on a network of community-contributed relay servers. Anyone can
+run a relay server, and it will automatically join the relay pool and be
+available to Syncthing users. The current list of relays can be found at
+https://relays.syncthing.net.
+
+To run a relay of your own, you will first need a server. Running a relay from a
+home computer is not recommended: the relay ideally needs to be running 24/7,
+and home internet connections often have poor upload bandwidth.
+
+The recommended specifications for a relay are:
+
+========= ==============
+CPU       At least 1GHz
+RAM       At least 256MB
+Bandwidth At least 1MBit/s in and out, the higher the better
+Traffic   At least 1TiB/month, the more the better
+========= ==============
+
+VPS's with these specs are available for about $5-$10/month.
+
+Installing and Running
+~~~~~~~~~~~~~~~~~~~~~~
+
+Download the latest release of the `relay server <https://github.com/syncthing/relaysrv/releases>`__
 for your operating system and architecture. Unpack the archive and save the
 binary to a convenient place such as `/usr/local/bin`.
 
@@ -38,6 +61,10 @@ smooth operation::
     Usage of relaysrv:
       -debug
             Enable debug output
+      -ext-address string
+            An optional address to advertising as being available on.
+            Allows listening on an unprivileged port with port forwarding from e.g.                                             
+            443, and be connected to on port 443.
       -global-rate int
             Global rate limit, in bytes/s
       -keys string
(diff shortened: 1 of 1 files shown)


users/relaying.rst
```diff
@@ -27,7 +27,30 @@ devices.
...
+and home internet connections often have poor upload bandwidth.
+
+The recommended specifications for a relay are:
+
+========= ==============
+CPU       At least 1GHz
+RAM       At least 256MB
+Bandwidth At least 1MBit/s in and out, the higher the better
+Traffic   At least 1TiB/month, the more the better
+========= ==============
```
AudriusButkevicius: How did you come up with these?

users/relaying.rst
```diff
@@ -27,7 +27,30 @@ devices.
...
 ----------------------
 
-To run a relay of your own, download the latest release of the `relay server <https://github.com/syncthing/relaysrv/releases>`__
+Syncthing relies on a network of community-contributed relay servers. Anyone can
+run a relay server, and it will automatically join the relay pool and be
+available to Syncthing users. The current list of relays can be found at
+https://relays.syncthing.net.
+
+To run a relay of your own, you will first need a server. Running a relay from a
+home computer is not recommended: the relay ideally needs to be running 24/7,
```
AudriusButkevicius: why the `:` ?

users/relaying.rst
```diff
@@ -66,10 +93,9 @@ smooth operation::
 Primarily, you need to decide on a directory to store the TLS key and
 certificate and a listen port. The default listen port of 22067 works, but for
 optimal compatibility a well known port for encrypted traffic such as 443 is
-recommended. This may require `additional setup
-<https://wiki.apache.org/httpd/NonRootPortBinding>`__ to work without running
-as root or a privileged user. In principle something similar to this should
-work on a Linux/Unix system::
+recommended. This may require additional setup to work without running
+as root or a privileged user, see below. In principle something similar to this
```
AudriusButkevicius: see the section `Running on port 443 as an unprivileged user` below.

users/relaying.rst
canton7: CPU and RAM were from looking at relaysrv's usage on my relays. Bandwidth was from the minimum rate limits used by current relays, and traffic is bandwidth + maths.

users/relaying.rst
canton7: The clause following the colon explains the stuff preceding it.

users/relaying.rst
canton7: Fair.

users/relaying.rst
```diff
@@ -82,3 +109,35 @@ global relay pool, unless a ``-pools=""`` argument is given.
...
+
+    *nat
+    :PREROUTING ACCEPT [0:0]
+    :POSTROUTING ACCEPT [0:0]
+    
+    -A PREROUTING -i eth0 -p tcp --dport 443 -j REDIRECT --to-port 22067
+    
+    COMMIT
+
+You will need to start ``relaysrv`` with ``-ext-address ":443"``. This tell
```
phil-davis: s/tell/tells/

users/relaying.rst
```diff
@@ -38,6 +61,10 @@ smooth operation::
     Usage of relaysrv:
       -debug
             Enable debug output
+      -ext-address string
+            An optional address to advertising as being available on.
+            Allows listening on an unprivileged port with port forwarding from e.g.                                             
+            443, and be connected to on port 443.
```
phil-davis: I guess this is an IP address plus port number. Perhaps:
An optional IP address and port number that is the real external address:port that the relay server can be contacted at. This might be a public IP address and port that is forwarded in to the relay server, or just a port that is being forwarded in the host itself. This also allows listening on an unprivileged port and forwarding port 443 to that unprivileged port, so user can still connect to port 443.

canton7: This is copied from the output of `relaysrv -help`. I was trying to keep it shortish to avoid spamming the console, but agreed that my wording isn't particularly clear.

phil-davis: I agree, it needs to be shorter than what I wrote above! But what exactly to write?

canton7: I'm not sure it's worth talking about an external IP. By default, the relay doesn't specify the IP it's located at, and the relay pool server looks at where the relay's registration request came from. 

How about...

An optional port that is the real external port of the relay server, e.g. a port which is forwarded by NAT. Use this and a firewall rule to have an unprivileged relaysrv listen on a privileged port.

phil-davis: That sounds good. And if you like, you could add/keep something like the example that is there already:
e.g. :443 and then forward external port 443 to the local port that relaysrv is using.

PR #92, opened 2015-12-10
Master mode: rephrasing how data flows.

devste:
I had to read the text about three times before realizing that it does actually describe the picture. Rephrased it to highlight the direction of data flow in master mode.


AudriusButkevicius:
Tyvm

PR #96, opened 2015-12-18
Add note about running syncthing instance on discovery server

kluppy:


calmh:
Cleaned up & merged, thanks.

PR #98, opened 2015-12-30
Support space in directory

jimmyjones2:
Currently the script breaks if versionspath or filepath contain a space, this fixes it.


calmh:
:+1:

PR #103, opened 2016-01-11
autostart.rst Mac OS X Updates

tuathail:
Some clarifications to the Mac OS X install including:
Spelled out the home directory (/home/<username>/bin) in step one. Some people might gloss over it and just put it in /bin otherwise.
4 replacements of username are required, not 2.
Replaced the ~ with /Users/<username> for consistency, but also if the user switched to root user to avoid it being copied to /var/root/Library.. instead of /Users/<username>/Library...


lkwg82:
bump

lkwg82:
bump

PR #107, opened 2016-01-17
Remove links

ojongerius:
They are either dead or not visible to non team members.


rumpelsepp:
I guess it is the latter.

calmh:
Huh. They used to work, but ah well.

PR #115, opened 2016-02-08
Fix device config re: prefixes

ProactiveServices:
Added tcp:// prefixes in docs and examples, added host name detail, fixes to IPv6 examples.


calmh:
Thanks!

PR #119, opened 2016-02-09
Add FAQ for syncing folders on same system

ProactiveServices:
This seems to be an FAQ so I propose adding it to the official list.


ProactiveServices:
Cheers, hadn't checked that after I fiddled with the heading.

PR #120, opened 2016-02-12
users/autostart.rst: Update path to *.plist file from Homebrew on Mac OS X

s1lent1um:


AudriusButkevicius:
No clue if this is right. Some apple users will have to verify and merge.

s1lent1um:
Actual path after running `brew install syncthing` is `/usr/local/Cellar/syncthing//0.12.17/homebrew.mxcl.syncthing.plist`
So I tried to make the command not dependent on specific version.

calmh:
I don't *think* this is correct. The homebrew formula is supposed to install the launchagent for the user, or at least provide instructions on doing so? (The existing part is probably not super correct either...)

PR #129, opened 2016-03-07
Windows does not support escaping (ref #2822)

ProactiveServices:


calmh:
Thanks. However, more importantly it needs to describe that brackets are actually interpreted as character ranges, as this is currently not mentioned anywhere :)

ProactiveServices:
I hadn't even noticed that was missing...too much regex in my life.

calmh:
@st-review merge this please

st-review:
OK, merged as cb62b52a3023b9849f3406dd3ef880658311fb4d. Thanks, @ProactiveServices!

PR #130, opened 2016-03-07
updated compiler version

lkwg82:
see compiler change to version 1.6 in https://github.com/syncthing/syncthing/releases/tag/v0.12.20

### DIFF
--- a/dev/building.rst
+++ b/dev/building.rst
@@ -23,7 +23,7 @@ may be unstable and unsuitable for general consumption.
 Prerequisites
 -------------
 
--  Go **1.3** or higher
+-  Go **1.6** or higher
 -  Git
 
 If you're not already a Go developer, the easiest way to get going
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Now that I am thinking about it, it would probably still work with 1.3 if you'd use our build scripts.

lkwg82:
what about the go1.5 vendors directory is this backward compatible?

AudriusButkevicius:
yeah 1.5 it works with a special variable set

calmh:
I reverted it. We *should* update this section a little as our build process now becomes a little different in 1.3-1.4, 1.5 and 1.6. However we *do* still build correctly on 1.3 and 1.4, the dependencies just need different handling. I'll write something up, thanks for bringing this to our attention.

calmh:
I pushed https://github.com/syncthing/docs/commit/7079b3f0c9257005722bd6eab929f42326de820c to clarify the situation.

PR #132, opened 2016-03-08
adjust path to theme path

lkwg82:


AudriusButkevicius:
Is that really the case? I thought that it's the default by default

lkwg82:
Try it! I changed as I experienced 404, when used documented param.

AudriusButkevicius:
```
Audrius@Audrius syncthing $ mkdir foo/
Audrius@Audrius syncthing $ echo bar >foo/index.html
Audrius@Audrius syncthing $ STGUIASSETS=foo/ ./bin/syncthing.exe &>/dev/null &
[1] 2984
Audrius@Audrius syncthing $ curl 127.0.0.1:8080
bar
```

lkwg82:
this is absolute correct, but index.html is not in ```gui```  it is in ```gui/default```

AudriusButkevicius:
Ok, this is confusing but I get it, we should explain that gui/default implies a path in the syncthing checkout/git clone/repo, as the way it's now to me implies that you are expected to put your index.html in something/default.

lkwg82:
What other paths would you start with, if not with the ones in the repo ?!.

AudriusButkevicius:
Anyways, this makes sense, I guess it confuses me just because I understand it well.

calmh:
No, this confuses me as well. If this change is correct I think we have a bug, or at least need to spend some time thinking of how the default + theme + override thing should really work.

lkwg82:
This problem also touches gui prototyping:

When STGUIASSETS is set and you will work on a theme which only has override a css (the dark theme is in this situation), you dont get it, because the corpus (html+js) is missing. Here you only can copy the whole theme and dismiss those parts not changed.

AudriusButkevicius:
@calmh define bug. You override assets of the current theme.

calmh:
In my mind, the bug is that STGUIASSETS should no longer point to the same hierachy that we compile in, and that the behavior is different when overriding vs when not. It used to be very simple (assuming $PATH has been requested):

1. `$STGUIASSETS/$PATH`
2. Compiled in asset `$PATH`

Now we're instead looking for

1. `$STGUIASSETS/$PATH`
2. Compiled in asset `$THEME/$PATH`
3. Compiled in asset `default/$PATH`

I think it would be cleaner if we instead looked for

1. `$STGUIASSETS/$THEME/$PATH`
2. `$STGUIASSETS/default/$PATH`
3. Compiled in asset `$THEME/$PATH`
4. Compiled in asset `default/$PATH`

Then we'd get back the old behavior of just being able to override the whole hierarchy and then hack on the dark theme or whatever as @lkwg82 also mentions. A (small) complication is that the theme selector would need to pick up themes present both in the compiled in assets and in the potential overridden assetsDir.

AudriusButkevicius:
Fine, lets revert this and open a ticket in syncthing.

PR #135, opened 2016-03-17
 Correct unit for rate limit (fixes syncthing/syncthing#2844)

wweich:
Correct unit for `maxSendKbps` and `maxRecvKbps` from kibibits to kibibytes.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -433,10 +433,10 @@ relayServer
     from the relay pool server, ``relays.syncthing.net``.
 
 maxSendKbps
-    Outgoing data rate limit, in kibibits per second.
+    Outgoing data rate limit, in kibibytes per second.
 
 maxRecvKbps
-    Incoming data rate limits, in kibibits per second.
+    Incoming data rate limits, in kibibytes per second.
 
 reconnectionIntervalS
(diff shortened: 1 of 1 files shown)


phil-davis:
I am firstly assuming that the OP is correct in saying that the units of measure are using bytes rather than bits.
Ideally the abbreviations would also use a capital B when referring to bytes. And in the case of "kilobyte" the abbreviation is supposed to be "kB" and "kibibyte" KiB.
So the names of the parameters could be written:
a) maxSendkBps and maxRecvkBps - if the parameter names are not case-sensitive when used in the config file, then this would require no downstream changes.
b) maxSendKiBps and maxRecvKiBps - but doing that requires changes to the underlying code that reads config file key/values.

calmh:
Agreed. They should be named as in option b. Changing the code is not an issue; changing the config format will probably annoy people though.

calmh:
And if we're anyway changing that, we might consider if this is the unit that makes the most sense or whether we should actually have the config in kilobits instead...

wweich:
I just tested option a). Syncthing ignored the setting, so it is in fact case sensitive and option b) is the way to go.

PR #136, opened 2016-03-21
Clarify remote states when ignoreDelete selected

kluppy:
As discussed on the forum: https://forum.syncthing.net/t/stuck-syncing-at-90/6991


calmh:
:+1:

(Although I'm very tempted to simplify this entire section to something like "This option is a hack on the update system. The in sync status will not make sense on either side when this option is set.")

PR #138, opened 2016-03-25
updated spec ...

lkwg82:
... according to syncthing/syncthing/pull/2668 (folder label)


specs/bep-v1.rst
```diff
@@ -327,6 +331,8 @@ Fields (Folder Structure)
 
 The **ID** field contains the folder ID, as a human readable string.
 
+The **Label** field contains the folder label, as human readable name for the **ID**.
```
AudriusButkevicius: The ID is already human readable, maybe human readable name for the folder?

PR #144, opened 2016-03-26
Fix docs for join session flow

vhbit:
`JoinSessionRequest` sounds more appropriate and actually works.


AudriusButkevicius:
Good spot, thanks.

PR #145, opened 2016-03-27
Add a Windows installer

sodacode:
Not sure why nobody has done this yet, maybe I'm missing something?
Wrote a small script for a Windows installer.


sodacode:
I need to close and update the script first.. It's installing to Documents and needs to install to appdata.

PR #147, opened 2016-04-04
Remove references to cacheIgnoredFiles

ProactiveServices:
Feature removed in PR #2930


calmh:
It's not actually removed though, just disabled by default. :)

ProactiveServices:
Thunderbird helpfully put most of your conversation in a new thread so I missed all of the messages pertaining to it not being deleted :-P

PR #149, opened 2016-04-07
Added documentation for running discosrv behind nginx reverse proxy

firecat53:
Added documentation for running the discovery server behind nginx reverse proxy as requested by @AudriusButkevicius. Hopefully this is what you were looking for. Please let me know if I can change anything (or made any dumb mistakes).

Scott

### DIFF
--- a/users/discosrv.rst
+++ b/users/discosrv.rst
@@ -101,48 +101,68 @@ IP address, you will need a dynamic DNS service.
 If you wish to use *only* your own discovery server, remove the ``default``
 entry from the list.
 
-Running the Discovery Server behind an Nginx Reverse Proxy
-----------------------------------------------------------
+Reverse Proxy Setup
+-------------------
 
-With syncthing/discosrv@4149fc1, the -http flag was added to the discovery
-server allowing use behind an http proxy. Using this feature, the discovery
-server can be run behind an SSL-secured Nginx reverse proxy. This also allows
-use of a subdomain name without requiring a port number added to the URL.
+The discovery server can be run behind an SSL-secured reverse proxy. This
+allows:
 
-Requirements:
+- Use of a subdomain name without requiring a port number added to the URL
+- Sharing an SSL certificate with multiple services on the same server
 
-- Discosrv version *at or after* the above commit.
+Requirements
+^^^^^^^^^^^^
 
-- Nginx configuration with the following extra directives to handoff the SSL
-  connection::
+- Run the discovery server using the -http flag  :code:`discosrv -http`.
+- SSL certificate/key configured for the reverse proxy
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
So I think this should be written to assume that -http is there (potentially updating the docs as it seems to be missing), and not mention all the commits and versions as it's just confusing as it doesn't say which commit my version is from etc.

I think there should be a `Reverse Proxy Setup` section explaining that you need to run with -http option, and pass X-
SSL-Cert header containing PEM encoded client certificate and X-Forwarded-For header containing clients real IP.

Then in the nginx subsection you could talk how to provide the 2 headers and force client to provide certificates and giving the example config of yours and pointing to the forum link.

Something simillar to https://docs.syncthing.net/users/reverseproxy.html, where it explains what it is a bit, and what it does and why it's useful/better.

Also, not sure why you think the UDP port forwarding is still required?

users/discosrv.rst
```diff
@@ -100,3 +100,67 @@ IP address, you will need a dynamic DNS service.
...
+use https://discovery.mydomain.com as the discovery server URL in the Syncthing
+settings::
+
+    upstream discovery.mydomain.com {
+        # Local IP address:port for discovery server
+        server 172.17.0.6:8443;
+    }
+    server {
+            server_name discovery.mydomain.com;
+            listen 80 ;
```
AudriusButkevicius: the extra space before ; on both listen lines is not needed.

users/discosrv.rst
```diff
@@ -100,3 +100,67 @@ IP address, you will need a dynamic DNS service.
...
+            ssl_session_cache shared:SSL:50m;
+            ssl_certificate /etc/nginx/certs/discovery.mydomain.com.crt;
+            ssl_certificate_key /etc/nginx/certs/discovery.mydomain.com.key;
+            ssl_dhparam /etc/nginx/certs/discovery.mydomain.com.dhparam.pem;
+            add_header Strict-Transport-Security "max-age=31536000";
+            ssl_verify_client optional_no_ca;
+            location / {
+                    proxy_pass http://discovery.mydomain.com;
+            }
+    }
```
AudriusButkevicius: This is actually missing both X-Forwarded-For and X-SSL-Cert, so I suspect this doesn't work.

firecat53:
Yipes...totally forgot to paste the proxy parameters...was in a hurry last night :(  I'll fix that up and work on the other suggestions. Thanks!

Scott

calmh:
@st-review merge please

users/discosrv: Document running discosrv behind nginx

st-review:
:ok_hand: Merged as 5ebce765249898169c66a226a686dc2b021400ad. Thanks, @firecat53!

PR #153, opened 2016-04-14
Explain basics of discovery server

ian-kelling:


calmh:
While this is somewhat true (that page does mention roughly what the discovery service does), I don't think it's the appropriate place to link to for an overviews. We should probably write an actual overview instead, somewhere.

PR #158, opened 2016-04-27
mark systemd second step as potentially redundant

DarkPark:


calmh:
This is already mentioned in the paragraph above the "how to set up a system service" header. However this entire sections should be trimmed down a bit as it's very wordy...

PR #159, opened 2016-04-29
Document HelloMessage

AudriusButkevicius:


specs/bep-v1.rst
```diff
@@ -60,10 +60,86 @@ as noted per message type - any message type may be sent at any time and
...
+*devices* MUST exchange Hello messages.
+
+Hello messages are used to carry additional information about the peer, which
+might be of interest to the user even if the peer is not permitted to
+communicate due to failing authentication.
+
+Hello messages MUST be prefixed with a magic number **0x9F79BC40**
+represented in network byte order (BE), followed by 4 bytes representing the
+size of the message in network byte order (BE), followed by the content of
+the Hello message itself. The content of Hello message MUST be less or equal
```
calmh: "The size of the contents of the Hello message ..."

specs/bep-v1.rst
```diff
@@ -60,10 +60,86 @@ as noted per message type - any message type may be sent at any time and
...
+
+::
+
+    struct HelloMessage {
+        string DeviceName<64>;
+        string ClientName<64>;
+        string ClientVersion<64>;
+    };
+
+
```
calmh: Do we add a note about the connection being unceremoniously dropped here, after sending the hello message, if we don't like the other device?

specs/bep-v1.rst
```diff
@@ -60,10 +60,86 @@ as noted per message type - any message type may be sent at any time and
...
+::
+
+    Prefix Structure:
+
+     0                   1                   2                   3
+     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |                             Magic                             |
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |                            Length                             |
```
calmh: I'm thinking here that wrapping the thing in a separate length field doesn't really buy us anything, given that we set limits like 64 bytes each on the strings later. Currently, the effective max length of the hello message is 3*64 bytes plus the length and magic fields.

calmh: I know the code is already running, but it's the way of things that they appear clearer when in the specification than in code :)

AudriusButkevicius: The way it is now, if we add extra fields, it's probably still protocol compatible with other versions, as the garbage that we don't use will be read off the socket anyway, whereas if we hardcode it at  4+3*64, the extra fields will be left for the protocol to read, and will potentially cause panics?

calmh: Indeed. Better as is.

calmh:
@st-review merge please

specs/bep-v1: Add Hello message, changed ClusterConfig

st-review:
:ok_hand: Merged as 8d245533a768769bf70bff8c861297baa1d92c45. Thanks, @AudriusButkevicius!

calmh:
Some issues the compiler complains about:

```
/Users/jb/src/github.com/syncthing/docs/specs/bep-v1.rst:224: SEVERE: Title level inconsistent:

Cluster Config (Type = 0)
~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jb/src/github.com/syncthing/docs/specs/bep-v1.rst:947: SEVERE: Title level inconsistent:

Trusted
~~~~~~~
/Users/jb/src/github.com/syncthing/docs/specs/bep-v1.rst:961: SEVERE: Title level inconsistent:

Read Only
~~~~~~~~~
```

I haven't looked but I guess this is level three headings under a level one or something.

AudriusButkevicius:
Fixed in the temp index PR.

PR #160, opened 2016-04-29
Document (?d) prefix

AudriusButkevicius:


users/ignoring.rst
```diff
@@ -54,25 +66,31 @@ The ``.stignore`` file contains a list of file or path patterns. The
    still relative to the repository *root*. Example:
    ``#include more-patterns.txt``.
 
--  A pattern beginning with ``!`` negates the pattern: matching files
+-  A pattern beginning with ``!`` prefix negates the pattern: matching files
```
calmh: "a ``!`` prefix" if we're going to use that sentence construction, I think?

users/ignoring.rst
```diff
@@ -20,6 +20,18 @@ If some files should not be synchronized to other nodes, a file called
 ``#include`` files that *are* synchronized between nodes. All patterns are
 relative to the repository root.
 
+.. warning::
+
+  Please note, ignored files do NOT imply garbage files such as *Thumbs.db*
```
calmh: I think this is a bit strong. I think something less urgent like a `.. note::` with something like "Note that ignored files can block the removal of an otherwise empty directory. See below for the `(?d)` prefix to allow deletion of ignored files." or so would be cleaner

AudriusButkevicius:
Addressed.

calmh:
Good stuff

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 69cdae014946b07f69c98de796c48f3c02ad18fa. Thanks, @AudriusButkevicius!

PR #182, opened 2016-06-01
(?d) example

uglygus:


AudriusButkevicius:
The empty directory quote is confusing, as to me it implies that .DS_Store should be empty, though it doesn't need to be.

uglygus:
I was trying to fit all on one line. How  about this: 
    bar/          # synced
    	.tmpfile  # ignored, (?d) allows deletion if the directory if otherwise empty
        baz       # synced

uglygus:
should read:
(?d).tmpfile  # ignored, (?d) allows deletion if the directory if otherwise empty

AudriusButkevicius:
Ignored, will be deleted if gets in the way of parent directory removal

AudriusButkevicius:
Can yoy add an extra space to fix the comment algnment? Also next time to make commits, drop a message as github does not notify about commits to PR's

users/ignoring.rst
```diff
@@ -113,6 +114,7 @@ and an ``.stignore`` file with the contents::
 all files and directories called "foo", ending in a "2" or starting with
 "qu" will be ignored. The end result becomes::
 
+    .DS_Store	  # ignored, will be deleted if gets in the way of parent directory removal
```
calmh: Shouldn't this actually have the `(?d)` prefix then?

uglygus:
how's that

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 8748339201835d4ca6fc38577d5eb07795a95542. Thanks, @uglygus!

PR #186, opened 2016-06-05
Hint at depreciated -gui-apikey and forthcoming API changes

norgeous:
Should provide at least a hint...
not sure of the wording or formatting of these rst files.
If not good enough please update the docs yourself.


calmh:
I adopted the deprecation note on -gui-apikey; the rest is too early days to document as it's currently undecided.

PR #190, opened 2016-06-15
Update specs for protocol buffers

calmh:
**Not done yet.** Relates to https://github.com/syncthing/syncthing/pull/3276.


specs/bep-v1.rst
```diff
@@ -196,331 +179,163 @@ For C=1:
...
+Protocol Buffer Schema
+~~~~~~~~~~~~~~~~~~~~~~
 
-The **ID** field contains the folder ID, as a human readable string.
+.. code-block:: proto
+
+    message ClusterConfigMessage {
+        repeated Folder folders = 1;
+
+        message Folder {
```
calmh: Do we want to nest messages in the proto definition in the spec? As far as I can understand the encoding is identical and the nesting only causes type naming to differ. We don't want to use it in production as the generated type names are ugly (but https://github.com/gogo/protobuf/issues/186), but perhaps it makes the structure clearer in the spec? Or perhaps not?

specs/bep-v1.rst
```diff
@@ -196,331 +179,163 @@ For C=1:
...
-::
+Fields (Folder Message)
+~~~~~~~~~~~~~~~~~~~~~~~
 
-     0                   1                   2                   3
-     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
-    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-    |                           Reserved                    |T|D|P|R|
-    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+The **id** field contains the folder ID, which is the unique identified of
```
phil-davis: s/indentified/identifier

specs/bep-v1.rst
```diff
@@ -196,331 +179,163 @@ For C=1:
...
+the folder.
 
-:Bit 31 ("R", Read Only):
-    is set for folders that the device will accept no updates from the network
-    for.
+The **label** field contains the folder label, as human readable name for
```
phil-davis: "as human readable name" -> "the human readable name"?

specs/bep-v1.rst
```diff
@@ -196,331 +179,163 @@ For C=1:
...
+The **ignore delete** field is set for folders that the device will ignore
+deletes for.
 
-The **Options** field contains a list of options that apply to the folder.
+The **disable temp indexes** field is set for folders that will not dispatch
+and do not wish to receive progress updates about partially downloaded files
+via Download Progress messages.
 
-Fields (Device Structure)
+The **devices** field is list of devices participating in sharing this
```
phil-davis: "is list of..." -> "is a list of..."

specs/bep-v1.rst
```diff
@@ -196,331 +179,163 @@ For C=1:
...
+The device **id** field is a 32 byte number that uniquely identifies the
 device. For instance, the reference implementation uses the SHA-256 of the
 device X.509 certificate.
 
-The **Name** field is a human readable name assigned to the described device
+The **name** field is a human readable name assigned to the described device
 by the sending device. It MAY be empty and it need not be unique.
 
-The list of **Addressess** is that used by the sending device to connect to
+The list of **addressess** is that used by the sending device to connect to
```
phil-davis: s/addressess/addresses/
And will need to check if this mis-spelling is elsewhere, to do a global change.

specs/bep-v1.rst
AudriusButkevicius: I say no.
In python it makes super ugly to use.

calmh: OK then we'll run non-nested everywhere.

specs/bep-v1.rst
```diff
@@ -62,468 +62,272 @@ another.
...
-    variable length opaque data
+When the **compression** field is **NONE**, the message is directly in
+protocol buffer format.
 
-string<>
-    variable length string
+When the compression field is **LZ4**, the message consists of a four byte
+integer describing the uncompressed message length followed by a single LZ4
+block. After decompressing the LZ4 block it should be interpreted as a
+protocol buffer message just a in the uncompressed case.
```
phil-davis: "just as in"

specs/bep-v1.rst
```diff
@@ -533,363 +337,206 @@ the message. An Index Update MAY NOT be sent unless preceded by an
...
+The **type** field contains the type of the described item. The type is one
+of **file (0)**, **directory (1)**, **symlink to file (2)**, **symlink to
+directory (3)**, or **symlink to unknown target (4)**. The distinction
+between the various types of symlinks is not required on all operating
+systems - the implementation SHOULD nonetheless indicate the target type
+when possible.
 
-:Bit 17 ("D"): is set when the item represents a directory. The block
-   list SHALL be of length zero.
+The **size** field contains the size the file, in bytes. For directories the
```
phil-davis: "size of the file"

specs/bep-v1.rst
```diff
@@ -533,363 +337,206 @@ the message. An Index Update MAY NOT be sent unless preceded by an
...
+The **no permissions** field is set when there is no permission information
+for the file. This is the case when it originates on a file system which
+does not support permissions. Changes to only permission bits SHOULD be
+disregarded on files with this bit set. The permissions bits MUST be set to
+the octal value 0666.
 
-The **Version** field is a version vector describing the updates performed
+The **version** field is a version vector describing the updates performed
 to a file by all members in the cluster. Each counter in the version
 vector is an ID-Value tuple. The ID is used the first 64 bits of the
```
phil-davis: The word "used" seems unneeded here.

specs/bep-v1.rst
```diff
@@ -899,207 +546,126 @@ returned. The following values are defined:
...
-indexes 3, 4, 5 which should be appended to the original list, which implies
-that blocks 0-5 are currently available.
+The **block indexes** field is a list of positive integers, where each
+integer represents the index of the block in the FileInfo message Blocks
+array that has become available for download.
 
-Block indexes MAY be added in any order.
-An implementation MUST NOT assume that block indexes are added in any specific
-order.
+For example an integer with with value 3 represents that the data defined in
```
phil-davis: "with with" - duplicated word.

calmh:
@phil-davis thanks!

calmh:
@syncthing/core @syncthing/documentation Can I ask you for some peer review of the new protocol spec? We don't want to have to change it in the foreseeable future, apart from adding backwards compatible things like new fields in protobuf messages. The thing to review is:

https://build.syncthing.net/docs/33/specs/bep-v1.html

specs/bep-v1.rst
```diff
@@ -62,468 +62,274 @@ another.
...
+    enum MessageType {
+        CLUSTER_CONFIG    = 0;
+        INDEX             = 1;
+        INDEX_UPDATE      = 2;
+        REQUEST           = 3;
+        RESPONSE          = 4;
+        DOWNLOAD_PROGRESS = 5;
+        PING              = 6;
+        CLOSE             = 7;
+    }
```
burkemw3: I highly recommend that enum types have sane default values. In most cases, I think that should be `UNKNOWN = 0`.

If a client doesn't understand a message, they'll be able to tell they can't understand it, instead of not being able to parse the message.

https://developers.google.com/protocol-buffers/docs/proto3#default

specs/bep-v1.rst
```diff
@@ -62,468 +62,274 @@ another.
...
 
-The **Label** field contains the folder label, as human readable name for the folder.
+    message Device {
+        bytes           id                = 1;
+        string          name              = 2;
+        repeated string addresses         = 3;
+        Compression     compression       = 4;
+        string          cert_name         = 5;
+        int64           max_local_version = 6;
```
burkemw3: Seems like this should be uint64, but I haven't actually researched if there's some backwards compatibility issue.

specs/bep-v1.rst
```diff
@@ -1137,59 +705,12 @@ by the actions of other cluster devices.
...
-
-The currently defined values allow maximum file size of 1220 GiB
-(10.000.000 x 128 KiB). The maximum message size covers an Index message
-for the maximum file.
+An implementation MAY impose reasonable limits on the length of messages and
+message fields to aid robustness in the face of corruption or broken
+implementations. An implementation should strive to keep messages short
+and to the point, favouring more and smaller messages over fewer and larger.
+For example, favour a smaller Index message followed by one or more Index
+Update messages rather than sending a very large Index message.
```
burkemw3: Does this mean recipients can reject messages? If it means senders should limit messages sizes, then I think it should say that more clearly. Also, guidelines are good, even if it's just saying what the reference implementation does.

specs/localdisco-v4.rst
```diff
@@ -0,0 +1,77 @@
...
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    /                                                               /
+    \                       Announce Message                        \
+    /                                                               /
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+
+There is no explicit length field as the length is given by the length of
+the discovery announcement packet itself.
+
+The Magic field is a 32 bit word representing 0x2EA7D90B in big endian byte
```
burkemw3: I recommend stating network byte order (i.e. big endian) for consistency with BEP spec. (Or maybe the other way around, but I like the BEP spec way better, personally).

specs/bep-v1.rst
```diff
@@ -62,468 +62,274 @@ another.
...
 
+In this document, in diagrams and text, "bit 0" refers to the *most
+significant* bit of a word; "bit 15" is thus the least significant bit of a
+16 bit word (int16) and "bit 31" is the least significant bit of a 32 bit
+word (int32). Non protocol buffer integers are always represented in network
+byte order (i.e., big endian) and are signed unless stated otherwise, but
+when describing message lengths negative values do not make sense and the
+most significant bit MUST be zero.
+
+The protocol buffer schemas in this document are in ``proto3`` syntax. This
```
burkemw3: I suggest a link to proto3 docs instead of fixed-width text

burkemw3:
Overall, appears mostly sane to me. Major recommendation is UNKNOWN field for enums.

I did not compare to the existing protocol to make sure everything is represented.

specs/bep-v1.rst
calmh: Yeah, we do assume reasonable defaults everywhere I think, but what's the purpose of being able to send a message of type UNKNOWN though? Or is this an accepted way of making a field explicitly `required`?

specs/bep-v1.rst
calmh: I don't like uints for things that may get arithmetic done on them.

specs/bep-v1.rst
calmh: It does, and the only reasonable way to do that at this point is to tear down the connection. We did have some very specific guidelines here as you can see, but that's not supported by protobuf. In Syncthing we enforce a 500M max message size which is quite large, but it could certainly be mentioned here.

specs/localdisco-v4.rst
calmh: Yeah we note somewhere early on that all integers are network byte order to avoid repeating it everywhere. :) We can repeat it once for emphasis though.

specs/bep-v1.rst
calmh: I want the definitions in the text so they can be discussed there, but it wouldn't hurt to provide a full schema file of course.

calmh:
Thanks for the review, I'll address the issues!

specs/localdisco-v4.rst
calmh: I kind if slid past this one

> I like the BEP spec way better, personally

Apart from this still being BEP, just the new version of it - what did you prefer in the previous version? Obviously I liked it when I wrote it, but the point of this change is to bring future extensibility into it and stop breaking the protocol every time we need to change something.

specs/bep-v1.rst
burkemw3: The purpose isn't to send an UNKNOWN message type but to receive it.

For example, lets say the protocol implements a new `WHIZ_BANG = 8` message type and starts sending it out. Old clients do not know about `WHIZ_BANG`, so the message type will default to `CLUSTER_CONFIG`, and parsing will fail, and ordering constraints will fail, and the client will only be able to throw up their hands and fail.

With the `UNKNOWN = 0`, old clients will see UNKNOWN, drop the message on the floor, and likely be able to continue participating.

We do have some other safeguards in place for this (e.g. protocol versioning). This would be an addition that I believe will give BEP options for inter-version compatibility, however.

calmh: Oh! I would have expected it to deserialize with the value set to 8 and there not existing any matching enum value for it. But if that's not the case then I'm all into what you're saying. Let me test this.

specs/localdisco-v4.rst
burkemw3: I see the network byte order call out early on in the BEP spec. It just so happens that the BEP spec appears earlier in my github review than the discovery spec, but that is not necessarily globally true. I agree that once at the top of each spec is sufficient.

I was ambiguous in "I like the BEP spec way better, personally". What I meant was that there was a difference between the way local discovery and BEP described byte order. I want to make them consistent. They could be converging on BEP's style or LD's style. I like the (BEP spec way) better. I was not saying 'I like the (BEP spec) (way better).'

calmh: I totally missed that you were talking about the local disco document here and not the BEP spec, and thought you were referring to the old BEP vs new proto buf thing. Sorry, then it makes perfect sense.

specs/bep-v1.rst
```diff
@@ -62,468 +62,277 @@ another.
...
+
+    message Header {
+        MessageType        type        = 1;
+        MessageCompression compression = 2;
+    }
 
-(unsigned) int:
-    (unsigned) 32 bit integer
+    enum MessageType {
+        UNKNOWN           = 0;
```
calmh: @burkemw3 I added this, but I'm not sure it's necessary after all. Quoting the proto3 spec (https://developers.google.com/protocol-buffers/docs/proto3#enum, a bit further down):

> During deserialization, unrecognized enum values will be preserved in the message, though how this is represented when the message is deserialized is language-dependent. In languages that support open enum types with values outside the range of specified symbols, such as C++ and Go, the unknown enum value is simply stored as its underlying integer representation. In languages with closed enum types such as Java, a case in the enum is used to represent an unrecognized value, and the underlying integer can be accessed with special accessors. In either case, if the message is serialized the unrecognized value will still be serialized with the message.

So how undefined values are represented depends on the language, but it doesn't seem that it ever folds to the default. As such is seems cleaner to not have `UNKNOWN` here?

AudriusButkevicius: Agree, no need to define unknown as part of the spec

calmh:
Seeing as the code has been merged and I think the docs here are up to date with the new code, I propose this for merge as well.

specs/bep-v1.rst
```diff
@@ -62,468 +62,276 @@ another.
...
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |         Header Length         |
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    /                               /
+    \            Header             \
+    /                               /
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+    |         Message Length        |
+    |           (32 bits)           |
+    +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
AudriusButkevicius: Why header len does not have 32bits stuff?

AudriusButkevicius: Thats not what I mean. I don't like that we put number of bits in brackets for 32bit ints and not 16bit ints

calmh: Aha! The diagram is 16 bits wide, so the "header length" field is by definition 16 bits. The "message length" takes up two lines as is 32 bits. This is how we used to indicate 64 bit fields in the 32 bit wide diagrams previously.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as a9b3fef697ed0c546ee10594878fa61be0579e69. Thanks, @calmh!

AudriusButkevicius:
This can always be changed, so don't care much if its incorrect, and yes, I did not read the whole thing.

zertrin:
Great work and Documentation !

Wondering if such extensive changes are still part of BEP **_version 1_**? Shouldn't you bump the version of the BEP?

AudriusButkevicius:
It's not yet official, hence iteration on the same non released version.

PR #201, opened 2016-07-01
users: reflect discosrv and relaysrv rename

aviau:
To be merged after https://github.com/syncthing/syncthing/pull/3371


calmh:
Looks good. Let me take a general look around for any more things that may need updating or refer to the old pages before merging.

users/strelaysrv.rst
```diff
@@ -74,12 +74,12 @@ as root or a privileged user, see `Running on port 443 as an unprivileged user`_
 below. In principle something similar to this should work on a Linux/Unix
 system::
 
-    $ sudo useradd relaysrv
-    $ sudo mkdir /etc/relaysrv
-    $ sudo chown relaysrv /etc/relaysrv
-    $ sudo -u relaysrv /usr/local/bin/relaysrv -keys /etc/relaysrv
+    $ sudo useradd strelaysrv
```
aviau: Should we suggest syncthing-relaysrv as a user here to reflect syncthing-relaysrv.service?

calmh: I was unclear in the other PR, but it would make more sense to me to make both the user and the service "strelaysrv" as well.

aviau: Oh, okay. Then the docs are fine and I should update the other PR.

aviau: Done

aviau:
I have restored file names without the `st` prefix because I am currently writing manpages in another PR and it makes more sense that the manpages have the prefix.

calmh:
I think these pages should be merged into the manages, really.

aviau:
> these pages should be merged into the manages

Yeah. I'll do that in the other two PRs and close this one.

PR #212, opened 2016-07-27
users/config: Add missing protocol prefix

ProactiveServices:


AudriusButkevicius:
I think we should say <protocol> and then define what the available protocols currently are.

calmh:
@st-review merge it

users/config: Protocol prefixes on addresses

st-review:
:ok_hand: Merged as a1b2fdc3397a5671040ebd272d3aa4140a6ad585. Thanks, @ProactiveServices!

PR #216, opened 2016-08-03
Add svc/random/string REST endpoint.

JodusNodus:


calmh:
Lets just call it "alphanumeric" and not specify the character set as there are a bunch of characters we don't use to avoid confusing users. Otherwise LGTM, thanks!

calmh:
@st-review merge it

rest: Add svc/random/string REST endpoint

st-review:
:ok_hand: Merged as 709ebc5a73d94f2b0bfecfb0ae41e0db122639b3. Thanks, @JodusNodus!

PR #219, opened 2016-08-10
Explicitly mention blocking behavior

imsodin:
As discussed in IRC, it might not be automatically clear that syncthing only returns once the scan has finished. This should make it clear.


calmh:
@st-jenkins merge it

rest: Explicitly mention blocking behavior of /rest/db/scan POST

st-review:
:ok_hand: Merged as a4da3178ee31fe7d337982a235d1a900be78843b. Thanks, @imsodin!

PR #220, opened 2016-08-22
Typo in stdiscosrv.rst

MChudak:
Synthing -> Syncthing


AudriusButkevicius:
Needs an extra equals at the bottom, otherwise it will not be generated into a header.

PR #230, opened 2016-10-03
Add Host Check bypass to apache and nginx sample config

avrelaun:


users/reverseproxy.rst
```diff
@@ -39,6 +40,7 @@ Nginx
 
     location /syncthing/ {
       proxy_set_header        Host $host;
+      #proxy_set_header        Host localhost; # Uncomment to bypass Host Check
```
canton7: Probably wants a comment saying to remove `proxy_set_header        Host $host;` if you do this?

calmh:
This needs a note on the concerns associated with doing this, namely that you open yourself up for attacks that Synchting doesn't protect against when fooled, hence this is a bad idea unless the proxy is doing some sort of authentication of the user before passing the request onwards.

avrelaun:
Is there a more secure way to access syncthing remotely ? I could add a link pointing to the security implications as detailed here: https://docs.syncthing.net/users/faq.html#why-do-i-get-host-check-error-in-the-gui-api

users/reverseproxy.rst
avrelaun: see 2e83649

calmh:
I think lying about the Host header is probably the wrong solution anyhow.

PR #234, opened 2016-10-16
remote gui via ssh from Windows

dinosore:


dinosore:
As there has been a pull request committed with @calmh since this one, I thought that it may be worth bumping.

As I have managed to access the gui on a raspberry pi from both linux and windows using ssh -L, I thought that windows users with remote Linux (quite common I would think) might be interested.

I realise that it is only a brief explanation and ripe for improvement by those more knowledgeable than I. 

If it isn't worth including until improved, I'd welcome the views of @canton7.

canton7:
I read this, but I've got very little experience with any of it, so decided not to comment.

calmh:
@st-review merge it

users/faq: Accessing remote GUI via ssh from Windows

st-review:
:ok_hand: Merged as 2ca8fef93530f74384273272409df916f86b3239. Thanks, @dinosore!

PR #245, opened 2016-11-09
Add caveat section to proxy doc

ckeen:
This warns users about mixing all_proxy and http_proxy env variables.

### DIFF
--- a/users/proxying.rst
+++ b/users/proxying.rst
@@ -45,3 +45,14 @@ prevent the fallback behavior. For example::
   ...
   [monitor] 11:33:13 INFO: Proxy settings detected
   [monitor] 11:33:13 INFO: Proxy fallback disabled
+
+Caveats
+-------
+
+Note that one needs to unset any existing ``http_proxy`` /
+``https_proxy`` variables when using ``all_proxy`` as the SOCKS proxy
+is used first, then syncthing will try to access peers through the
+specified HTTP proxy. This will fail in most cases, as the specified
+proxy is often an internal name and unknown to the other side of your
+proxy or it will not accept connections from the other side of your
+firewall.
--- a/users/proxying.rst
+++ b/users/proxying.rst
@@ -49,10 +49,8 @@ prevent the fallback behavior. For example::
 Caveats
 -------
 
-Note that one needs to unset any existing ``http_proxy`` /
-``https_proxy`` variables when using ``all_proxy`` as the SOCKS proxy
-is used first, then syncthing will try to access peers through the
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
This is not true though, in some cases you might have to have that set to access discovery servers.

ckeen:
@AudriusButkevicius well this would work only if the proxy set is valid on the SOCKS endpoint, isn't it?

AudriusButkevicius:
Well both need to be valid

ProactiveServices:
Having a look through the code I can't figure out which envvar would take precedence, if any. If Syncthing will always prefer `all_proxy` then it seems the caveat isn't needed. 
Relatedly - if Syncthing does use `http(s)_proxy` should this be documented, or is it obvious to a Linux user?

AudriusButkevicius:
I believe that comes from the stdlib, yet could be documented.

ckeen:
I think this discussion is totally besides my original point. The behaviour I have observed is:

set all_proxy to a SOCKS port tunnerling into a machine behind the firewall on the internet.

localhost:$SOCKS ---> proxyhost

Then requests will get tunnelled through to proxyhosts but the requests there will have the defined http_proxy in them.

In my example it's http_proxy=proxy:8080, that will fail on proxyhost as there is no machine called proxy.

AudriusButkevicius:
Are you saying http requests should not go through the socks proxy if http proxy is defined?

ckeen:
I am saying that it should go through the socks proxy but users that don't wish to use another http*_proxy on the proxy host should be made aware of the fact that syncthing will do so. Or if that behaviour is not the desired one, it's a bug after all not just a documentation issue.

AudriusButkevicius:
Well I can imagine that you might want either behaviour, yet we do not support that, so we should just document that any http proxy specified will be resolved after the socks proxy

ckeen:
I have updated the patch, maybe that's a better wording then?

calmh:
I have no idea if the behavior described is the one we actually have, but the wording is fine to me and I'm going to assume you all have done your homework on this :)

calmh:
@st-review merge it

st-review:
@calmh: Build status is `pending`. I'll wait until it goes green and then merge!

calmh:
lolwhat

calmh:
@st-jenkins  retest this please

st-review:
:ok_hand: Merged as 5d02d82339a94ecabf464cbcfcdf3f16687b3749. Thanks, @ckeen!

PR #250, opened 2016-11-21
FAQ: How do I sync the ~/Sync/ folder between my devices?

jonathancross:
As a new user, I was very confused by this process.
So I am adding it to the FAQ.

Direct link to preview: https://build.syncthing.net/docs/105/users/faq.html#how-do-i-sync-the-sync-folder-between-my-devices


calmh:
Handled the other way, but thanks for the prod!

jonathancross:
Okay, thanks.

PR #257, opened 2016-11-24
fixed supervisor config path

Stefan-Code:
At least on my recent Ubuntu system, the right path for the supervisord.conf file is `/etc/supervisor/supervisord.conf`

EDIT: this seems to be a Ubuntu/Debian thing, otherwise the config path is indeed /etc/supervisord.conf

### DIFF
--- a/users/autostart.rst
+++ b/users/autostart.rst
@@ -198,7 +198,7 @@ On Ubuntu-like systems
 Using Supervisord
 ~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
I'd say "or"

Stefan-Code:
@AudriusButkevicius ok?

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as e91a88d05d86901b823afc42904f132a91b11340. Thanks, @Stefan-Code!

PR #259, opened 2016-12-16
rename master and read-only to send-only

Smiley73:
Updating the documentation to match the new "send-only" folder name.


Smiley73:
I wasn't sure if I should rename foldermaster.rst to foldersendonly.rst. Especially since the receiveonly folder type is just around the corner.

AudriusButkevicius:
Yeah you should rename it, and I suspect there are some references to that file in other files.

calmh:
It is also linked from the GUI so at minimum we'll need a redirect for historical purposes.

Smiley73:
@calmh I added a "page moved" to the old foldermaster.rst, but could not find a way to auto redirect.

AudriusButkevicius:
@st-review lgtm

st-review:
@AudriusButkevicius: Noted! Need another LGTM or explicit merge command.

calmh:
Looks good to me as well, but lets not merge this until it's actually out there in a release, lest we cause more confusion than necessary. I'll update the link the GUI while I remember it.

calmh:
If we wanted to be a little bit future proof and anticipate that other pull request, this should maybe be a `foldertypes.rst` that currently only describes send-only (the normal mode being self evident>?

AudriusButkevicius:
Agree.

Smiley73:
I like the foldertypes.rst suggestion. I'll work on it as time permits (might be later this week).

JohnVeness:
I also agree with renaming the file to foldertypes.rst, but as has already been said, there definitely should be a redirect from the old page name. People will keep bookmarks and links from external websites pointing to particular pages, so in general whenever a page URL is changed, a redirect from the old one should be in place for some years if possible.

Smiley73:
Okay I made the suggested changes. This should be ready to go.

AudriusButkevicius:
Looks perfect. Leaving for @calmh to merge as the release is to go out

st-review:
:ok_hand: Merged as e44c72f70b1e5e2e18955829642baaf40b16c5bf. Thanks, @Smiley73!

PR #261, opened 2016-12-21
Updated SSH tunneling command

mclang:
Added parameters -f and -N so that SSH tunnel is created in the background without opening shell.

### DIFF
--- a/users/firewall.rst
+++ b/users/firewall.rst
@@ -71,11 +71,11 @@ If you have SSH access to the machine running Syncthing but would rather not
 open the web GUI port to the outside world, you can access it through a SSH
 tunnel instead. You can start a tunnel with a command like the following::
 
-    ssh -L 9999:localhost:8384 machine
+    ssh -f -N -L 9999:localhost:8384 machine
 
 This will bind to your local port 9999 and forward all connections from there to
-port 8384 on the target machine. This still works even if Syncthing is bound to
-listen on localhost only.
+port 8384 on the target machine without opening SSH shell to the remote machine.
+This still works even if Syncthing is bound to listen on localhost only.
 
 Via a Proxy
 -----------


calmh:
I suspect this will just cause more confusion about what happened, for the people unfamiliar with these options. Those familiar with them will already know what to do.

mclang:
Well I didn't remember, I had to look for the right parameters.
Maybe this possibility should be added as another command, or explained better to prevent confusion?

calmh:
Oh, I don't know those by heart either. I just think that since this document is explaining ssh forwarding to someone who may never have used it at all, ever, keeping it as simple as possible is a win.

mclang:
So you think that leaving current command as it is and adding another example like what I am proposing as alternative is also too confusing for new users?

ProactiveServices:
If by default, or in most cases, the example command doesn't work, I'd suggest that the example should be "fixed" to include the extra parameters. Otherwise it should be left as it is, since it's not our place to document `ssh`.

JohnVeness:
I agree it isn't the place of Syncthing to document ssh. If you went down that route, all you've documented so far is a command-line method, whereas it is also possible to do ssh tennelling using, for example, ConnectBot in Android, or PuTTY in Windows, and you don't really want to start documenting all those other ways of doing things.

I wonder, is there a generic "ssh tunnelling" instructions on a well-maintained website (a bit like https://portforward.com/) that can be pointed to? Then you could just say something like "You can use SSH tunnelling to access the web interface from another computer, see XXXXX.com".

ProactiveServices:
@mclang care to take another look at your proposal and the comments since?

mclang:
Sorry, I somehow missed notifications about this

Link to compiled docs gives 404, so I can't check them.

You are right, it is no place for Syncthing to start documenting all things SSH can do, so I back away and close this pull request.

Cheers!

PR #263, opened 2016-12-21
make.bat: Various minor fixes/improvements

ProactiveServices:
Quoted path given to del, rmdir /s commands. Ouch!
Removed spurious parameter in del command.
dirhtml: Added mkdir for successful build
Initial sphinx-build execute test: correct nul redirection
Forces removal of any quotes around envvars then explicity calls them in
each instance, to ensure they work if spaces are used in paths/vars.
Added error text if no matching arguments passed.
Correct slashes to make them Windowsy.

### DIFF
--- a/make.bat
+++ b/make.bat
@@ -4,9 +4,12 @@ REM Command file for Sphinx documentation
 
 if "%SPHINXBUILD%" == "" (
 	set SPHINXBUILD=sphinx-build
+) else (
+	REM Strip quotes from envvar, as we explicitly quote all paths that may have spaces in them.
+	set SPHINXBUILD=%SPHINXBUILD:"=%
 )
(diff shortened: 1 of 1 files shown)


ProactiveServices:
Note I did run each test to check they worked. Two of the latex ones don't work at all as there seems to be a file missing, and I can't test doctest as I haven't figured out how to steer Sphinx extensions yet.

canton7:
Given that this file is copied from the Sphinx project, I'm going to go ahead and say you should probably get your changes included upstream, and then let them flow down to Syncthing? If we diverge from upstream, that can only cause issues in the future when we try and update Sphinx.

ProactiveServices:
Whoops, I hadn't realised that. Thanks for the pointer.

PR #282, opened 2017-01-19
Exclude draft/ and README.rst to fix warnings and update copyright year

imsodin:


conf.py
```diff
@@ -47,7 +47,7 @@
 
 # General information about the project.
 project = u'Syncthing'
-copyright = u'2015, The Syncthing Authors'
+copyright = u'2017, The Syncthing Authors'
```
canton7: No! No No No!

The date is the date at which you claim copyright on the work. If you update this to 2017, you're claiming that you wrote the work in 2017. If someone copied the work in 2016, but you're claiming you wrote it in 2017, your claim to copyright looks very shaky.

[More info on SO](http://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated/2391555#2391555).

imsodin: Will fix this, thanks for the pointer.

canton7: It needs to be 2015, not 2016, as you changed it back to.

imsodin: Done in https://github.com/syncthing/docs/commit/6b4203296b3b6b62e1bb29ec75ea181d56c7611e, and really sorry for this mess...

PR #285, opened 2017-01-24
specs: Remove redundant specification index page

imsodin:
This is a matter of taste, and I don't like having a title `SPECIFICATIONS` and immediately following the tile `Specifications` with more or less redundant information.
If the short descriptions should be preserved, they could be added to the individual specs pages between the main title and the first subtitle.
Opinions?


canton7:
-1 from me I'm afraid. We've lost a useful summary.

AudriusButkevicius:
Don't care much, but my hoarder character says "keep it"

calmh:
https://docs.syncthing.net/specs/index.html is a useful page by itself and I think it should stick around.

There is some redundancy in the menu links. If the subpages were actual subheadings instead it would be cleaner. If that's not doable, I think status quo is preferable.

kluppy:
Change the link in the menu to Overview and keep the rest from the original.
We keep the descriptions and the page makes sense without the Nav bar having redundant headings.

imsodin:
I called it "Introduction" for consistency with dev section.

ProactiveServices:
@st-review merge

st-review:
:ok_hand: Merged as fa4d267fe42f2fd70c91777c4061670d8783a720. Thanks, @imsodin!

ProactiveServices:
You managed to please everyone? Sorcerer.

PR #288, opened 2017-01-30
Add an entry about daemonization on Linux

Ferroin:
More than 90% of the people who come asking about this are likely to be using a Linux system which doesn't have systemd.  Apparently some such individuals can't do any research beyond Syncthing documentation and the issue tracker, so add an entry to the FAQ to give them some reasonable pointers on what to look at to figure out how to do this on their own without bugging the developers.

Point people at the example configs first, as they may not know they exist (I actually didn't for a few months after I started using Syncthing), then point them at alternatives.

start-stop-daemon is found on a majority of Linux systems these days.  It's practically ABI on Debian derived systems (including Ubuntu), and is a core part of the init system on Gentoo and a handful of other non-systemd distros.  As a result of this and the fact that it's insanely easy to use (and usually very well documented), suggest it first and then point people at less immediately available stuff like daemontools and S6.

Hopefully this should cover the options for most systems, and should also mitigate the number of people asking about how to do this (and when such an issue pops up, whoever answers it can point them here).  The question didn't seem to fit well anywhere specifically, so I just put it at the end of the page.

### DIFF
--- a/users/faq.rst
+++ b/users/faq.rst
@@ -437,3 +437,21 @@ GitHub does not provide a single URL to automatically download the latest
 version. We suggest to use the GitHub API at
 https://api.github.com/repos/syncthing/syncthing/releases/latest and parsing
 the JSON response.
+
+
+How do I run Syncthing as a daemon process on Linux?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


calmh:
Lgtm

calmh:
@st-review merge it

users/faq: Add entry about daemonization

st-review:
:ok_hand: Merged as 58fe076cc54d2d42eaed3ae0df57aef120a87a32. Thanks, @Ferroin!

PR #294, opened 2017-02-13
Two FAQ updates

grinapo:
- about audit trail (without any tech details though)
- about how changes are seen, based on the forum discussion

Feel free to correct or edit! Thanks!

### DIFF
--- a/users/faq.rst
+++ b/users/faq.rst
@@ -409,27 +409,27 @@ How can I view the history of changes?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The default web GUI contains a ``Global Changes`` button at the bottom which
-displays the ongoing changes since the last (re)start of the daemon. 
+displays the ongoing changes since the last (re)start of Syncthing. 
 With the ``-audit`` option you can enable a persistent, detailed log of changes
-and most activities the daemon does, which contains a ``JSON`` formatted 
+and most activities, which contains a ``JSON`` formatted 
 sequence of events in the ``~/.config/syncthing/audit-_date_-_time_.log`` file.
 
 Does the audit log contain every change?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The audit log (and the ``Global Changes`` window) see the changes what your
-Syncthing see. When Syncthing is continuously connected it usually see every change
-happening, and see which nodes initiated the change. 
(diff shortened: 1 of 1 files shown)


imsodin:
I directly committed some language changes, github review isn't very practical to propose small changes to text.

Maybe neighbour should be replaced with remote device to be consistent with the GUI?

Are you sure about the second FAQ entry? I am a a bit confused by the part about multiple remote devices. They should already be in a consistent state so we normally don't need to decide between remote devices, just between the received and local index.

grinapo:
@imsodin FAQ: yes, but if anyone wants to use it any other way: feel free. 

About consistency: they are not necessarily in sync; for example when there are two groups of nodes which were split, doing their changes, and I connect them both they will both believe their latest is The Reality™ while I will see that either one of them is newer or they're conflicting. I will be the one who actually pull both groups in sync with the rest of the mesh (if I undesrstand correctly).
This entry was born from the questions about reliability of the "originating node" in audit log and the realisation that the protocol "collapses" changes between last and current state (so "non-last changes"  will not be visible in local audit log either). Phrasing that is not easy, and I'm an outsider as well, so every input is most welcome.

users/faq.rst
```diff
@@ -405,6 +405,32 @@ creates ``syncthing.log`` in Syncthing's home directory (run ``syncthing
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The audit log (and the ``Global Changes`` window) see the changes that your
+Syncthing sees. When Syncthing is continuously connected it usually sees every change
+happening immediately and thus knows which node initiated the change. 
+When topology gets complex or when your node reconnects after some time offline, 
+Syncthing synchronises with its neighbours: It gets the latest synchronised state 
+from the neighbour, which is the *result* of all the changes between the last 
+known state (before disconnect or network delay) and the current state at the 
+neighbour, and if there were updates, deletes, creates, conflicts, which were
```
ProactiveServices: IMO this could be succinctly described by something more like "...which is the result of all changes since last connection, not the individual changes which led to the current state." I feel it's a little too wordy as it is.

grinapo: @ProactiveServices fine with me, I'm not a native speaker.

calmh:
@st-review merge it

users/faq: Describe global changes and auditing

st-review:
:ok_hand: Merged as ed81a8726322d729b25251385a8b0a9fe20c5389. Thanks, @grinapo!

PR #295, opened 2017-02-14
Update autostart.rst

jaimejim:


ProactiveServices:
Not sure why this needs to duplicate https://docs.syncthing.net/users/contrib.html?highlight=debian#debian-ubuntu

PR #300, opened 2017-02-22
dev/rest: Document /rest/system/browse (fixes #112)

### LINKED ISSUE #112: Documentation: "/rest/system/browse" is not present in docs
Method is simple but it's good to have all methods in docs without exceptions)

imsodin:
See #112.
Should there also be a warning about (disk) resource intensity as for [/rest/db/browse](https://docs.syncthing.net/rest/db-browse-get.html)?

### DIFF
--- a/rest/system-browse-get.rst
+++ b/rest/system-browse-get.rst
@@ -1,0 +1,39 @@
+GET /rest/system/browse
+===================
+
+Returns a list of directories matching the path given by the optional parameter
+``current``. The path can use `patterns as described in go's filepath package
+<https://golang.org/pkg/path/filepath/#Match>`_ and a `*` will be suffixed. It
+is also possible to give a path to a directory to list its contents by adding a
+trailing path separator.
+
+.. code-block:: bash
+
+    $ curl -H "X-API-Key: yourkey" localhost:8384/rest/system/browse | json_pp
+    [
+        "/"
+    ]
+
+    $ curl -H "X-API-Key: yourkey" localhost:8384/rest/system/browse?current=/var/ | json_pp
+    [
+        "/var/backups/",
+        "/var/cache/",
+        "/var/lib/",
+        "/var/local/",
+        "/var/lock/",
+        "/var/log/",
+        "/var/mail/",
+        "/var/opt/",
+        "/var/run/",
+        "/var/spool/",
+        "/var/tmp/"
+    ]
+
+    $ curl -H "X-API-Key: yourkey" localhost:8384/rest/system/browse?current=/var/*o | json_pp
+    [
+        "/var/local/",
+        "/var/lock/",
+        "/var/log/",
+        "/var/opt/",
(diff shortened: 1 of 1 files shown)


imsodin:
Regarding the resource intensity: No warning should be ok according to https://github.com/syncthing/docs/issues/112#issuecomment-281699715.

ProactiveServices:
Did you mention the suffixing of `*` on purpose? IMO it inferred that only `*` was supported, so not sure it should be mentioned on its own.

rest/system-browse-get.rst
```diff
@@ -2,10 +2,9 @@ GET /rest/system/browse
 ===================
 
 Returns a list of directories matching the path given by the optional parameter
-``current``. The path can use `patterns as described in go's filepath package
-<https://golang.org/pkg/path/filepath/#Match>`_ and a `*` will be suffixed. It
-is also possible to give a path to a directory to list its contents by adding a
-trailing path separator.
+``current``. The path can use `patterns as described in Go's filepath package
+<https://golang.org/pkg/path/filepath/#Match>`_. It is also possible to list a
+directory's contents by adding a trailing path separator.
```
imsodin: I don't think the asterisk bit is insignificant. According to the linked patterns `foo` does not match `foobar` but  for this API call it does.

imsodin:
> Did you mention the suffixing of * on purpose? IMO it inferred that only * was supported, so not sure it should be mentioned on its own.

@ProactiveServices This is not at all what it means. Everything in the link is supported, but in addition to every query string you give "*" is appended. So according to the linked Go spec `foo` does not match `foobar` but for this API call it does, as `foo` becomes `foo*`.

imsodin:
Looking at this again I am not sure whether the appending an asterisk in all cases is intentional. I think it should only be appended when listing subdirectories, i.e. the the code at https://github.com/syncthing/syncthing/blob/master/cmd/syncthing/gui.go#L1235 should be

``` go
        if strings.HasSuffix(current, pathSeparator) {
                if !strings.HasSuffix(search, pathSeparator) {
		        search += pathSeparator
                }
                search += "*"
	}
subdirectories, _ := osutil.Glob(search)
```

Or is there a particular reason for this specific behaviour?

calmh:
I think intentional, as this is the mechanism the GUI uses to auto complete `c:/foo` to `c:/foobar` and `c:/foobaz` if those are present.

imsodin:
I kind of forgot about this.

@ProactiveServices Can you have a look whether it is understandable like this?
I don't mention listing subdirectories any more, as it is inferred by appending an asterisk to all patterns (to make it explicit, I added a small example).

calmh:
@st-review merge it

st-review:
:ok_hand: Merged as 79aaf6a764945ff734a4b3575d0e1b173617015d. Thanks, @imsodin!

PR #304, opened 2017-03-03
Add Caddy to reverseproxy.rst

vrischmann:
Small change to add [Caddy](https://caddyserver.com/) in the reverse proxy section.

Caddy is super easy to configure so normally I only have to add the `proxy` block, but with syncthing I had to disable some HTTP timeouts, otherwise it just wouldn't work. With the default timeouts that you can see [here](https://caddyserver.com/docs/timeouts) syncthing would inevitably pop the "Connection Error" popup, and I would see quite a lot of connection errors in my Chrome console.

### DIFF
--- a/users/reverseproxy.rst
+++ b/users/reverseproxy.rst
@@ -46,6 +46,21 @@ Nginx
       proxy_pass              http://localhost:8384/;
     }
 
+Caddy
+~~~~~
+
+.. code-block:: nginx
+
+    proxy /syncthing localhost:8384 {
+        transparent
+    }
+    
+    timeouts {
+        read none
+        write none
+        header none
+    }
+
 Folder Configuration
 --------------------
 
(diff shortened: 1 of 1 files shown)


users/reverseproxy.rst
```diff
@@ -46,6 +46,21 @@ Nginx
       proxy_pass              http://localhost:8384/;
     }
 
+Caddy
+~~~~~
+
+.. code-block:: nginx
```
imsodin: Nitpicking: Doesn't look like Caddy uses the same config structure as nginx, but is also not known by sphinx itself, so should probably just be dropped.

vrischmann: Right, with the GitHub preview it looked ok but I see in the compiled docs it doesn't work.

imsodin:
@st-review merge

users/reversproxy: Add Caddy configuration

st-review:
:ok_hand: Merged as f0e3aa58947490b35b06db06853ff3d1a4998681. Thanks, @vrischmann!

PR #305, opened 2017-03-04
dev/rest: Update pause/resume to syncthing/syncthing#3989

imsodin:
syncthing/syncthing#3989

### DIFF
--- a/rest/system-pause-post.rst
+++ b/rest/system-pause-post.rst
@@ -3,6 +3,6 @@ POST /rest/system/pause
 
 Pause the given device or all devices.
 
-Takes the optional parameter ``device`` (device string). When ommitted,
+Takes the optional parameter ``device`` (device ID). When ommitted,
 pauses all devices.  Returns status 200 and no content upon success, or status
 500 and a plain text error on failure.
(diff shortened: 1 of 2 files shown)


ProactiveServices:
@st-review lgtm

st-review:
@ProactiveServices: Noted! Need another LGTM or explicit merge command.

rest/system-pause-post.rst
```diff
@@ -1,6 +1,8 @@
 POST /rest/system/pause
 =======================
 
-Pause the given device.
+Pause the given device or all devices.
 
-Takes the mandatory ``device`` parameter and returns status 200 and no content upon success, or status 500 and a plain text error on failure.
+Takes the optional parameter ``device`` (device name). When ommitted,
```
imsodin: It is actually the device name (as in UI, definitely better than device string), not the device ID. See:
https://github.com/syncthing/syncthing/blob/master/cmd/syncthing/gui.go#L1130

calmh:
All REST methods that deal with devices take device *ID*s, nothing else.

imsodin:
For some reason my mind made the connection string -> name and was thus confused by `DeviceIDFromString`. I should have looked what the function actually does. Thanks for the correction.

calmh:
@st-review merge

st-review:
:ok_hand: Merged as e58c571a60d3e958f442915783e2e7892b97326b. Thanks, @imsodin!

PR #314, opened 2017-04-06
Remove the candidate warning about the snap

elopio:
I've been using the stable snap for a while without any issues, with updates and everything. I think it's safe now to recommend people to install the stable channel.

### DIFF
--- a/users/contrib.rst
+++ b/users/contrib.rst
@@ -79,9 +79,9 @@ On any of the `Linux distributions that support snaps <https://snapcraft.io/docs
 
    $ snap install syncthing
 
-If you want to help testing the upcoming release, you can install the snap from the candidate channel:
+If you want to help testing the upcoming release, and get the newer features earlier, you can install the snap from the candidate channel:
 
-   $ snap install syncting --candidate
+   $ snap install syncthing --candidate
 
 CentOS
 ~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


imsodin:
It should still mention that one can also use the candidate channel to get release candidates.

elopio:
That's a good point @imsodin. How does it look like now?

users/contrib.rst
```diff
@@ -77,9 +77,11 @@ Snap
 
 On any of the `Linux distributions that support snaps <https://snapcraft.io/docs/core/install>`_: ::
 
-   $ snap install syncthing --candidate
-   
-(Note that this is an experimental and unstable release, at the moment)
+   $ snap install syncthing
+
+If you want to help testing the upcoming release, you can install the snap from the candidate channel:
```
imsodin: I would also mention the positive aspects, something like "and get the newest features".

users/contrib.rst
```diff
@@ -77,9 +77,11 @@ Snap
...
+
+   $ snap install syncting --candidate
```
imsodin: s/syncting/syncthing/

elopio: Oh, I need to stop being in the computer and go for food :)

Thanks, I will make the changes.

calmh:
@st-review lgtm

st-review:
@calmh: Noted! Need another LGTM or explicit merge command.

imsodin:
@st-review lgtm

st-review:
:ok_hand: Merged as ec9259ee423ca99e0b47a22d17dc0ee4d468ecb7. Thanks, @elopio!

elopio:
thank you!

PR #316, opened 2017-05-10
Path-fix and note added.

woosting:


calmh:
Hmm. The thing you corrected is indeed a bit unclear, but I don't think it's supposed to imply that Syncthing is installed into `/Syncthing`? At least I hope not because that would be really weird. Perhaps a larger rewrite to clarify would be good.

woosting:
Agreed, the path receding `/` (slash) mentioned in the note was a typo (and should be removed to stay in-line with the article's convention that `Syncthing/` is the installation directory (or at least the location of the executable)).

However the most important correction is in the: Syncthing/etc/**linux-systemd**/system/syncthing@.service path addition in the [How to set up a system service][1] section (which I do not consider clarification, but rather correction).

[1]:https://build.syncthing.net/docs/374/users/autostart.html#how-to-set-up-a-system-service

calmh:
The thing about symlinking to `/Syncthing/syncthing` on Debian derivates is clearly incorrect though.

woosting:
> The thing about symlinking to `/Syncthing/syncthing` on Debian derivates is clearly incorrect though.

@calmh Agreed; I made a change to fix the CLI syntax error.

**Context:** Syncthing would not start when I installed manually by wgetting the binaries rather than using apt-install (in that case I did not have the issue). Analyzing the logs made me come to the conclusion that I stated in the note. I saw that the system expected the executable on a specific location and subsequently creating the link fixed it. Of course this could have been incidental and/or specific to my situation at hand, so feel free to leave out that part if you still disagree (I was just trying to pitch-in my two-cents to help this very nice peace of FOSS a bit. Feel free to simply ignore / close my PR if it frustrates/hampers its development instead of helping it forward)!

imsodin:
Your approaching this the wrong way around. If you dont place the binary in the usual location, adjust the systemd unit file (or whatever else you use to start it), dont create a symlink. So if you really think this information is necessary, better state that if you have the binary in a non-standard location, you need to adjust the service file. Personally I dont think this needs to be documented. If you dont install via the "normal way" (package manager) this should be something you know/can determine yourself.

imsodin:
@st-jenkins retest this please

imsodin:
@calmh TeamCity is failing and apparently doesn't care about st-jenkins.

calmh:
Yeah that's an experiment of mine that's not 100% successful and has escaped its cage. I'll rectify it.

calmh:
@st-review merge it

users/autostart: Fix systemd unit path

st-review:
:ok_hand: Merged as 175d100b81fc2a13e71db6a3e6526c4bca2bd2e9. Thanks, @woosting!

PR #320, opened 2017-05-25
Depend on built-in Go environment variables

crazyhatfish:
Users may have set up their Go installation in a different folder (also Go's site suggests a default of `~/go/` for the GOPATH instead of just `~/`) so use Go's built-in environment variables for building.

Secondly, Go's `get` command automatically takes care of setting up the correct directories and calls Git to clone the repository. It just simplifies the procedure slightly.

### DIFF
--- a/dev/building.rst
+++ b/dev/building.rst
@@ -47,18 +47,16 @@ Building (Unix)
     # This should output "go version go1.8" or higher.
     $ go version
 
-    # Go is particular about file locations; use this path unless you know very
-    # well what you're doing.
-    $ mkdir -p ~/src/github.com/syncthing
-    $ cd ~/src/github.com/syncthing
-    # Note that if you are building from a source code archive, you need to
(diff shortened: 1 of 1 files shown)


imsodin:
@st-review lgtm

st-review:
@imsodin: Noted! Need another LGTM or explicit merge command.

calmh:
Only downside here is that Go 1.8+ defaults to `~/go` as `$GOPATH`, meaning
 - `$GOPATH` may no longer be set, and that can be fine
 - We should maybe recommend setting `GOPATH=~/go` if we recommend anything.

crazyhatfish:
That's weird, didn't know they made that default, especially seeing as Go is generally a pretty strict language. Though I agree, changing the default directory to say `~/go` over `~/` would be minimum required change.

calmh:
The build stuff now cares much less about where the source is. The build page should be updated to reflect that, but this is (happily) outdated. :)

PR #324, opened 2017-06-14
Document meaning of default in listen addresses

imsodin:


calmh:
@st-review merge

users/config: Document meaning of default in listen addresses

st-review:
:ok_hand: Merged as 4160d5539907c8bed26643d64074a3be101bcb28. Thanks, @imsodin!

users/config.rst
```diff
@@ -591,6 +591,10 @@ Listen Addresses
 
 The following address types are accepted in sync protocol listen addresses:
 
+Default listen addresses (``default``)
+    This is equivalent to ``tcp://0.0.0.0:22000`` and
```
canton7: What does "and" mean here? How do I put both of these in the listen address box?

AudriusButkevicius:
We should probably state that "as of writing this doc" and point at the source code.

users/config.rst
calmh: The `listenAddress` explanation should state that the element can be present any number of times, but that's sort of unrelated to this improvement.

canton7: My point is, I would expect it to say something like

> This is equivalent to \`\`tcp://0.0.0.0:22000 dynamic+https://relays.syncthing.net/endpoint\`\`

imsodin: The meaning is the same, your variant is in "code/input field" notation while mine is written out in a sentence. The notation should probably be mentioned in the first paragraph, e.g. by writing

    The following address types are accepted in sync protocol listen addresses and multiple addresses can be entered separated by ", ":

calmh: This document is about the config file. There is no comma or space separation here, only tags containing values.

PR #330, opened 2017-07-19
Clarify per-device basis of versioning settings

argotechnica:
This might have helped me avoid a dumb mistake that lead to data loss (not realizing versioning has to be configured on each device)


calmh:
@st-review merge it

users/versioning: Clarify per-device basis of versioning settings

st-review:
:ok_hand: Merged as 62523e1063c6ab57f6a3ccb92eb5647768cc7991. Thanks, @argotechnica!

calmh:
(fwiw there is no setting that is *not* per device, but this doesn't harm so merging anyway)

PR #335, opened 2017-08-15
Update config.rst

bugith:
document defaultFolderPath


users/config.rst
```diff
@@ -584,6 +586,9 @@ tempIndexMinBlocks
     When exchanging index information for incomplete transfers, only take
     into account files that have at least this many blocks.
 
+defaultFolderPath
+    Where you create most of your shared folders.
```
imsodin: Can you mention the (only) effect this has: The UI will propose to create new folders at this path and this can be disabled by setting this to an empty string "".

imsodin:
Old version of #336

PR #339, opened 2017-08-27
Add example of `.stignore`.

doronbehar:
Describe what one can do with `.stignore` to control nested directories
pair.


AudriusButkevicius:
Things that are not supported, should not be in the documentation.

calmh:
Yes. I think nested folders are fine, actually, with the caveats they imply. So maybe we should soften that in the FAQ. But this is not the way.

PR #343, opened 2017-09-04
Updated faq with info about adding new folders

Vium:
I found the folder id field a little bit confusing at first glance and I couldn't find anything in the docs. perhaps this helps.


imsodin:
Thanks for the contribution, that's a nice description. In my opinion our FAQ is already way too crowded, so whenever possible new info should go an appropriate section instead. In this case I would move this to gui.rst into a new section titled "Add Folder".

Vium:
OK, created new pull request for gui.rst

PR #349, opened 2017-09-30
Add caution re mountpoints and .stfolder creation side-effect

### LINKED ISSUE #340: Document behaviour of `-reset-database`
Migrated from https://github.com/syncthing/syncthing/issues/4334

The only documentation in the [wiki](https://docs.syncthing.net/users/syncthing.html?highlight=#cmdoption-reset-database) on `-reset-database` is:

> Reset the database, forcing a full rescan and resync.

[This forum post](https://forum.syncthing.net/t/recreate-synced-folders-after-data-loss/10423/5?u=ataraxy) shows that `.stfolders` are created.

This has implications for unmounted disks so a caveat is probably in order.

[POST /rest/system/reset](https://docs.syncthing.net/rest/system-reset-post.html) may be affected also.

HaleTom:
Partially resolves #340.

### DIFF
--- a/rest/system-reset-post.rst
+++ b/rest/system-reset-post.rst
@@ -7,3 +7,5 @@ By specifying the ``folder`` parameter with a valid folder ID, only
 information for that folder will be erased::
 
 	$ curl -X POST -H "X-API-Key: abc123" http://localhost:8384/rest/system/reset?folder=default
+
+**Caution**: See ``-reset-database`` for ``.stfolder`` creation side-effect and caution regarding mountpoints.


lkwg82:
This hint confuses me. How do I handle this?

HaleTom:
Sorry, handle what? Mounting? Being aware that folders may be created?

imsodin:
At the moment it is just a big cautioning, but only a very terse explanation with a reference to the longer explanation. There should be either a link to the `-reset-database` explanation (my preference) or you should explain in full at this point as well.

PR #350, opened 2017-10-30
Add gnome extension to user contribs

njam:
I was looking for something like this and was wondering why it's not mentioned in the contributions list.
I'm not the author, @jaystrictor is it ok to add it here?

### DIFF
--- a/users/contrib.rst
+++ b/users/contrib.rst
@@ -59,6 +59,13 @@ Mac OS
 
 - `BitBar plugin <https://github.com/sebw/bitbar-plugins>`_
 
+Linux
+~~~~~
+
+- `Syncthing Icon <https://extensions.gnome.org/extension/989/syncthing-icon/>`_
+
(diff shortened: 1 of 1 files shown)


imsodin:
Thanks for the contribution, looks like a neatly done integration. To keep in line with existing headers, I would use "Linux" instead of "Gnome" and then explain shortly what it is (gnome tray integration).

jaystrictor:
@njam Sure, with pleasure.

njam:
@imsodin I force pushed a new version, wdyt?
I'm avoiding the term "tray" because there *was* a concept called "tray" in GNOME Shell, but it was removed in 3.26.

imsodin:
@st-review merge

users/contrib: Add gnome shell extension

st-review:
:ok_hand: Merged as c00776f5e0efbbc4eaa85672747c79d3f41e80bb. Thanks, @njam!

PR #355, opened 2017-11-10
user/config: Update configuration elements

imsodin:
<strike>Do *not review or merge*, I just want this to be out there so that everyone else can use it.

I started updating the config elements but it isn't finished and I don't know when/if I'll finish, so if anyone intends to do the same, feel free to use this as a starting point.</strike>

PR #358, opened 2017-12-04
lib/config: Fix unfinished config elements from #355 (fixes #357)

### LINKED ISSUE #357: Wrong description for STUN options
I was looking for an explanation what the `<stunServer>` option should be set to when I don't want to use any (only IPv6 external connectivity). The documentation on the corresponding options (`<stunKeepaliveSeconds>` as well) just contains some unrelated copy-and-paste residue:

[blob/master/users/config.rst](https://github.com/syncthing/docs/blob/master/users/config.rst#options-element)

imsodin:


AudriusButkevicius:
We should mention that default value gets expanded to those, and that stun keepalive is how frequently we contact the stun server to maintain NAT mapping. Setting it to 0 disables contacting stun servers.

imsodin:
@AudriusButkevicius done. Please edit if the STUN/KCP wording doesn't really fit, I don't know that lingo.

AudriusButkevicius:
@st-review merge

st-review:
:ok_hand: Merged as 6b8e8349d223ef8b503e4d29386096b284fbe4b4. Thanks, @imsodin!

PR #363, opened 2017-12-31
removes reference to xdr

lkwg82:


AudriusButkevicius:
Relays still use XDR, but I guess it's mostly manual these days.

PR #364, opened 2017-12-31
link to build.go

lkwg82:


AudriusButkevicius:
I wonder if we should hardcode the version, yet I guess this is to protect against the fact that changes to the file might move the line.

imsodin:
I don't think linking is the right thing to do in general, as there is no explanation in the build.go file. The human readable information is in this doc, so this doc should be updated instead of adding a reference to code. Anyone who wants/can read code already has that reference implicitly (`go run build.go` instructions).

AudriusButkevicius:
Actually I agree with that.

PR #367, opened 2018-01-03
Be explicit that symlinks are not followed

wikey:
Listing symlinks under things that are "synced" can cause confusion about about what is actually copied over: the link itself or the contents of the linked file. Indicating that symlinks are synced but not followed clarifies this situation and parallels the language used for Hard Links below.


calmh:
@st-review merge

users/faq: Be explicit that symlinks are not followed

st-review:
:ok_hand: Merged as 17f4729dd7563a4a7d5512e9fa6cf7ccfb80048b. Thanks, @wikey!

PR #369, opened 2018-01-12
Describe new discovery server

calmh:
Corresponding to https://github.com/syncthing/syncthing/pull/4648


users/stdiscosrv.rst
```diff
@@ -70,7 +64,7 @@ By default, Syncthing uses a number of global discovery servers, signified by
 the entry ``default`` in the list of discovery servers. To make Syncthing use
 your own instance of stdiscosrv, open up Syncthing's web GUI. Go to settings,
 Global Discovery Server and add stdiscosrv's host address to the comma-separated
-list, e.g. ``https://disco.example.com:8443/v2/``. Note that stdiscosrv uses port
+list, e.g. ``https://disco.example.com:8443/``. Note that stdiscosrv uses port
```
AudriusButkevicius: Why is /v2 removed?

users/stdiscosrv.rst
```diff
@@ -120,28 +111,80 @@ from clients there are three options:
   authenticate the server based on it's certificate and domain name.
 
 - Use any certificate pair and let clients authenticate the server based on
-  it's "device ID" (similar to Syncthing-to-Syncthing authentication). In
-  this case, using ``syncthing -generate`` is a good option to create a
-  certificate pair.
+  it's "device ID" (similar to Syncthing-to-Syncthing authentication). This
+  option can be used with the certificate automatically generated by the
```
imsodin: s/it's/its/
Maybe make Syncthing-to-Syncthing authentication a link to https://docs.syncthing.net/dev/device-ids.html#connection-establishment

imsodin:
From this I don't yet understand how the load balancing happens. The replication is clear, but not how Syncthing clients need to be configured with multiple replicating servers. I assume one needs to establish and point clients at a separate non-Syncthing specific "thing" like you did set up on https://discovery.syncthing.net? If so, a sentence pointing towards that would probably be reasonable.

users/stdiscosrv.rst
calmh: It's not necessary any more with either the existing global servers or just stdiscosrv by it's own, as there is no longer any v1 support. It'll allow /v2 if it comes in, though.

calmh:
@st-review merge

st-review:
:ok_hand: Merged as 4c7b156c12640146689f4e046a30f78578d5e0ed. Thanks, @calmh!

PR #372, opened 2018-01-26
 include/env-vars: Consolidate env vars in separate file and update (fixes #359)

### LINKED ISSUE #359: Clean up deadlock env vars
Reflect changes from syncthing/syncthing#4598 in docs.

It would be great if this information could be generated from main.go, but that's a bigger issue.

imsodin:
I intended to add STRECHECKDBEVERY and remove STDEADLOCK, then saw the todo and did that as well to lower future maintenance. I put the new file with the common environment variables text into a new `includes` dir, as I suspect there is still more potential use for that.


imsodin:
This is lying around idle for quite some time now - anyone time to quickly go over this to discover any blatant blunders on my side (hopefully not) before I just merge this?

calmh:
Lgtm! Sttrace facilities is missing at least fs, walkfs, didn't go through it closely.

PR #376, opened 2018-02-19
docs/users/autostart.rst

moviuro:
Windows > add executable name for the "Task scheduler"

It's just a minor edit, for people not familiar with Windows' utils (like myself)


calmh:
Thanks

PR #377, opened 2018-02-27
Update autostart.rst

schnappijedi:
Added simple method to start Syncthing automatically on Xfce.

### DIFF
--- a/users/autostart.rst
+++ b/users/autostart.rst
@@ -185,7 +185,7 @@ access the GUI type 127.0.0.1:8384 (by default) into Safari.
 Linux
 -----
 
-On Linux systems with a Graphical User Interface/ Desktop Environment.
+On Linux systems with a Graphical User Interface.
 ~~~~~~~~~~~~~~~~~~~~~~
 
 #. Find a program or menu entry similar to 'Startup Applications'.
@@ -193,13 +193,9 @@ On Linux systems with a Graphical User Interface/ Desktop Environment.
 #. Fill out the form:
 
    -  Name: Syncthing
-   -  Command: ``/path/to/syncthing/binary -no-browser -home="/home/your\_user/.config/syncthing"``
-   
-- The "Name:" field can read anything you want. A good choice might be: ``Syncthing``
-  
-- The "Description" field can be left blank or you can add any description you want.
+   -  Command: ``syncthing -no-browser``
 
-- The "Command" field must be edited to coincide with the location of the Syncthing binary and Syncthing data location (default data location is: /home/user/.config/syncthing) on your computer. You can put the Syncthing binary anywhere you want. Remember to set correct permissions especially if outside of home directory. A good location for the Syncthing binary might be: ``/user/home/Syncthing/syncthing``
+The "Command" field must coincide with the location of the Syncthing binary on your computer. ``which syncthing`` can be used to locate the binary. When Syncthing is installed through a package manager the location relative to root is not usually necessary which is why ``syncthing -no-browser`` works. If using a Syncthing binary file directly the command must point to the location of the Syncthing binary realtive to root. Example: ``/home/user/syncthing/syncthing -no-browser``
 
 Using Supervisord
 ~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


imsodin:
I don't think it is useful to have explicit instructions for different desktop environments. I know that we already have an "Ubuntu-like" entry (which should probably be a "Unity" entry). I would propose to instead of having desktop env specific explanations change that to one general entry about "Autostart using a desktop environment", where we essentially just mention that most DEs have an autostart setting and which command to add there.

calmh:
I agree. This is well intended but doesn't scale, and the page as is should be cleaned up.

schnappijedi:
First pull request. Wanted to get a feel for how a pull worked more than anything. Agree with points above. Would organize page like this.

Begin with short discussion of first deciding, regardless of system, if one wants Syncthing to start with a user or with the system and what the difference and implications are. Right now this is touched on in the start as a Windows service section but not mentioned in the systemd section despite being relevant for both (as well as being relevant to the Windows Task Manager startup part).

Specific steps for each method are pretty good right now and sections are organized correctly. Agree Ubuntu section should just be changed like @imsodin suggested. Especially relevant since Ubuntu isn't even using Unity anymore anyways.

All in all though this page is good. If you want to offer guidelines on how you as the developers would like to see it cleaned up would be happy to officially contribute to the most useful program that have ever used.

imsodin:
Great that you want to (already do) contribute to Syncthing's documentation. There's definitely plenty of room to optimize it.

In general what you propose sounds good. One point with regards to starting with system or user and the security implications: On windows it seems like running it with the system also means running it with higher privileges. That's not the case on linux, where you can run Syncthing on system startup as an "unprivileged" user. So if you want to include linux in this security segment, it needs to talk about running as root, not generally running on system startup.

schnappijedi:
Kept it simple. Made simple change of auto-starting on "Ubuntu" to auto-starting on "Linux systems with a Graphical User Interface/ Desktop Environment." Took a few edits to finalize changes. Not sure how to only submit final edit.

users/autostart.rst
```diff
@@ -185,15 +185,21 @@ access the GUI type 127.0.0.1:8384 (by default) into Safari.
 Linux
 -----
 
-On Ubuntu-like systems
+On Linux systems with a Graphical User Interface/ Desktop Environment.
```
imsodin: Simplify to the same style as other titles: "Using Desktop Environment"?

schnappijedi:
Was operating under assumption of creating direction for autostarting the binary, not Syncthing installed from any given distribution package manager. Good point as most looking at these directions will likely be using package manager.

What do you think? Solely touch on autostarting assuming using a package manager or divide into two sets of Linux Desktop autostart directions, one for binary one for Syncthing downloaded through package manager?

On a related note when Syncthing is downloaded through a package manager is the "Automatic upgrade" through the GUI  usually disabled/ removed by default since the package manger is doing the updates?

The -home option should have been removed. It was left over from the prior Ubuntu Unity directions.

imsodin:
I'd keep it as simple as possible, e.g. keep your first sentence and maybe add the info how to find the binary:
```
The "Command" field must be edited to coincide with the location of the Syncthing binary. If you don't know where that is, you may find it with the command ``which syncthing``. 
```
That applies regardless of whether you download the binary manually or use a package manager.

And the automatic upgrade is already disable in that case. There is a build flag to disable automatic upgrades.

imsodin:
I'd still make everything much more generic, i.e. remove all references to fields, forms, ... and don't use numerated steps. Basically just mention the option of using the DE's autostart menu and giving and explaining the command to use. Also I would always recommend the absolute path - you never know whether some DE doesn't care about $PATH for autostart commands (that would be somewhat ridiculous, but better be safe). And I'm still in favor of just "Using Desktop Environments" as title.

schnappijedi:
Never got email about reply. Looks like the changes we were all looking for were made. Will propose a  small change and then this can be closed whether accepted or not.

Is there a way to bring my fork up to date before proposing small change or does a new pull request have to be made to accomplish this?

PR #378, opened 2018-03-02
And KCP listener documentation with static relay

camAtGitHub:
Existing documentation talks about how to use a private relay but it is not clear that in doing it that way it then removes the TCP and KCP listeners, resulting in poor performance on local LAN segments etc. This commit attempts to clarify that.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -739,10 +739,16 @@ TCP IPv6 wildcard and port (``tcp6://[::]:22000``, ``tcp6://:22000``)
 TCP IPv6 address and port (``tcp6://[2001:db8::42]:22000``)
     This results in Syncthing listening on the specified address and port, IPv6
     only.
+    
+KCP listener address (``kcp://0.0.0.0:22000``, ``kcp://:22020``)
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Kcp support was reverted as it caused more issues than benefits.

camAtGitHub:
Hi Audrius, would you like me to change the doco? 
I certainly spent hours trying to find out how to make my private relay work with syncthing normal protocols. Clarifying the ability to join statements with commas would be good.

AudriusButkevicius:
Your change is fine, just remove kcp. Also, it might be nice to have "default relay://..." also as an example.

imsodin:
If you go into combinations, please also state the difference between web UI and config (xml): In the UI you can combine by ", " but in the config (xml) you need multiple listenAdress elements.

AudriusButkevicius:
The dynamic+ was correct, @imsodin just meant that we should mention that the format is different in the config (not space separated strings in a single xml element, but multiple xml elements with a single string each)

xor-gate:
Didn't become KCP feature obsolete?

PR #379, opened 2018-03-07
users: Update scanning explanation with watcher

imsodin:
Linked with syncthing/syncthing#4789


imsodin:
The other PR is merged now, anyone want to merge this (@AudriusButkevicius @calmh)? Users seem to be jumping on the watcher train and the first asked about documentation. If there are things to change, you need to do it yourself or wait till about 9. April or so :)

PR #381, opened 2018-03-22
users/faq: Add entry about increasing inotify watches

imsodin:
This integrates https://github.com/syncthing/syncthing-inotify/#troubleshooting-for-folders-with-many-files-on-linux into our own documentation to be linked to in the appropriate error message (syncthing/syncthing#4833).

### DIFF
--- a/users/faq.rst
+++ b/users/faq.rst
@@ -515,10 +515,10 @@ On many Linux distributions you can run the following to fix it::
     echo "fs.inotify.max_user_watches=204800" | sudo tee -a /etc/sysctl.conf
 
 On Arch Linux and potentially others it is preferred to write this line into a
-separate file, i.e., you should run::
+separate file, i.e. you should run::
 
     echo "fs.inotify.max_user_watches=204800" | sudo tee -a /etc/sysctl.d/90-override.conf
 
-This only takes effect after a reboot. To adjust the limit immediately, run ::
+This only takes effect after a reboot. To adjust the limit immediately, run::
 
     sudo sh -c 'echo 204800 > /proc/sys/fs/inotify/max_user_watches'
--- a/users/faq.rst
+++ b/users/faq.rst
@@ -497,25 +497,25 @@ runit.
 
 .. _inotify-limits:
 
-How do I increase the inotify limit to get my fs watcher to work?
------------------------------------------------------------------
+How do I increase the inotify limit to get my filesystem watcher to work?
+-------------------------------------------------------------------------
 
 You are probably reading this because you encountered the following error with
-the fs watcher on linux:
+the filesystem watcher on linux:
 
     Failed to start filesystem watcher for folder yourLabel (yourID): failed to
     setup inotify handler. Please increase inotify limits, see
     https://docs.syncthing.net/users/faq.html#inotify-limits
 
-Linux typically restricts the amount of watches per user (usually 8192). So if
-you have a more directories, you need to adjust that number.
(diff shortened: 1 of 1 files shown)


users/faq.rst
```diff
@@ -494,3 +494,31 @@ to do this.  The most well known is called daemontools, and can be found in the
...
+
+You are probably reading this because you encountered the following error with
+the fs watcher on linux:
+
+    Failed to start filesystem watcher for folder yourLabel (yourID): failed to
+    setup inotify handler. Please increase inotify limits, see
+    https://docs.syncthing.net/users/faq.html#inotify-limits
+
+Linux typically restricts the amount of watches per user (usually 8192). So if
+you have a more directories, you need to adjust that number.
```
calmh: s/So if/When/

users/faq.rst
```diff
@@ -494,3 +494,31 @@ to do this.  The most well known is called daemontools, and can be found in the
...
+
+On many Linux distributions you can run the following to fix it: ::
+
+    echo "fs.inotify.max_user_watches=204800" | sudo tee -a /etc/sysctl.conf
+
+On Arch Linux and potentially others it is preferred to write this line into a
+separate file, i.e. you should run ::
```
calmh: Here, and above/below, you can end the line with just the double colon to both show a normal colon and make the following paragraph a code block. That is,

`... you should run::`

becomes

screen shot 2018-03-23 at 13 27 20

calmh: comma after "i.e.", i.e., `... separate file, i.e., you should ...`

(minor :trollface:)

users/faq.rst
```diff
@@ -494,3 +494,31 @@ to do this.  The most well known is called daemontools, and can be found in the
 standard package repositories for  almost every modern Linux distribution.
 Other popular tools with similar functionality include S6 and the aforementioned
 runit.
+
+.. _inotify-limits:
+
+How do I increase the inotify limit to get my fs watcher to work?
```
calmh: I don't think "fs" is established enough to warrant a lower case abbreviation. `FS watcher" or "filesystem watcher"?

users/faq.rst
imsodin: I could argue that this is a British vs American thing and I learned most English reading Tolkien, but truth told it just hurts my German-biased language sense (there's never ever a comma after the equivalent "d.h.") - but I digress :)  
Oh and as I am wasting time already:  
```
simon@simon-x220-deb:~/Coding/syncthing-docs$ grep -r "i\.e\.[^,]" | wc -l
105
```
I am not going to waste more time to fix these as well :P

calmh: You're right, leave this one alone!

https://english.stackexchange.com/questions/16172/should-i-always-use-a-comma-after-e-g-or-i-e

users/faq.rst
```diff
@@ -494,3 +494,31 @@ to do this.  The most well known is called daemontools, and can be found in the
...
+On many Linux distributions you can run the following to fix it::
+
+    echo "fs.inotify.max_user_watches=204800" | sudo tee -a /etc/sysctl.conf
+
+On Arch Linux and potentially others it is preferred to write this line into a
+separate file, i.e., you should run::
+
+    echo "fs.inotify.max_user_watches=204800" | sudo tee -a /etc/sysctl.d/90-override.conf
+
+This only takes effect after a reboot. To adjust the limit immediately, run ::
```
calmh: space before :: to be removed :)

PR #387, opened 2018-05-12
Adjust folder type in accordance with syncthing/syncthing#4942

Martchus:


imsodin:
Thanks! Following are some further notes - if you want to take care of them, then great, otherwise this is a follow-up todo list:  
Grepping turned up some more results for read/write-only:
```
./users/config.rst
./intro/getting-started.rst
./rest/system-config-get.rst
./rest/svc-report-get.rst
```
In the config specs (or config post rest description) there should probably also be a legacy notice with old and new names and a note, that old ones are currently still supported, but will be deprecated at some point.

Martchus:
I updated further occurrences except the log in `intro/getting-started.rst` because it also has the Syncthing version at the beginning which would make it a bit weird.

I added deprecation notes as well.

users/config.rst
```diff
@@ -206,12 +206,16 @@ path
 type
     Controls how the folder is handled by Syncthing. Possible values are:
 
-    readwrite
+    sendreceive
         The folder is in default mode. Sending local and accepting remote changes.
+        Note that this type was previously called "readwrite" which is deprecated
```
imsodin: Sorry, this flew under my radar. However I'd anyway only merge it once 0.14.49-rc.1 is released, so no rush.

I'd either put the deprecation note in its own paragraph or maybe even use the deprecated directive, something like this

```
    sendonly
        The folder is in "send-only" mode -- it will not be modified by
        Syncthing on this device.
        
        readonly
            .. deprecated:: v0.14.49

            This is the old name of this type, which is still accepted in
            incoming configs.
```

PR #396, opened 2018-07-03
events: Add missing name property for DeviceRejected

xor-gate:
It seems the documentation was missing a property for the DeviceRejected event which is implemented here https://github.com/syncthing/syncthing/blob/7b0d8c2e77750caa335aa3216dffcff4ffeb0115/lib/model/model.go#L1524.

### DIFF
--- a/events/devicerejected.rst
+++ b/events/devicerejected.rst
@@ -12,6 +12,7 @@ to talk to.
         "time": "2014-08-19T10:43:00.562821045+02:00",
         "data": {
             "address": "127.0.0.1:51807",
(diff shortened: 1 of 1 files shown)


events/devicerejected.rst
```diff
@@ -12,6 +12,7 @@ to talk to.
         "time": "2014-08-19T10:43:00.562821045+02:00",
         "data": {
             "address": "127.0.0.1:51807",
+            "name" : "My dusty computer",
```
AudriusButkevicius: No space prior colon

PR #402, opened 2018-07-11
specify localhost in nginx reverse proxy conf

### LINKED ISSUE #401: Nginx reverse proxy config should explicitly set 'proxy_set_header Host localhost'
As per #380 (but for Nginx), the Nginx reverse proxy config snippet should explicitly note that `proxy_set_header Host localhost` is (or at least may be) required. 

Whilst `$host` may work in some cases, it won't work in all (see Nginx docs on [`$host`](http://nginx.org/en/docs/http/ngx_http_core_module.html#var_host)).

I was pulling my hair out until I came across [this blog](https://f-a.nz/dev/setup-syncthing-on-debian-server-with-nginx-reverse-proxy/). As suggested adding `<insecureSkipHostcheck>true</insecureSkipHostcheck>` (within the `<gui .. > </gui>` section of the Syncthing `config.xml`) worked around it but it felt a bit dirty...

Then I saw this: https://github.com/syncthing/syncthing/issues/3573#issuecomment-245411714

Problem solved!

JedMeister:
closes #401

### DIFF
--- a/users/reverseproxy.rst
+++ b/users/reverseproxy.rst
@@ -38,7 +38,7 @@ Nginx
 .. code-block:: nginx
 
     location /syncthing/ {
-      proxy_set_header        Host $host;
+      proxy_set_header        Host localhost;
       proxy_set_header        X-Real-IP $remote_addr;
       proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
       proxy_set_header        X-Forwarded-Proto $scheme;


calmh:
This is the wrong approach, imho. It tells Syncthing that the request is coming from localhost and can be treated with the security implications of that, which is a lie. Lying to Syncthing in this way is dangerous. If you want to live dangerously there are appropriately named options to set Syncthing-side.

JedMeister:
@calmh - Thanks for the clarity. 

As it stands though, the current "suggested" reverse proxy config in the docs doesn't work - Syncthing gives a `host check error`. IMO that's not very useful...

If having Syncthing behind a reverse proxy is undesirable, then perhaps that could be clarified in the docs? Do you have any suggested wording for that? I'm happy to try again on a PR for the docs.

I think it should also be noted, that if you do need to set up a reverse proxy, that `<insecureSkipHostcheck>true</insecureSkipHostcheck>` (within the `<gui .. > </gui>` section of the Syncthing `config.xml`) is required. Would you be happy for that to also be added?

JedMeister:
As I just noted in the (now closed) issue, the other option is to not bind Syncthing to localhost. Although IMO that makes the idea of a reverse proxy somewhat redundant. Although it's quite possible I'm missing something...?!

AudriusButkevicius:
I think the right option is to add a section explaining what settings to set in syncthing and how.

JedMeister:
Ok I'll have a crack at that when I have time. As noted in the issue, I'd just like some clarification on what the security implications are first. Thanks in advance.

AudriusButkevicius:
It might be in the docs for the config, if not, I'd have to remember what the problem is, but I think it's todo with DNS rebinding attacks.

JedMeister:
Thanks. I'll have a bit of a read and see what I can find.

JedMeister:
So would I be right to add something like the below to the docs? I'm thinking at the bottom of the existing [Server Configuration](https://github.com/syncthing/docs/blob/master/users/reverseproxy.rst#server-configuration) section:
```
When reverse proxying Syncthing, best practice dictates that only TLS connections
should be allowed between the reverse proxy and the user/internet.

Note that if you choose to also bind Syncthing to localhost, Syncthing's host
security checks will need to be disabled. This can be done by enabling 
``insecureSkipHostcheck``. I.e. add to the ``<gui>`` section of ``conf.xml``:

.. code-block:: none

    <insecureSkipHostcheck>true</insecureSkipHostcheck>
```
If you're not happy with that, do you have any suggestions to make is acceptable?

AudriusButkevicius:
We should just say that "By default you will have to do xyx", suggest that it would be done via the web ui advanced config, and potentially link to FAQ.

PR #411, opened 2018-08-09
Added /v2/ for discovery server to docs

cromefire:
It's not documented and if you forget it it will serve 404s

### DIFF
--- a/users/stdiscosrv.rst
+++ b/users/stdiscosrv.rst
@@ -67,7 +67,7 @@ Global Discovery Server and add stdiscosrv's host address to the comma-separated
 list, e.g. ``https://disco.example.com:8443/v2/``. Note that stdiscosrv uses port
 8443 by default. For stdiscosrv to be available over the internet with a dynamic
 IP address, you will need a dynamic DNS service. Do not forget the trailing ``/``
-of ``/v2/``
+of ``/v2/``.
 
 If you wish to use *only* your own discovery server, remove the ``default``
 entry from the list.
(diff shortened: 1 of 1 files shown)


cromefire:
Ref syncthing/syncthing#5107

imsodin:
You are probably using an old version of stdiscosrv, in the up-to-date one you don't need v2. See https://github.com/syncthing/docs/pull/369

cromefire:
@imsodin It's the `0.14.43`, the stable for deb. It may be worth to mention it at least if you get 404s

AudriusButkevicius:
Thats fairly recent...

AudriusButkevicius:
Perhaps they look for the code in some old repo or something, the code clearly does not talk about v2 anymore.

imsodin:
It's one version too old, the change was introduced in 0.14.44: https://github.com/syncthing/discosrv/releases/tag/v0.14.44 No idea why ubuntu packages this version.

cromefire:
Maybe you should look into build system, it's `Ubuntu 18.04` (don't think that's important), Debian package `syncthing-dicosrv`.

I think there should be a hint in the docs nonetheless.
I'd think something like:

> If your client is getting 404s you can try to append `/v2/` to the address (the last `/` is important).

Does this sound okay for you?

AudriusButkevicius:
Ww should add a note that versions prior to xxx need /v2/ appended to the discovery server address.

cromefire:
Shall I do it or are you doing it?

AudriusButkevicius:
Pls do.

PR #412, opened 2018-08-10
Mentioned that older versions may need /v2/

cromefire:
See #411

### DIFF
--- a/users/stdiscosrv.rst
+++ b/users/stdiscosrv.rst
@@ -68,8 +68,8 @@ list, e.g. ``https://disco.example.com:8443/``. Note that stdiscosrv uses port
 8443 by default. For stdiscosrv to be available over the internet with a dynamic
 IP address, you will need a dynamic DNS service.
 
-.. warning::
-   Versions prior to ``0.14.44`` may need ``/v2/`` appended to the discovery
+.. deprecated:: v0.14.44
+   Prior version may need ``/v2/`` appended to the discovery
    server address, e.g. ``https://disco.example.com:8443/v2/``.
 
 If you wish to use *only* your own discovery server, remove the ``default``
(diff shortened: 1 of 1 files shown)


users/stdiscosrv.rst
```diff
@@ -68,6 +68,10 @@ list, e.g. ``https://disco.example.com:8443/``. Note that stdiscosrv uses port
 8443 by default. For stdiscosrv to be available over the internet with a dynamic
 IP address, you will need a dynamic DNS service.
 
+.. warning::
+   Versions prior to ``0.14.44`` may need ``/v2/`` appended to the discovery
+   server address, e.g. ``https://disco.example.com:8443/v2/``.
```
imsodin: I'd use `.. deprecated:: v0.14.44` and remove the may, something like 
> Prior versions needed ``v2/`` appended to the discovery  server address, e.g. ``https://disco.example.com:8443/v2/``."

cromefire: I think depending on you version (for example if it's even older) it may no be needed.

AudriusButkevicius:
Lets get rid of the may, discovery server versions prior to tuat probably don't exist.

cromefire:
There you go

users/stdiscosrv.rst
```diff
@@ -68,6 +68,10 @@ list, e.g. ``https://disco.example.com:8443/``. Note that stdiscosrv uses port
 8443 by default. For stdiscosrv to be available over the internet with a dynamic
 IP address, you will need a dynamic DNS service.
 
+.. deprecated:: v0.14.44
+   Prior version need ``/v2/`` appended to the discovery
```
imsodin: I feel a bit silly about the amount of review going into such a little change (thanks already for your patience), but could do version->versions.

cromefire: It would have saved me 6 hours of time so it's relatively important I think, but it okay, I know that

cromefire:
All fixed

PR #413, opened 2018-08-10
doc update: default is not HTTPS

splitbrain:
When installing syncthing for the very first time, the GUI is not starting on SSL. This adjusts the Getting Started chapter accordingly.


calmh:
It’s available on https, but a new user will probably be less confused by standard http.

PR #414, opened 2018-08-13
users/faq: Reword section on GUI listening

ProactiveServices:
The config.xml example no longer leaves TLS disabled and with no password. Reworded to make these two settings part of the norm, rather than a footnote.

### DIFF
--- a/users/faq.rst
+++ b/users/faq.rst
@@ -305,25 +305,24 @@ This is an area that we are working to improve in the long term.
 How do I access the web GUI from another computer?
 --------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


calmh:
It's not reasonable to expect the user to manually calculate a bcrypt hash though. This section is specifically about being able to get to the interface from outside...

ProactiveServices:
Yeah, good point.

calmh:
Now the rewording isn't correct, as it says the config changes will enable passwords and stuff. The old variant mentioned after the config changes that this *should* be done...

st-review:
:robot: *beep boop*

PR #415, opened 2018-08-20
Update Syncthing-Android maintainers

Nutomic:
I just noticed that the list of maintainers was out of date, as I am currently not active as maintainer, and afaik Zillode hasn't done anything on the project in quite a while.

@Catfriend1 I didn't see a name on your profile, so I assume you want to be pseudonymous.

By the way, you might want to add other projects that are linked from the website, like SyncTrayzor or syncthing-macos.

### DIFF
--- a/intro/project-presentation.rst
+++ b/intro/project-presentation.rst
@@ -22,7 +22,6 @@ syncthing-android
 `syncthing-android`_ is the Android packaging and native UI on top of
 Syncthing.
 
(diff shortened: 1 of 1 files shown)


calmh:
I think one of the two just ejected, unfortunately. Otherwise sound in principle.

Catfriend1:
Thanks @Nutomic You did a great work on syncthing-android!

PR #426, opened 2018-10-16
dev/building: Recommend Go 1.11

nstapelbroek:
Heya :wave: 

I just started to look into syncthing. While reading the docs I noticed that the Docker builder was using 1.11 while the docs describe go1.9 is preferred. This PR should update the docs telling that the latest stable go1.11 is prefered. 

I don't know how to view the go version of the TeamCity builds so I could be wrong. :see_no_evil:


calmh:
Clean change, thanks, sorry this took so long.

PR #430, opened 2019-01-01
Update for Syncthing v1.0.0

imsodin:
The chosen example versions and scenarios might no longer be ideal, we should update them when the discussion on the new versioning policy has come to a conclusion.

I mostly just want to make a new release so our documentation version matches the "real" one.


calmh:
Maybe also config.py for the 0.14 that is in the footer etc?

PR #431, opened 2019-01-07
Reduced Ambiguity in "Why does it use…"

Thovthe:
If this is accurate, I think it's valuable to spell it out clearly for anyone who may be concerned about security but also too (insert reason here) to confirm.

### DIFF
--- a/users/faq.rst
+++ b/users/faq.rst
@@ -115,8 +115,9 @@ Why does it use so much CPU?
 #. When new or changed files are detected, or Syncthing starts for the
    first time, your files are hashed using SHA-256.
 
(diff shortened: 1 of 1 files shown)


Thovthe:
Not sure why anything is breaking, simply a markdown document change.

canton7:
Don't think it's you.

```
10:53:22 PM: /usr/local/bin/build: line 32: sphinx-build: command not found
```

@calmh?

AudriusButkevicius:
Yet the change has typos in the word always and I think it makes the sentence harder to process

calmh:
Merged with tweaks, thanks.

PR #432, opened 2019-01-13
Readability improvement on "Introducer" documentation. 

shaomeng:


calmh:
It's still crap, but it is a slight improvement :)

PR #435, opened 2019-01-26
added FAQ - How do I reset the GUI password?

ken-marsh-abb:


users/faq.rst
```diff
@@ -524,3 +524,21 @@ separate file, i.e. you should run::
...
+| <gui enabled="true" tls="false" debugging="false">
+|    <address>127.0.0.1:8384</address>
+|    **<user>syncguy</user>**
+|    **<password>$2a$10$s9wWHOQetp46Cq7GPye69.KqKantdUrbHCeLnZ9A6VlzkEPPYHDAC</password>**
+|    <apikey>9RCKohqCAyrj5RjpyZdR2wXmQ9PyQFeN</apikey>
+|    <theme>default</theme>
+| </gui>
+|
+
+After this the GUI can be accessed without authentication. No restart of Syncthing or refresh of the browser is *required*, but for consistent results it is recommended.
```
AudriusButkevicius: I think this is not true, as it would only work if the config is reloaded. Also, syncthing might decide to save the config overwriting changes you've made manually, so this operation should be done while syncthing is not running.

ken-marsh-abb: I was surprised myself to see the change take effect without restart, it's possible I misled myself into only thinking that's what I did. It is probably wise to stop Syncthing as standard procedure anyway, whether necessary or not. I'll run through the procedure a couple times and rewrite.

PR #437, opened 2019-01-29
Added FAQ about how to reset GUI password if forgotten/lost

ken-marsh-abb:
2nd revision - restart required, edit while not running recommended.

### DIFF
--- a/users/faq.rst
+++ b/users/faq.rst
@@ -528,7 +528,7 @@ This only takes effect after a reboot. To adjust the limit immediately, run::
 How do I reset the GUI password?
 --------------------------------
 
-If you've forgotten/lost the GUI password, you can reset it to blank by deleting the **<user>** and **<password>** XML tags from the **<gui>** block in file ``config.xml``. The location of the file depends on OS and is described in the configuration documentation.
+If you've forgotten/lost the GUI password, you can remove it by deleting the **<user>** and **<password>** XML tags from the **<gui>** block in file ``config.xml``. This should be done while Syncthing is not running. The location of the file depends on OS and is described in the configuration documentation.
 
 For example the two lines **in bold** below would be removed from the file.
 
@@ -540,7 +540,3 @@ For example the two lines **in bold** below would be removed from the file.
 |    <theme>default</theme>
 | </gui>
 |
-
-Restarting Syncthing is required for the change to take effect. To avoid potentially confusing conflicts it is recommended to make the change while Syncthing is not running.
-
-After this the GUI can be accessed without authentication and password reconfigured using normal procedure in the GUI.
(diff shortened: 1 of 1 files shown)


users/faq.rst
```diff
@@ -524,3 +524,23 @@ separate file, i.e. you should run::
 This only takes effect after a reboot. To adjust the limit immediately, run::
 
     sudo sh -c 'echo 204800 > /proc/sys/fs/inotify/max_user_watches'
+
+How do I reset the GUI password?
+--------------------------------
+
+If you've forgotten/lost the GUI password, you can reset it to blank by deleting the **<user>** and **<password>** XML tags from the **<gui>** block in file ``config.xml``. The location of the file depends on OS and is described in the configuration documentation.
```
AudriusButkevicius: "you can remove it" rather than reset it to blank.

users/faq.rst
```diff
@@ -524,3 +524,23 @@ separate file, i.e. you should run::
...
+| <gui enabled="true" tls="false" debugging="false">
+|    <address>127.0.0.1:8384</address>
+|    **<user>syncguy</user>**
+|    **<password>$2a$10$s9wWHOQetp46Cq7GPye69.KqKantdUrbHCeLnZ9A6VlzkEPPYHDAC</password>**
+|    <apikey>9RCKohqCAyrj5RjpyZdR2wXmQ9PyQFeN</apikey>
+|    <theme>default</theme>
+| </gui>
+|
+
+Restarting Syncthing is required for the change to take effect. To avoid potentially confusing conflicts it is recommended to make the change while Syncthing is not running.
```
AudriusButkevicius: I think we should say this operation should be done while syncthing is not running which then removes the need for talking about the restarts.

users/faq.rst
```diff
@@ -524,3 +524,23 @@ separate file, i.e. you should run::
...
+
+After this the GUI can be accessed without authentication and password reconfigured using normal procedure in the GUI.
```
AudriusButkevicius: This extra sentence does not add much to the explanation, so perhaps can be dropped?

users/faq.rst
```diff
@@ -524,3 +524,23 @@ separate file, i.e. you should run::
...
+For example the two lines **in bold** below would be removed from the file.
+
+| <gui enabled="true" tls="false" debugging="false">
+|    <address>127.0.0.1:8384</address>
+|    **<user>syncguy</user>**
+|    **<password>$2a$10$s9wWHOQetp46Cq7GPye69.KqKantdUrbHCeLnZ9A6VlzkEPPYHDAC</password>**
+|    <apikey>9RCKohqCAyrj5RjpyZdR2wXmQ9PyQFeN</apikey>
+|    <theme>default</theme>
+| </gui>
+|
```
imsodin: In a section of the FAQ above we have used 
```

.. code-block:: xml

    <gui enabled="true" tls="false">
```

So for consistency I'd do the same here.

ken-marsh-abb: I considered that but wasn't easily able to make lines **bold** inside a code block.

PR #439, opened 2019-02-28
users/faq: Add nuance to nested folder statement (fixes #428)

### LINKED ISSUE #428: [FAQ] Nested shared folder safe with exclusion rule?
https://docs.syncthing.net/users/faq.html#am-i-able-to-nest-shared-folders-in-syncthing

> 2.2.12. Am I able to nest shared folders in Syncthing?
> Do not share a folder which is inside another shared folder. This behaviour is in no way supported, recommended or coded for in any way, and comes with many pitfalls.

I think the wording could be more nuanced here, because if the parent shared folder `/parent` has an exclusion rule for `/parent/child`, then I think it is safe to also have `/parent/child` as a separate shared folder.

Or I am missing something?

Edit: I am running a similar setup for more than a week now, and have not encountered any issues.

imsodin:
My intention was to keep it not recommended, but make it clear that it can be done and mention the one know catastrophic case (circular includes) - I don't think explaining any further is useful, if the user doesn't understand what circular includes are, they just shouldn't do nested folders (or do research aka read the forum).

### DIFF
--- a/users/faq.rst
+++ b/users/faq.rst
@@ -206,9 +206,11 @@ See the previous question.
 Am I able to nest shared folders in Syncthing?
 ----------------------------------------------
 
-Do not share a folder which is inside another shared folder. This behaviour
-is in no way supported, recommended or coded for in any way, and comes with
-many pitfalls.
+Sharing a folder that is within an already shared folder is possible, but it has
+its caveats. What you must absolutely avoid are circular shares. This is just
+one example, there may be other undesired effects. Nesting shared folders is not
+supported, recommended or coded for, but it can be done successfully when you
+know what you're doing - you have been warned.
 
 How do I rename/move a synced folder?
 -------------------------------------


AudriusButkevicius:
It doesn't explain what that circular thing is. There are other caveats seeing your own changes as conflicts etc.

thomas-zmo08179:
@imsodin , if you say "it can be done successfully when you know what you're doing", what would you recommend? 
DeviceA <-----> DeviceB <------> DeviceC

imsodin:
> It doesn't explain what that circular thing is.

First comment:  
> I don't think explaining any further is useful, if the user doesn't understand what circular includes are, they just shouldn't do nested folders (or do research aka read the forum).

Do you disagree with that?

> There are other caveats seeing your own changes as conflicts etc.

PR:  
> This is just one example, there may be other undesired effects.

I won't try and list all possible caveats -> should I not mention circular includes at all?

> @imsodin , if you say "it can be done successfully when you know what you're doing", what would you recommend?

If you read the rest, I don't really want to do recommendations. Most certainly not on the github issue tracker, the forum would be the appropriate place for that.

calmh:
Honestly this is almost more scary than what was already there :D

AudriusButkevicius:
I think before it was just don't do it, now it's don't do it because it's beyond what you peasant can understand.

PR #444, opened 2019-03-16
Document how to fetch dependencies.

tomfitzhenry:


calmh:
No, this is not necessary, and in fact does not work with the rest of the build instructions (using Go modules).

PR #446, opened 2019-03-24
Note the performance impact of negative patterns.

sa3dany:
Add a note about the performance impact of negative patterns on scans.

## Example
Starting with this ignore pattern:
```
// Ignore everything
*
```

Scanning a folder is instant. Now add any negative patterns:
```
// Includes
!/foo

// Ignore everything
*
```

Now it feels like Syncthing is scanning entire tree.

I also found this [post](https://forum.syncthing.net/t/looking-for-ways-to-speed-up-scanning-on-large-folders/12717/2) by Jakob Borg on the Syncthing forum:

> Ignoring files won’t help. If you can ignore large parts of the tree, like directories at the top level, that would help as there would be less to scan. Don’t use any negative patterns (beginning with !) as Syncthing will then need to do a full scan regardless of the other ignore patterns.

### DIFF
--- a/users/ignoring.rst
+++ b/users/ignoring.rst
@@ -83,6 +83,11 @@ The ``.stignore`` file contains a list of file or path patterns. The
 
    Prefixes can be specified in any order (e.g. "(?d)(?i)"), but cannot be in a
    single pair of parentheses (not ":strike:`(?di)`").
+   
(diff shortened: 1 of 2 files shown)


sa3dany:
Done and thank you for the review.

PR #453, opened 2019-05-31
Fix the unknown variable error when using nginx

misaka00251:
The example config file of Nginx contain multiple errors:

```
proxy_set_header Connection $proxy_connection;
proxy_set_header X-Forwarded-Proto $proxy_x_forwarded_proto;
```

When added these lines, Nginx will output something like this:

```
nginx: [emerg] unknown "proxy_connection" variable
```

Tested environment: 

```
Nginx version: nginx/1.14.1
```

### DIFF
--- a/users/stdiscosrv.rst
+++ b/users/stdiscosrv.rst
@@ -242,7 +242,7 @@ the Syncthing settings.
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection $http_connection;
     proxy_set_header X-Real-IP $remote_addr;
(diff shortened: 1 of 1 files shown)


canton7:
I can't speak for the other bits, but `proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;` works just fine for me. [See the docs](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#variables). I can't find any docs for `$http_add_x_forwarded_for`?

misaka00251:
Sorry, my fault. I didn't check the docs of that.

PR #456, opened 2019-07-08
rest/system-config-post: Correct outdated info

imsodin:


calmh:
The right answer is to check the system/config/insync endpoint after the post request succeeds. If that returns true there is no need to restart. Otherwise, there is.

imsodin:
Ah thanks, wasn't aware of that!

PR #457, opened 2019-07-10
Update system-config-post.rst

vincentardern:
Fixed grammar in second sentence, and possibly reduced ambiguity.


AudriusButkevicius:
Should be succeeds, and directly is a bad word, ir should be immediately applied etc.

vincentardern:
Sounds good to me.

AudriusButkevicius:
Thanks

PR #460, opened 2019-07-16
Document -umask option.

DanielO:


AudriusButkevicius:
Is there a thing like that? Why can't you set the umask in the shell before starting?

DanielO:
I did try that but the FreeBSD rc.d script already has 2 lots of escaping to change UID then wrap syncthing with daemon (to write a PID file and redirect logging) that it was simpler to modify the syncthing code..

Arguably daemon could grow a umask option but that's a lot slower to change :)

AudriusButkevicius:
I think we should discuss it on the actual or. My bad on discovering this first.

DanielO:
OK, sure.

PR #461, opened 2019-07-16
Allow SPHINXBUILD to be read from the environment.

DanielO:
Without this the Sphinx path is not read from the environment so the suggestion in the error message doesn't work.


calmh:
Thanks!

PR #464, opened 2019-08-06
Document QUIC transport in listening addresses

olifre:
Also mention that multiple listening addresses can be specified by comma.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -792,11 +792,12 @@ setLowPriority
 Listen Addresses
 ^^^^^^^^^^^^^^^^
 
-The following address types are accepted in sync protocol listen addresses:
+The following address types are accepted in sync protocol listen addresses
+(multiple can be specified by comma separation):
 
 Default listen addresses (``default``)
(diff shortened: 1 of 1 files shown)


users/config.rst
```diff
@@ -792,11 +792,12 @@ setLowPriority
 Listen Addresses
 ^^^^^^^^^^^^^^^^
 
-The following address types are accepted in sync protocol listen addresses:
+The following address types are accepted in sync protocol listen addresses
+(multiple can be specified by comma separation):
```
AudriusButkevicius: I think its space separated? Or could be either

olifre: In the code here:
https://github.com/syncthing/syncthing/blob/3cb7b8f22bf254a5270acdce52686863979cf35d/gui/app.js#L52
and here:
https://github.com/syncthing/syncthing/blob/3cb7b8f22bf254a5270acdce52686863979cf35d/gui/app.js#L174
I think (if my understanding is correct) that comma-separated is the only accepted format, at least in the GUI.

AudriusButkevicius: You are right.

PR #465, opened 2019-08-17
Changed Supervisor related docs to recommend using separate config files

MarcinOrlowski:
Changed supervisor related docs to recommend creating separate config file instead of messing with main one.

### DIFF
--- a/users/autostart.rst
+++ b/users/autostart.rst
@@ -197,22 +197,20 @@ On any distribution (Arch, Debian, Linux Mint, Ubuntu, openSUSE)
 
 Using Supervisord
 ~~~~~~~~~~~~~~~~~
-
-Create the ``syncthing.conf`` file in supervisor's ``conf.d`` folder (the folder is located
-at ``/etc/supervisor/conf.d``)::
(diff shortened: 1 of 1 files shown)


users/autostart.rst
```diff
@@ -198,16 +198,27 @@ On any distribution (Arch, Debian, Linux Mint, Ubuntu, openSUSE)
 Using Supervisord
 ~~~~~~~~~~~~~~~~~
 
-Add the following to your supervisor config file::
+Create the ``syncthing.conf`` file in supervisor's ``conf.d`` folder (the folder is located
+at ``/etc/supervisor/conf.d``)::
```
AudriusButkevicius: In supervisors x folder which is actually /etc/foo/x, why not just /etc/foo/x straight away?

AudriusButkevicius: Is it different between ubuntu/debian?

MarcinOrlowski: Not sure I understand your question. Debian/Ubuntu installation uses `conf.d` subfolder as default location for custom configs which helps keeping all stuff organized.

AudriusButkevicius: Sure, as long as they are in the same place no point talking about conf.d and then giving the full path, give the full path straight away.

MarcinOrlowski: Done

PR #467, opened 2019-09-06
Update autostart.rst

PawelWozniak:
Added Information to get service file from git. I as newbie was looking on local system for that file.


calmh:
Eh. The files are included in our release archives, like https://github.com/syncthing/syncthing/releases/download/v1.2.2/syncthing-linux-amd64-v1.2.2.tar.gz, no git required.

PR #468, opened 2019-09-14
Clarify usage of multiple listen addresses

TwizzyDizzy:
Before it was not clear how to specify multiple listen addresses in the config file. Before, you would assume to separate them by comma in the <listenAddress> block, which does not work. Comma separation is for the GUI, the config file needs multiple to be specified like this:

```
<listenAddress>ADDRESS1</listenAddress>
<listenAddress>ADDRESS2</listenAddress>
```

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -792,8 +792,7 @@ setLowPriority
 Listen Addresses
 ^^^^^^^^^^^^^^^^
 
-The following address types are accepted in sync protocol listen addresses
-(multiple can be specified by comma separation):
+The following address types are accepted in sync protocol listen addresses. If you want Syncthing to listen on multiple addresses, you can have multiple ``<listenAddress></listenAddress>`` sections. The same is achieved in the GUI by entering several addresses separated by comma.
 
 Default listen addresses (``default``)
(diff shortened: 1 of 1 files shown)


users/config.rst
```diff
@@ -792,8 +792,7 @@ setLowPriority
 Listen Addresses
 ^^^^^^^^^^^^^^^^
 
-The following address types are accepted in sync protocol listen addresses
-(multiple can be specified by comma separation):
+The following address types are accepted in sync protocol listen addresses. If you want Syncthing to listen on multiple addresses, you can have multiple ``<listenAddress></listenAddress>`` sections. The same is achieved in the GUI by entering several addresses separated by comma.
```
calmh:
```suggestion
The following address types are accepted in sync protocol listen addresses. If you want Syncthing to listen on multiple addresses, you can have multiple ``<listenAddress>`` tags. The same is achieved in the GUI by entering several addresses separated by comma.
```

TwizzyDizzy: Yep, even better and more concise :)

calmh:
Thanks!

PR #470, opened 2019-10-02
Update introducer.rst

nekr0z:
Warn users about adverse effects of an "introducer loop".


users/introducer.rst
```diff
@@ -18,3 +18,11 @@ If an introducer adds or removes any devices or shares, or changes device-share
 Note that devices which are introduced cannot be removed so long as the introducer device is still marked as such, and if they are unshared from a folder they will be re-added.
 
 Introducer status is transferable; that is, an introducers' introducer will become your introducer as well.
+
+It is not a good idea to set two devices as introducers to each other. While this will work for adding devices, removing a device may present a problem, as the two devices will be constantly "re-introducing" the removed device to each other:
```
calmh: This is good. I think the example is over complicating things maybe, might be less confusing to have just the initial paragraph?

nekr0z: Given the overall concise style of the documentation now that I think about it you're right.

PR #474, opened 2019-10-23
users/config: Key type is ECDSA nowadays

GermanCoding:
And has been for ages now :)


calmh:
Maybe just remove that specification altogether and just say "the device's public and private key"?

GermanCoding:
I like when something's specific. For me it highlights that 'key' is an actual cryptographic standard, but I understand that not everyone may see it that way. Anyway, if you don't like it I can remove it.

PR #476, opened 2019-11-05
Update doc to reflect removal of voxgratia.org STUN server

jowouters:
Update documentation to reflect removal of stun.voxgratia.org  ( See #6011 )


calmh:
Thanks!

PR #480, opened 2019-12-15
Draft docs for untrusted devices

calmh:


users/untrusted.rst
```diff
@@ -0,0 +1,204 @@
...
+        "U1" [label="U1\n(Encrypted)", style=filled, color="/accent3/2"]
+
+        T1 -> U1 [label="Encrypted by T1"]
+        U1 -> T1 [label="Decrypted by T1"]
+    }
+
+From this setup it's also possible to add further trusted devices, say
+``T2``, and have these sync the data from the untrusted device ``U1``
+without being in contact with ``T1``. Using the *same folder password* on
+``T2`` makes the existing data on ``U1`` intelligeble and the plaintext data
```
bugith, line 33: intelligeble would be intelligible

imsodin:
I added a warning to the two pages that this feature is still in testing. I propose we merge it now. Apparently search engines are already directing users at the PR docs, causing confusion (see https://github.com/syncthing/syncthing-android/issues/1615)

PR #486, opened 2020-01-27
Change Discovery Port to 21025

z2z:


calmh:
What makes you think 21025 is the current discovery port?

PR #487, opened 2020-01-27
Added windows -home path note

z2z:
Added ... "For Custom path in Windows ``X:\ABC\DEF``, directory ``ABC`` has to exist."


calmh:
I think the rule is the same for all platforms, namely that either the directory must exist or Syncthing must have permission to create it.

PR #497, opened 2020-03-18
Upgrade REST API example for configuration to current format version 30.

acolomb:
The docs leave out many properties added since configuration format version 15.  Even worse, the `pendingFolder` sub-structure has been moved underneath the device objects.  So the current documentation is misleading in that regard.

The changes for the new example are kept minimal, based on what a currently running instance of v1.4.0 reports, with some hand-crafted edits.

### DIFF
--- a/rest/system-config-get.rst
+++ b/rest/system-config-get.rst
@@ -3,109 +3,107 @@ GET /rest/system/config
 
 Returns the current configuration.
 
-.. code-block:: bash
+.. code-block:: json
 
     {
-        {
-          "version": 15,
-          "folders": [
-            {
-              "id": "GXWxf-3zgnU",
-              "label": "MyFolder",
-              "path": "...",
-              "type": "sendreceive",
-              "devices": [
-                {
-                  "deviceID": "..."
-                }
-              ],
-              "rescanIntervalS": 60,
-              "ignorePerms": false,
-              "autoNormalize": true,
-              "minDiskFreePct": 1,
-              "versioning": {
-                "type": "simple",
-                "params": {
-                  "keep": "5"
-                }
-              },
-              "copiers": 0,
-              "pullers": 0,
-              "hashers": 0,
-              "order": "random",
-              "ignoreDelete": false,
-              "scanProgressIntervalS": 0,
-              "pullerSleepS": 0,
-              "pullerPauseS": 0,
-              "maxConflicts": 10,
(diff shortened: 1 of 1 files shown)


rest/system-config-get.rst
```diff
@@ -3,109 +3,159 @@ GET /rest/system/config
...
+	"useTLS": false,
+	"apiKey": "pGahcht56664QU5eoFQW6szbEG6Ec2Cr",
+	"insecureAdminAccess": false,
+	"theme": "default",
+	"debugging": false,
+	"insecureSkipHostcheck": false,
+	"insecureAllowFrameLoading": false
+      },
+      "ldap": {
+	"addresd": "",
```
calmh: The `addresd` is a travesty and we should fix it now :(

acolomb: In the code or in the docs? That was taken literally from a running config.

acolomb:
You were (too) quick :-)

Not sure why the syntax highlighting does not work for this page. It does for `system-connections-get.rst`, which uses a wild mixture of indentation levels, though.

rest/system-config-get.rst
```diff
@@ -3,109 +3,159 @@ GET /rest/system/config
...
+	  "allowedNetworks": [],
+	  "autoAcceptFolders": false,
+	  "maxSendKbps": 0,
+	  "maxRecvKbps": 0,
+	  "ignoredFolders": [],
+	  "pendingFolders": [
+	    "time": "2019-06-05T10:21:22+02:00",
+	    "id": "cpkn4-57ysy",
+	    "label": "SomeonesFolder"
+	  ],
```
calmh: This is broken, might be why the syntax highlight doesn't kick in

rest/system-config-get.rst
calmh: In the code, I'll update this after.

rest/system-config-get.rst
acolomb: Quite right, that was a manual edit. Will fix ASAP.

acolomb:
I pushed another commit fixing the syntax error: acolomb@22f050e1f3402fc1508a410a23aee247666fe81c

acolomb:
Thanks for the push access. Fix applied and it looks beautiful again.

calmh:
Thank you!

PR #498, opened 2020-03-18
rest: API for pending devices and folders.

acolomb:
Basically mirror the information that currently lives within the GET
/rest/system/config endpoint, using two new endpoints under
/rest/system.

Note that for pendingFolders, the structure is inverted: In the
config, they live underneath the "devices" tree, here the offering
devices are listed underneath the pendingFolders entry with the same
information.

Include some example content (totally made up for now ~~as there is no
implementation yet~~).


acolomb:
As discussed in https://github.com/syncthing/syncthing/pull/5758, this is a first step and RFC to agree on the API.

imsodin:
Looks sensible to me, except that wrapping the entire thing into a dict with a single key named after the endpoint feels redundant, I'd just return the contents.

acolomb:
That was inspired by looking at some other endpoint responses. It makes it easier to add other / meta information at a later time, for example a "totals" section like in `/rest/system/connections`.

However, I'm open to changing it in the name of the KISS principle, sacrificing hypothetical extensibility. If that is the preferred consensus...

AudriusButkevicius:
Connections is a bad example. Dics should not be used for anything else than a single object.

We probably also want endpoints to mutate these.

acolomb:
@AudriusButkevicius Good idea actually. Noted on my to-do list, but I'll focus on the database stuff first. In a first step, adding them to the config like the GUI currently does will still work and not require any changes that would be moot after moving to a separate endpoint.

Do you have a concrete idea what these mutating requests should look like?

acolomb:
For the `pendingDevices` endpoint, it was suggested to live as just another property within the `GET /rest/system/connections` endpoint response. Would anyone favor that approach?

In my opinion the `/rest/system/*` structure is proliferating. If we're already introducing new stuff here, should we consider a different second-level path for this, e.g. `/rest/cluster/*`? `discovery` and `connections` would fit there as well, long-term.

acolomb:
See also #499 for the next evolutionary step concerning suggested candidate devices.

acolomb:
> Connections is a bad example. Dics should not be used for anything else than a single object.

The `GET /rest/system/error` endpoint also follows that pattern by the way.

AudriusButkevicius:
Yes, and some old crufty file exploration endpoints.

acolomb:
As suggested by @calmh [here](https://github.com/syncthing/syncthing/pull/6443#discussion_r431773076), the entries are now returned as object attributes with the device / folder ID as key.

The offering devices for a pending folder are still contained in a lone `offeredBy` object attribute.  Should that be skipped as well?  I think the structure is easier to understand this way.

Kindly requesting feedback before implementing the change in the actual code PR.

st-review:
:robot: *beep boop*

dev/rest.rst
```diff
@@ -32,9 +32,21 @@ Config Endpoints
 
 .. toctree::
    :maxdepth: 1
+   :glob:
```
imsodin, line 35: There's no globbing here, is there?

rest/cluster-pending-devices-get.rst
```diff
@@ -0,0 +1,17 @@
+GET /rest/cluster/pending/devices
+=================================
+
+.. versionadded:: 1.FIXME
```
imsodin, line 4: Probably 1.13.0 :)

dev/rest.rst
```diff
@@ -32,9 +32,21 @@ Config Endpoints
...
 
    /rest/config/... <../rest/config.rst>
 
+Cluster Endpoints
+-----------------
+
+Concerns the mesh network structure.
```
imsodin, line 42: Does "mesh" add anything to "network structure"? Seems redundant to me. And all the other subtopics don't have an explanation, maybe remove this entirely making this entire section just a plain, compact listing of endpoints.

rest/system-config-get.rst
```diff
@@ -3,7 +3,8 @@ GET /rest/system/config
 
 .. deprecated:: v1.12.0
    This endpoint still works as before but is deprecated. Use :ref:`rest-config`
-   instead.
+   instead.  Also note that the ``pendingFolders`` and ``pendingDevices``
+   fields have been removed as of v1.FIXME.
```
imsodin, line 7: I wouldn't start documenting individual changes to the config. We didn't do that until now and I don't see us (consequently) doing it in the future. Having just a few mentioned might confuse and/or create wrong assumptions that it is documented. And having them in the config is not an error.

I just wanted to say that there should however be a deprecation note in users/config.rst, however the pending elements were never documented there.

imsodin (commented):
Thanks for also taking care of the docs!

rest/system-config-get.rst
acolomb: So what do we do about the examples?  Just remove the corresponding lines?

Is there an easy way to browse older versions of the deployed documentation? Like we know it from readthedocs for example.

dev/rest.rst
acolomb: Just copied from the other subsections.  No idea if it's needed at all.

acolomb: `/home/acolomb/src/syncthing-docs/dev/rest.rst:44: WARNING: toctree contains reference to nonexisting document 'rest/cluster-*'`

No, it's needed.

dev/rest.rst
acolomb: I think an explanation for each group would be magnificent, but we can remove this one if it stands out.  Do you think the other groups could be explained in a similar manner?

rest/cluster-pending-devices-get.rst
acolomb: I wasn't sure if it would be v1.12.2.  I assume there is something already baking to justify v1.13.

rest/system-config-get.rst
imsodin: I'd just remove them.

Good question. I remember discussing that, probably when talking about alternative ways to organize/edit documentation. I don't think there's any way to select past version currently.

dev/rest.rst
imsodin: This is the "Config Endpoints" section though, not cluster. There is globbing in cluster (and everywhere else), but not in config.

dev/rest.rst
imsodin: I think all the other endpoints are more or less self-explanatory. At least adding something like "Query information from database" wouldn't add much to "Database Endpoints". "Cluster Endpoints" is a bit special in that respect, as I don't think it's immediately obvious what is meant by "Cluster". So lets keep the explanation.

acolomb: In that case, I find the "mesh network" appropriate.  It's a separate class of networks and an established technical term AFAIK.

dev/rest.rst
acolomb: Sorry, just a merge gone bad then. Fixed.

rest/system-config-get.rst
acolomb: There are the previews for every checked commit in a PR, but they vanish after some time.  Maybe the same mechanism could be used to deploy to a permanent location for each tag?

acolomb: > I'd just remove them.

Done.

rest/cluster-pending-devices-get.rst
imsodin: This. It's a config and api change.

acolomb: Well explained. :-)

acolomb:
Is this okay to merge, now that the implementation landed in `main`?  Probably good for wrapper developers to start early.

calmh (approved):

PR #501, opened 2020-03-29
users/autostart: On systemd info explicitly reference our packaging

luni3359:
Fixing a mistake and reminding the user that the repo also includes these service files. I accidentally followed step #2 because I didn't notice the service files were already installed.


luni3359:
Here, it was just minor:
``you can skip step 2 when you setting up → you can skip step #2 when you're setting up``

PR #505, opened 2020-04-07
users/ignoring: Document top-level include pattern optimization

sa3dany:
This documents the changes addded in syncthing/syncthing#6508

### DIFF
--- a/users/ignoring.rst
+++ b/users/ignoring.rst
@@ -85,6 +85,17 @@ The ``.stignore`` file contains a list of file or path patterns. The
    Prefixes can be specified in any order (e.g. "(?d)(?i)"), but cannot be in a
    single pair of parentheses (not ":strike:`(?di)`").
 
+.. note::
+
+   Include patterns (that begin with ``!``) cause Syncthing to traverse and
+   :ref:`watch <scanning>` the entire directory tree regardless of other
+   ignore patterns.
+
+   Top-level Include patterns (beginning with ``/`` and matching at the root of
+   the folder only) are treated as special cases and will not force Syncthing to
+   scan the entire directory tree. For example: ``!/foo`` is a top-level include
+   pattern, while ``!/foo/bar`` is not.
+
 Examples
 --------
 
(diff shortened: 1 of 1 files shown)


users/ignoring.rst
```diff
@@ -140,3 +134,23 @@ all files and directories called "foo", ending in a "2" or starting with
...
+    *
+
+This will only sync ``bar``, ``baz.txt`` and ``foo`` while ignoring everything
+else in the folder.
+
+.. note::
+
+   Syncthing optimizes the performance of this use case by only scanning and
+   watching the included files and directories. This applies **only** to top-level
+   include patterns. So having a pattern like ``!foobar`` or ``!**foo**`` without a
```
imsodin, line 155: If giving concrete examples, I think something like ``!/foo/bar`` should be present.

sa3dany:
Thanks for the review. Since the optimization was specifically added to improve the scanning performance of the `include-only` use-case, I thought it would be fitting to include an example.

Your are right about the traversal note being out of place.

Maybe I should scrap the example from this pull request to keep it simple and also because no example is better than a bad one.

What about editing the note about the affect of negative patterns on scanning to:

```
Include patterns (that begin with ``!``) cause Syncthing to traverse and
:ref:`watch <scanning>` the entire directory tree regardless of other
ignore patterns.

Starting with v1.x.x, top-level include patterns no longer force Syncthing to
scan the entire directory tree.
```

I feel that `top-level` warrants some explaining through some example patterns. But I don't want to add too much inside of a `Note`. So if you feel the above is not clear then I will have dedicate a section on the page or an example as I initially tried to do.

imsodin:
> Maybe I should scrap the example from this pull request to keep it simple and also because no example is better than a bad one.

Agreed.

I am not sure whether the note would really be too long with just a few examples, maybe that will work out. If not, please explain the performance/traversal implications on it's own, i.e. not as part of an example like now.

users/ignoring.rst
sa3dany: Yup, after looking at the tests in your pull requests I understood the behavior better. 👍

users/ignoring.rst
```diff
@@ -91,8 +91,13 @@ The ``.stignore`` file contains a list of file or path patterns. The
...
    ignore patterns.
 
-Example
--------
+   Top-level Include patterns (beginning with ``/`` and matching at the root of
+   the folder only) are treated as special cases and will not force Syncthing to
+   scan the entire directory tree. For example: ``!/foo`` is a top-level include
+   pattern, while ``!/foo/bar`` is not.
+
+Examples
```
imsodin, line 99: Nit: It's still a single example now.

users/ignoring.rst
```diff
@@ -91,8 +91,13 @@ The ``.stignore`` file contains a list of file or path patterns. The
    :ref:`watch <scanning>` the entire directory tree regardless of other
    ignore patterns.
 
-Example
--------
+   Top-level Include patterns (beginning with ``/`` and matching at the root of
```
imsodin, line 94:
```suggestion
   Top-level include patterns (beginning with ``/`` and matching at the root of
```

users/ignoring.rst
```diff
@@ -91,8 +91,13 @@ The ``.stignore`` file contains a list of file or path patterns. The
...
+   the folder only) are treated as special cases and will not force Syncthing to
```
imsodin, line 95: I'd drop the the text in parentheses, the examples in the end should make it clear already.

PR #512, opened 2020-05-07
users/config: Update example config to match Syncthing v1.5.0

tomasz1986:
Update the Config File Format example to match the default configuration
file used in Syncthing v1.5.0. Also, update the Configuration, Folder,
Device, and Options Elements accordingly. The GUI and LDAP Elements do
not need to be updated.

Add:
- autoAcceptFolders
- copyOwnershipFromParent
- crashReportingEnabled
- crashReportingURL
- databaseTuning
- ldap
- maxConcurrentIncomingRequestKiB
- maxFolderConcurrency
- maxRecvKbps
- maxRequestKiB
- maxSendKbps
- modTimeWindowS

Change order:
- defaultFolderPath
- filesystemType
- setLowPriority

Change value:
- fsWatcherEnabled (to "true")
- id (to "default")
- listenAddress (to "tcp://0.0.0.0:8384" and "dynamic+https://relays.syncthing.net/endpoint")
- rescanIntervalS (to "3600")
- urAccepted (to "-1")
- urSeen (to "3")
- urUniqueID (to "")
- version (to "30")

Remove:
- minHomeDiskFreePct

Replace:
- pullers (with pullerMaxPendingKiB)
- stunKeepaliveSeconds (with stunKeepaliveStartS and stunKeepaliveMinS)

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -129,6 +129,7 @@ The following shows an example of the default configuration file (IDs will diffe
             <stunKeepaliveStartS>180</stunKeepaliveStartS>
             <stunKeepaliveMinS>20</stunKeepaliveMinS>
             <stunServer>default</stunServer>
+            <databaseTuning>auto</databaseTuning>
         </options>
     </configuration>
 
--- a/users/config.rst
+++ b/users/config.rst
@@ -130,6 +130,7 @@ The following shows an example of the default configuration file (IDs will diffe
             <stunKeepaliveMinS>20</stunKeepaliveMinS>
             <stunServer>default</stunServer>
             <databaseTuning>auto</databaseTuning>
+            <maxConcurrentIncomingRequestKiB>0</maxConcurrentIncomingRequestKiB>
         </options>
     </configuration>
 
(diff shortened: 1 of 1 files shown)


tomasz1986:
No, it is my pleasure to be able to help at least a little bit :slightly_smiling_face:.

I have just one question. I see that commits in these pull requests are merged into one when accepted, and the commit message is just a list of all the single commits.

Would it be better if I squash and pre-merge them locally, and then create a pull request of the merged commit with a more meaningful message?

Right now, this does not look very pretty: https://github.com/syncthing/docs/commit/401060e978b77073485752749c7c0c38ea935a62.

calmh:
So what we usually do is have one commit with a good commit message. Then, changes might happen as a result of review and these get piled on top as "wip", "review" or whatever commits without useful subjects. At the end we squash and clean up to the initial message. In your case the individual subjects seemed roughly useful so I didn't have the heart to just delete all of them, nor the energy to clean it up to be beautiful.

tomasz1986:
Thank you for a quick answer. I was not sure, and locally while working I prefer to keep the commits very specific, so I just left everything as it was. Next time, I will try to do 

> one commit with a good commit message

instead. Just in case, please let me know any time there is something wrong in my future pull requests.

PR #514, opened 2020-05-07
users/config: Remove Windows XP info and fix LOCALAPPDATA case

tomasz1986:
Not only does Syncthing not support Windows XP, but also the current
versions do not even run on Windows older than 7. Thus, remove the lines
mentioning Windows XP, as leaving them could indicate official support,
which is not the case.

Also, the location of %LOCALAPPDATA%/Syncthing listed as "Windows 7+" is
technically incorrect, since Windows Vista also uses the same place to
store Syncthing files. However, the current versions of Syncthing do not
run under Windows Vista either, hence there is no reason to mention it
specifically.

In addition, change the environment variable LOCALAPPDATA to all
uppercase to match the notation used by Microsoft in the official
Windows documentation.

See: https://docs.microsoft.com/windows/deployment/usmt/usmt-recognized-environment-variables

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -10,8 +10,7 @@ Synopsis
 
     $HOME/.config/syncthing
     $HOME/Library/Application Support/Syncthing
-    %AppData%/Syncthing
-    %localappdata%/Syncthing
+    %LOCALAPPDATA%/Syncthing
 
 Description
 -----------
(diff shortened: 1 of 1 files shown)


tomasz1986:
I hate to do this, because I like(d) Windows NT 5.x (2000/XP/2003) way more than the current iterations, but I do not think that there is any point in mentioning Windows XP when Syncthing does not even run in the OS any more.

calmh:
Yeah, time to let go of XP. Paragraph is also slightly outdated in that since 1.5.0 there are the -config and -data options to counter the "... uses a single directory for ..." part of the docs...

PR #515, opened 2020-05-08
users/autostart: Rework Task Scheduler configuration

tomasz1986:
users/autostart: Rework Task Scheduler configuration

Task Scheduler is capable of running Syncthing in multiple ways. It is
a good idea to show a way to run it "at log on", and also "on startup".
This way, the user can choose whether they want to start and use
Syncthing while being logged on, or whether they want to have Syncthing
running in background all the time, started right after Windows boots,
and even when the computer is being used by other people.

Furthermore, the "run at log on" option is a good alternative to using
the Startup folder in the Start Menu, especially since Microsoft have
made access to the folder difficult in Windows 10. The folder is
currently not displayed in the Start Menu any more, and can be accessed
only by navigating manually to its actual location inside %APPDATA%,
which itself is a hidden folder, invisible to regular Windows users.

In addition, due to the complexity of Task Scheduler, add screenshots
to clearly show each configuration option on the screen, so that even
beginners should be able to follow the instructions easily.

The actual modifications are as follows:

- Add a short explanation on Task Scheduler itself with a link to the
  official documentation by Microsoft.
- Add simple instructions on how to start Task Scheduler.
- Add a new option describing how to run Syncthing "at system startup"
  in addition to the current "run at log on".
- Add additional explanations to each of the two methods.
- Move optional settings to a new section called "Additional
  configuration in Task Scheduler".
- Add new additional settings to optionally run Syncthing on battery
  power, and using a specific network connection.
- Add screenshots to each step of the instructions.
- Fix the language to match the terminology used by Microsoft (e.g. log
   in -> log on).
- Add cosmetic fixes (indentation, code marking, grammar).


tomasz1986:
I have made quite a lot of changes, so please check the preview, and let me know if there are any issues, or if something should be made differently.

users/autostart.rst
```diff
@@ -20,32 +20,141 @@ there are a number of easy solutions.
...
+background.
+
+You can launch Task Scheduler by going to ``Start Menu > Windows
+Administrative Tools`` and clicking on ``Task Scheduler``
+
+|Windows Task Scheduler Start Menu Screenshot|
+
+or by opening ``taskschd.msc`` from the Run command (``Win+R``).
+
+|Windows Task Scheduler Run Screenshot|
```
imsodin, line 35: These two screenshots seem a bit superfluous, the explanation should be fairly clear. And I'd consider opening the startmenu (or hitting win key) plus typing "task sched..." the easiest and most fool proof option (unless of course that doesn't work, but it should, right?).

users/autostart.rst
```diff
@@ -20,32 +20,141 @@ there are a number of easy solutions.
...
+
+#. Create a new Task ("Actions" sidebar > "Create Task...").
+
+   |Windows Task Scheduler Screenshot|
+
+#. General tab:
+
+   #. Name the task (for example "Syncthing").
+
+   |Windows Task Scheduler General Logon Screenshot|
```
imsodin, line 54: The name field seems pretty unmistakable and in your face to me.

users/autostart.rst
```diff
@@ -20,32 +20,141 @@ there are a number of easy solutions.
...
+      connection changes or becomes unavailable later.
+
+   |Windows Task Scheduler Conditions Screenshot|
+
+#. Actions Tab:
+
+   #. Select the previously created action and click "Edit...".
+   #. Enter the path to the parent folder of ``syncthing.exe`` in "Start in
+      (optional)". This will allow you to use paths relative to this folder
+      in Syncthing.
```
imsodin, line 154: What does this do? Respectively where in Syncthing do you want to use relative parts?

users/autostart.rst
tomasz1986: Yes, that is true. I just wanted to add screenshots to each and every step for the sake of thoroughness. Should I remove this one?

users/autostart.rst
tomasz1986: You can use relative paths for Syncthing folders, e.g. `./Sync` will create a folder `C:\Syncthing\Sync`. Without the "Start in" field defined, the folder will be created in `%SYSTEMROOT%\System32`, so it will be `C:\Windows\System32\Sync`.

Personally, I would say that defining the "Start in" field is a good practice.

users/autostart.rst
tomasz1986: The problem with typing is that it is locale-dependent, so "task sched..." will work only in the English version of Windows. In the other language versions, you need to use the name of Task Scheduler in that language.

The Start Menu path is also language dependent, but at least it is located in the same place, so it is easy to find, even if your Windows is not in English. On the other hand, ``taskschd.msc`` works regardless of the Windows language.

When it comes to the screenshots, as with the Task name thing, I added them for completeness. They are probably not necessary, although from my experience, some people simply need visual cues to be able to navigate around. Written instructions just do not cut it :smirk:.

tomasz1986:
> Nice :)
> 
> Is this accessible without admin. Respectively when started at system startup, does it run as the user (not some kind of admin/elevated privileges process)? If yes, are there any drawbacks over NSSM? Otherwise I'd say lets drop that alltogether - this looks both more user friendly and save.

Hmm, I personally have not used NSSM, so I would need to check it first, but with my instructions, Syncthing starts under the user account who has created the task (which is "DESKTOP-BF8ORGA\User" in the screenshots).

The difference between "run at log on" and "run at startup" is that the former starts the Task when the user logs on, and ends it when the user logs off, while the latter starts the task immediately after boot, and keeps it running regardless of whether the user logs on or not. The Task is still run under the same regular user account though.

As far as I understand, NSSM would be equal to using the "SYSTEM" account in Task Scheduler, which is not something that I would recommend to do.

users/autostart.rst
tomasz1986: Just to add, I myself use relative paths in Syncthing, at least on my home computer, where I have a separate `S:` partition specifically to store Syncthing and all my folders there. If something happens, I can then move everything together to a different location, and still be able to run Syncthing only by changing the "Start in" folder, without touching the actual configuration.

users/autostart.rst
imsodin: In my opinion yes. I don't see the value in the completeness alone, but I see value in having to scroll less. And here I think the added clarity is almost zero, at least below the annoyance of scrolling. Obviously that's personal bias as someone who gets on well with written instructions, so feel free to override me if you feel strongly about this.

users/autostart.rst
imsodin: Ah right, the locale - just recently faced that problem myself.

Personally I don't see how these help (the first one doesn't show where in the menu this item is and the second one is a popup with a single input field, not much to mistake there), but I am not the target audience so again: Feel free to override me if you feel it helps.

I don't disagree on the "some people" account, but I believe there's a limit. And that limit is where the hand-holding gets more time-intensive and the learning effect is too low, than just doing it for them (in this context: them asking "the family geek" or writing to the forum if they're advanced). And e.g. for the task name the field is so in your face that I'd say that limit is reached. While the added page length (more scrolling) affects the 99.999% that do not fail to find the very first input field on the dialogue.

users/autostart.rst
imsodin: > It is possible to use relative paths for Syncthing folders, e.g. `./Sync` will create a folder `C:\Syncthing\Sync`. Without the "Start in" field defined, the folder will be created in `%SYSTEMROOT%\System32`, i.e. `C:\Windows\System32\Sync` (which is ... bad).

That's in my opinion a reason to not allow relative paths at all. I think the inconvenience for you to have to type the disk letter of your separate drive does not compare to the weirdness of having a folder in `C:\Windows\System32`. However this hasn't come up as a problem yet, so I am definitely not going to actively destroy the usage scenario of our most active docs contributor ;)

imsodin:
> As far as I understand, NSSM would be similar to using the "SYSTEM" account in Task Scheduler, which is not something that I would recommend to do. Do you think that the NSSM option should be removed completely?

In my opinion yes. Recommending running with elevated privileges is not something we should do in my opinion (as there is an alternative for system startup, which is important). Someone who needs that for obscure reason should be able to determine how to do it on their own. But then again I am like the opposite of an authority on windows, so maybe we shouldn't do this on my account.
  
The NSSM docs were originally committed by @calmh and once modified by @ProactiveServices, and @Catfriend1 might have some windowsy insight too: What do you think about dropping NSSM instructions (running at system start with elevated privileges) as one can run syncthing un-privileged at system start with task scheduler? To me looks like there's only drawbacks to that method.

users/autostart.rst
tomasz1986: Yeah, the Start Menu screenshot is problematic. In order to show the location clearly, the whole Start Menu would have to be included, which is very tall, even in my virtual installation of Windows 10 Enterprise 2019 LTSC (i.e. only built-in tools and no Metro applications). That is why I cut it like that :sweat:.

As for the `Win+R` method, the screenshot is indeed unnecessary once you get the Run command window open, as there is nothing to mess up with there.

I would still like to include some kind of a helpful visual clue for those, who will be opening Task Scheduler for the fist time though... I will see what I can do here without adding superfluous screenshots.

users/autostart.rst
tomasz1986: I actually agree, but at the same time, this would kill any portable usage of Syncthing, i.e. being able to keep both the installation, and the synced folders on a USB external drive and use it on different machines, so it is a double-edged sword.

tomasz1986:
I hope that I will not hurt the author's feelings, but if I were to express my opinion, then I would actually propose to additionally remove the Startup folder instructions. They do not offer anything more than the Task Scheduler already does (with way more functionality), and what is more important, Microsoft has hidden the Startup folder from the Start Menu in Windows 10. The folder now requires going to %APPDATA% manually, which is a hidden system folder. This basically makes it inaccessible for regular users, who do not know how to show hidden files and folders.

users/autostart.rst
tomasz1986: Actually, I would propose to keep these screenshots. The reason is that there is a  very important difference between

`|Windows Task Scheduler General Startup Screenshot|`

image

and

`|Windows Task Scheduler General Logon Screenshot|`

image

Of course, I am talking about the bottom part. I think that it is a good idea to have both of them, so that the difference between the two methods is clearly visible.

On the other hand, it would maybe be better to have small thumbnails instead of full screenshots, and then anyone could just click and view each of them in full when necessary. I do not think that we have this functionality in Sphinx though.

Catfriend1:
Hi,

In my opinion you're on the right way here to recommend the Task Scheduler as an easy self-doable way to normal users. It has some disadvantages but nothing critical. E.g. if there are two many tasks simulatenously running windows will hit the max task limit and defer running other tasks. That's not coming up as a blocker until you'd "misuse" the scheduler for constantly running apps more than X times. The user can increase the max task limit.

Personally, I'm in favor to use nssm as a service wrapper and like its reliable clean start and stop mechanisms and that it may hold the shutdown sequence for some seconds until syncthing.exe terminated completely. I'm very cautious here because I'm fearing database corruption in theory. The system account has a lot of power, less security when running syncthing there and I like to configure ST remotely in our corporate environment. That's why I will soon present the next step is my syncthing-msi approach - a nssm thingy. I hope this maybe can be listed on the community contributions page as a link later, so to get back to topic: I wouldn't recommend to send normal users to configure the System account + nssm + syncthing on their own because I did this since 2016 and learned a lot of mistakes and problems that can come up by this. A unique msi package for sys admins and pro users would be best from my point of view plus removing the nssm part from the docs.

calmh:
I know nothing about NSSM and certainly didn’t write the instructions for it. That said, a Windows service should be able to run under any valid user account, so it shouldn’t need to be limited to running as SYSTEM

tomasz1986:
@Catfriend1 Just for the record, the task number limitations are explained in https://support.microsoft.com/help/2696472. I have personally never encountered this issue, and I highly doubt that any home user would, but I can imagine this being a problem on servers or similar. As for the shutdown mechanism, Task Scheduler seems to simply do an immediate `taskkill /f`, which is definitely not the cleanest method (although for sure is effective).

I am a little busy with work right now, but I will try to do something about the screenshots shortly and at least try to make them a little bit less "in your face" (and possibly remove the superfluous ones).

ProactiveServices:
The reason I was using NSSM was that I wanted Syncthing instances to run irrelevant of what user accounts were being used, I wanted them to run with different user accounts in groups that had very restricted privileges, and have their own locked-down file system access. My secondary requirements were clean exiting, managed restarts via the Service Manager, failure alerting via integration with the Event Log, delayed start and a few others. For this use case, using a service-based approach strikes me as more appropriate than using Task Scheduler.

I'm not against using the Task Scheduler per se, I'm just not sure of a use case which precludes a user-oriented wrapper like SyncTrayzor, which serves excellently, with simple options and no "gotchas". Tasksch has always felt, to me, to be a method of running one-shot programs for condition-based reasons rather than a step towards running an on-going service.

However, we all know that there is variety in the weird and wonderful things people do with Syncthing and this could help people who have use cases (or skill sets) which differ from ours. These new docs look comprehensive and thoughtful and I think there is definitely overlap between the two uses ("proper" service vs. more advanced user-level operation); I'm not sure one replaces the other.

ProactiveServices:
> This basically makes it inaccessible for regular users, who do not know how to show hidden files and folders. I believe that it is much easier and fool-proof to just launch and configure Task Scheduler instead.

For users who do not use a wrapper I think the Start Menu method is still a  simple way to run the binary, if an ugly one, even with Microsoft hiding it. onfiguring a tasksched entry is certainly more complicated and fraught - it's a process I'm comfortable with, and proficient at, but it's easy to make mistakes with, so I'm not sure it's more user-friendly. I went down the Start Menu route to add a second instance of SyncTrayzor.

tomasz1986:
Thank you very much for expressing your opinion on this!

>Tasksch has always felt, to me, to be a method of running one-shot programs for condition-based reasons rather than a step towards running an on-going service.

In most cases that is true, but Microsoft themselves have a few always-running tasks, and I have seen 3rd party software using Task Scheduler to keep their programs running in background all the time, so I would say that the situation is not exactly black and white here :wink:.

>For users who do not use a wrapper I think the Start Menu method is still a simple way to run the binary, if an ugly one, even with Microsoft hiding it.

If we were talking about Windows 7 and older, then yes, I would agree 100%, but with Windows 10, the current instructions have become outdated and basically impossible to follow. The Startup folder does not exist (at least visibly in the Start Menu), so there is nothing to right-click and open. The only way to get to it is to physically navigate to the folder, whose parent is hidden, i.e. also "invisible" for non-power users.

That said, I had actually began to update the Startup folder instructions for Windows 10 locally, but then @imsodin mentioned removing NSSM, and with that the whole much bigger discussion began :open_mouth:.

I am still rather a new user of Syncthing, and as such I do not want to push my opinion in any way. My question is though, which methods to run Syncthing in Windows should remain in the Docs? Should we have as many methods as possible, or rather stick to a few selected ones?

Right now, we have:

- Task Scheduler
- Startup folder
- Third party (SyncTrayzor)
- NSSM (service)

In addition to these, there is also a way to run from Registry, which is similar to the Startup folder method, but this one is not mentioned in the Docs at the moment.

My favourite is Task Scheduler, but as with everything, personal preferences differ. However, if the Startup folder is to stay, then I will update it for Windows 10. I just need to know whether it should be kept or removed first :sweat_smile:.

users/autostart.rst
imsodin: Fine, all screenshots it is :)

There is https://pythonhosted.org/sphinxcontrib-images/ and that might be nice, but I don't expect you to do it - I have no clue how difficult it is and how well it would work.

imsodin:
The consensus I see is: We like having options, and NSSM is one of them. As long as there's demand and instructions are not woefully outdated, I wouldn't remove options. I was just wondering whether NSSM has been made redundant by the task scheduler option to run at system startup, which it apparently has not :)

users/autostart.rst
tomasz1986: Yeah, I have also found some possible addons to Sphinx online, but I guess that they would require a server-side installationand configuration, right?

By the way, I have an idea in my head to merge the two Task Scheduler methods, so that most of the screenshots could be used only once, and for both. This way, we could reduce the number of duplicate screenshots, and also make the whole instructions shorter. I am going to try it in the next few days, and then update the commit.

tomasz1986:
No problem with that. I will work on updating the other instructions for Windows 10 then, but I need to finish this Task Scheduler thing first.

tomasz1986:
I am very sorry for everything taking so long. The commit should be ready to merge now. Please check the changes and let me know whether any fixes are necessary.

I have merged the "run at log on" and "run at startup" methods into one explanation, with two choices under the Triggers tab. The rest of the configuration is basically the same for both, so there is no reason to have them separated. This has also allowed me to greatly reduce the number of screenshots. I have also polished the language, and added a tip about not using the "run at startup" option when Syncthing folders are stored on a BitLocker protected partition.

users/autostart.rst
```diff
@@ -20,32 +20,111 @@ there are a number of easy solutions.
...
-    #. Make sure Enabled is checked.
-    #. Click "OK".
-#. Actions Tab:
-    #. Click "New...".
-    #. [Action] should be set as "Start a program".
-    #. Enter the path to syncthing.exe in "Program/Script".
-    #. (optional) Enter the path to the parent folder of syncthing.exe in "Start in (optional)". This will allow you to use paths relative to this folder in Syncthing.
-    #. Click "OK".
+Task Scheduler is a built-in administrative tool, which can be used to
+start Syncthing automatically either at user log on, or at system
```
imsodin, line 24: Just a tiny nit: To my understanding there's the verb "log in" when you enter credentials to get access to something, and the noun "login" describing that process. So to my understanding it should be "login" here instead of "log on" (and same for all other occurrences). Disclaimer: I am not a language authority, so might be wrong - refute me if appropriate ;)

tomasz1986: I actually explained it at the very end of the commit message :wink:.

>Fix the language to match the terminology used by Microsoft (e.g. log
in -> log on).

The key here is that Microsoft uses "log on" (with a space), both as a verb, and a noun. Specifically in Task Scheduler, it clearly says `at log on`, so I think that it is better to stick to the Microsoft's terminology to avoid any confusion, at least in the Windows-related instructions.

I think that these are related:

https://stackoverflow.com/questions/406016/ui-terminology-logon-vs-login
https://stackoverflow.com/a/406522 (regarding Microsoft)

PR #516, opened 2020-05-08
users/autostart: Link to Windows solutions at the top of the section

tomasz1986:
Add easy-to-navigate and descriptive internal links to the sub-sections
of the Windows section. This way, the users can easily see what kind of
options are available, and then go straight to the one, which interests
them the most.

### DIFF
--- a/users/autostart.rst
+++ b/users/autostart.rst
@@ -19,14 +19,15 @@ there are a number of easy solutions.
 
 Built-in solutions:
 
-- `Run at user log on or at system startup using Task Scheduler
-  <#task-scheduler>`__
-- `Run at user log on using Startup folder <#start-on-login>`__
+- :ref:`Run at user log on or at system startup using Task Scheduler <autostart-windows-taskschd>`
+- :ref:`Run at user log on using the Startup folder <autostart-windows-startup>`
 
 Other solutions:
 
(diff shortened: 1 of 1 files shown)


users/autostart.rst
```diff
@@ -17,6 +17,17 @@ Windows
 There is currently no official installer available for Windows. However,
 there are a number of easy solutions.
 
+Built-in solutions:
+
+- `Run at user log on or at system startup using Task Scheduler
+  <#task-scheduler>`__
+- `Run at user log on using Startup folder <#start-on-login>`__
```
calmh, line 24: I think these should be rst-style references, not based on the HTML anchors. As is, this will break in pdf and man page exports etc. Otherwise LGTM

tomasz1986: Let me fix them. I am not very accustomed to the whole RST formatting yet. I would like to finish https://github.com/syncthing/docs/pull/515 first though, as this one is somewhat co-dependent. I will try to do both this week.

tomasz1986: I think that I have fixed the URLs. Please check and feel free to merge the commit if everything is OK :slightly_smiling_face:.

tomasz1986:
I can see that there are conflicts after the recent commits. I will try to fix them and then update the PR shortly.

PR #517, opened 2020-05-09
users/stdiscosrv: Add Apache Reverse Proxy config to the discosrv documentation

Pitxyoki:
I had a hard time figuring out how to configure Apache httpd, syncthing-discosrv, as well as the Syncthing clients to use a discovery server behind an httpd reverse proxy.

Although the documentation is good enough for serving the GUI behind a proxy, and probably for serving the discovery server behind an nginx proxy, it was not that easy for having it behind and httpd proxy.

Additionally, it was not clear to me at all that the `id=...` parameter should be omitted after this configuration.

This change tries to make these things explicit in the documentation so that others do not face the same problems I had.

### DIFF
--- a/users/stdiscosrv.rst
+++ b/users/stdiscosrv.rst
@@ -219,8 +219,8 @@ Requirements
 - The "X-SSL-Cert" HTTP header must be passed through with the PEM-encoded
   client SSL certificate. This will be present in POST requests and may be empty
   in GET requests from clients. If you see syncthing-discosrv outputting
-  :code:`no certificates` when POSTs are issued on it, that's because the
-  proxy is not passing this header through.
+  :code:`no certificates` when receiving POST requests, that's because the proxy
+  is not passing this header through.
 - The proxy must request the client SSL certificate but not require it to be
   signed by a trusted CA.
 
@@ -311,7 +311,7 @@ configuration:
 For more details, see also the recommendations in the
 `Reverse Proxy Setup <https://docs.syncthing.net/users/reverseproxy.html>`__
 page. Note that that page is directed at setting up a proxy for the
-Syncthing WebUI. You should do the proper path and port adjustments to proxying
+Syncthing web UI. You should do the proper path and port adjustments to proxying
 the discovery server and your particular setup.
 
 
--- a/users/stdiscosrv.rst
+++ b/users/stdiscosrv.rst
@@ -205,20 +205,46 @@ allows:
 - Use of a subdomain name without requiring a port number added to the URL
 - Sharing an SSL certificate with multiple services on the same server
 
+Note that after this configuration, clients have to omit the :code`?id=...`
+parameter of the discovery server URL on their configuration. Client-side
+validation will be done by checking the visible proxy server's certificate.
+
 Requirements
 ^^^^^^^^^^^^
 
 - Run the discovery server using the -http flag  :code:`stdiscosrv -http`.
 - SSL certificate/key configured for the reverse proxy
-- The "X-Forwarded-For" http header must be passed through with the client's
+- The "X-Forwarded-For" HTTP header must be passed through with the client's
   real IP address
-- The "X-SSL-Cert" must be passed through with the PEM-encoded client SSL
-  certificate
+- The "X-SSL-Cert" HTTP header must be passed through with the PEM-encoded
+  client SSL certificate
 - The proxy must request the client SSL certificate but not require it to be
   signed by a trusted CA.
 
+Apache
+""""""
+Add the following lines to the configuration:
+
+.. code-block:: apache
+
+    SSLProxyEngine On
+    SSLVerifyClient optional_no_ca
+    # The following is required, otherwise syncthing-discosrv
+    # outputs 'no certificates' when a POST is issued on it
+    RequestHeader set X-SSL-Cert "%{SSL_CLIENT_CERT}s"
+    # Apparently the RemoteIPHeader directive is not really required.
+    # I couldn't understand why it is recommended for a simple reverse proxy
+    # with no other services in front.
+    # Using this would require using a2enmod remoteip
+    #RemoteIPHeader X-Forwarded-For
+
+See also the recommendations in the
+`Reverse Proxy Setup <https://docs.syncthing.net/users/reverseproxy.html>`__
+page.
+
+
 Nginx
-^^^^^
+"""""
 
 These three lines in the configuration take care of the last three requirements
 listed above:
--- a/users/stdiscosrv.rst
+++ b/users/stdiscosrv.rst
@@ -205,43 +205,24 @@ allows:
 - Use of a subdomain name without requiring a port number added to the URL
 - Sharing an SSL certificate with multiple services on the same server
 
-Note that after this configuration, clients have to omit the :code`?id=...`
-parameter of the discovery server URL on their configuration. Client-side
-validation will be done by checking the visible proxy server's certificate.
+Note that after this configuration, **clients have to omit the** :code:`?id=...`
+**parameter from the discovery server URL on their configuration**. Client-side
+validation will be done by checking the visible proxy server's HTTPS certificate.
 
 Requirements
 ^^^^^^^^^^^^
 
-- Run the discovery server using the -http flag  :code:`stdiscosrv -http`.
+- Run the discovery server using the -http flag: :code:`stdiscosrv -http`.
 - SSL certificate/key configured for the reverse proxy
 - The "X-Forwarded-For" HTTP header must be passed through with the client's
   real IP address
 - The "X-SSL-Cert" HTTP header must be passed through with the PEM-encoded
-  client SSL certificate
+  client SSL certificate. This will be present in POST requests and may be empty
+  in GET requests from clients. If you see syncthing-discosrv outputting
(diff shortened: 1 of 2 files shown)


users/stdiscosrv.rst
```diff
@@ -205,20 +205,46 @@ allows:
...
+    SSLProxyEngine On
+    SSLVerifyClient optional_no_ca
+    # The following is required, otherwise syncthing-discosrv
+    # outputs 'no certificates' when a POST is issued on it
+    RequestHeader set X-SSL-Cert "%{SSL_CLIENT_CERT}s"
+    # Apparently the RemoteIPHeader directive is not really required.
+    # I couldn't understand why it is recommended for a simple reverse proxy
+    # with no other services in front.
+    # Using this would require using a2enmod remoteip
+    #RemoteIPHeader X-Forwarded-For
```
AudriusButkevicius, line 239: This is required, because that's how discovery server gets remote peers address, which it then uses when telling others where to find the device.

Pitxyoki: So, I was studying this a bit more by sniffing what Apache is sending back do stdiscosrv.

On the POST requests, I see the following headers:
```
    127.0.0.1.56506 > 127.0.0.1.8443: (...)
(...)
........POST / HTTP/1.1                                                                                                                                                   
Host: localhost:8443                           
User-Agent: Go-http-client/1.1                                                                
Content-Type: application/json                                                                                                                                               
Accept-Encoding: gzip                                                                                                                                                               
X-SSL-Cert: -----BEGIN CERTIFICATE----- (...) -----END CERTIFICATE-----                                                                                              
X-Forwarded-For: <visible IP of the client>                                                                                                                                         
X-Forwarded-Host: <public address of my Apache server>                                                           
X-Forwarded-Server: <public address of my Apache server>                                                                                                                                           
Connection: Keep-Alive                                                                         
Content-Length: 118                                                                                                                                                            
                                                                                              
{"addresses":["relay://<public address of my Apache server>:22067/?id=("]}                                                              
```

On GET requests, I see the following:
```
    127.0.0.1.56414 > 127.0.0.1.8443: (...)

.qF..qF.GET /?device=(...) HTTP/1.1
Host: localhost:8443                                                  
User-Agent: Go-http-client/1.1   
Accept-Encoding: gzip                  
X-SSL-Cert: (null)                                
X-Forwarded-For: <visible IP of the client>
X-Forwarded-Host: <public address of my Apache server>                                             
X-Forwarded-Server: <public address of my Apache server>                                                   
Connection: Keep-Alive                                                               
```

So, this leads me to conclude that apparently httpd adds the `X-Forwarded-For` field by default, without any extra configuration.
Should we still recommend a more complex configuration (running a2enmod + adding that line) given that, at least on httpd 2.4.38, this is not needed by default?

I'm more inclined to add this explicit information there, so that users may want to try with no a2enmod+RemoteIPHeader, and if they fail, they can try adding it.

AudriusButkevicius: I have no idea, you're the apache specialist here.

AudriusButkevicius:
Also I think apache should probably go under nginx, as we run nginx and are more prepared to support it.

The first thing people will see is apache, install it, and when when it doesn't work, we will have no idea how to fix it, as we ourselves don't run apache.

users/stdiscosrv.rst
Pitxyoki: Heh, fair enough. I updated the PR with this information, hopefully making it simple enough that you won't be annoyed by too many support requests regarding it.

users/stdiscosrv.rst
```diff
@@ -205,20 +205,27 @@ allows:
...
 - SSL certificate/key configured for the reverse proxy
-- The "X-Forwarded-For" http header must be passed through with the client's
+- The "X-Forwarded-For" HTTP header must be passed through with the client's
   real IP address
-- The "X-SSL-Cert" must be passed through with the PEM-encoded client SSL
-  certificate
+- The "X-SSL-Cert" HTTP header must be passed through with the PEM-encoded
+  client SSL certificate. This will be present in POST requests and may be empty
+  in GET requests from clients. If you see syncthing-discosrv outputting
+  :code:`no certificates` when POSTs are issued on it, that's because the
```
AudriusButkevicius, line 222: When receiving POST requests

users/stdiscosrv.rst
```diff
@@ -280,6 +287,35 @@ Server and Syncthing using Nginx, `Let's Encrypt`_ and Docker can be found here_
...
+configuration:
+
+.. code-block:: apache
+
+    RemoteIPHeader X-Forwarded-For
+
+For more details, see also the recommendations in the
+`Reverse Proxy Setup <https://docs.syncthing.net/users/reverseproxy.html>`__
+page. Note that that page is directed at setting up a proxy for the
+Syncthing WebUI. You should do the proper path and port adjustments to proxying
```
AudriusButkevicius, line 314: web UI

users/stdiscosrv.rst
```diff
@@ -205,20 +205,27 @@ allows:
 - Use of a subdomain name without requiring a port number added to the URL
 - Sharing an SSL certificate with multiple services on the same server
 
+Note that after this configuration, **clients have to omit the** :code:`?id=...`
+**parameter from the discovery server URL on their configuration**. Client-side
+validation will be done by checking the visible proxy server's HTTPS certificate.
```
AudriusButkevicius, line 210: I don't think this is true, the ID is used for verifying discovery server is who it says it is.
It has to be a valid TLS cert, or if it's self signed, the hash of the cert is turned into an ID.

Pitxyoki: That's odd.
When I use this configuration, I get the following when I define `?id=<discovery server ID>`:

```[THISDEV] 2020/05/12 22:10:05.990327 global.go:155: DEBUG: globalClient.Lookup https://<myserver>.org/syncthingdiscosrv?device=SOMEDEV security: incorrect device id
[THISDEV] 2020/05/12 22:10:05.990827 cache.go:124: DEBUG: lookup results for SOMEDEV
[THISDEV] 2020/05/12 22:10:05.991038 cache.go:125: DEBUG:   addresses:  []
[THISDEV] 2020/05/12 22:10:06.019513 global.go:238: DEBUG: announce POST: security: incorrect device id
```

When I do not set such ID:
```[THISDEV] 2020/05/12 21:12:50.765689 cache.go:102: DEBUG: lookup for SOMEDEV at global@https://<myserver>.org/syncthingdiscosrv
[THISDEV] 2020/05/12 21:12:50.765739 cache.go:103: DEBUG:   addresses: [relay://<myserver>:22067/?id=RELAYID]
[THISDEV] 2020/05/12 21:12:50.765770 cache.go:124: DEBUG: lookup results for SOMEDEV
[THISDEV] 2020/05/12 21:12:50.765788 cache.go:125: DEBUG:   addresses:  [relay://<myserver>.org:22067/?id=RELAYID]
[THISDEV] 2020/05/12 21:12:50.773228 global.go:243: DEBUG: announce POST: 204 No Content
[THISDEV] 2020/05/12 21:12:50.773275 global.go:270: DEBUG: announce Reannounce-After: 3392 <nil>
```
Note especially the POST parts.

Using tcpdump, I see the discovery server replying with HTTP 204 for both cases.
So, this is the client not being able to validate what comes from the server when the id is set. The server is indeed considering all requests valid.

By reading the page in https://docs.syncthing.net/specs/globaldisco-v3.html#authentication , I see the following:
> The “id” parameter is not, in fact, sent to the discovery server - it’s used by Syncthing itself to know which certificate to expect on the server side.

So, it seems that the protocol is something like:
1. Client sends POST to the discovery server
2. Discovery server replies with a 204.
3. Because that response comes through HTTPS, the client will try to validate the certificate associated with that response against the locally-configured ID.

This means that, when the connection is coming from a proxy (that by definition creates a new HTTPS connection), the proxy's certificate will never be possible to validate against the client's locally-configured ID for the discovery server.
This will be true for any proxy server, be it Apache, Nginx or any other.

The only validation done on the client side, when no ID is configured, appears to be that the certificate associated to the proxy's HTTPS certificate is valid. I checked this briefly by setting an invalid certificate on the server. The syncthing client got this:
```
[THISDEV] 2020/05/12 22:48:25.412384 global.go:238: DEBUG: announce POST: Post https://<myserver>.org/syncthingdiscosrv: x509: certificate has expired or is not yet valid
```
Additionally, and as you say:
> Discovery server has to have a valid TLS cert, or if it's self signed, the hash of the cert is turned into an ID, for client to verify there is no man in the middle.

With a proxy setup, we want to make sure there is a man in the middle. The assurance the assurance can get is if the client validates the configured "man-in-the-middle"'s (our proxy server) HTTPS certificate is valid regarding its DNS name. And the client validating this certificate is what gives us the same confidence level we get when the ID is used in the no-proxy setup.

Is this correct?
If you can confirm this assessment, I can try to make the doc simpler to explain this.

AudriusButkevicius: It is correct, but removing the ID is not universally the answer.
It's only the answer if the proxy serves a valid certificate, if it's not a valid certificate, then the user should somehow compute the id of the cert and pin the cert that way.

Pitxyoki: I rephrased this paragraph. What do you think?

AudriusButkevicius:
Thanks

PR #520, opened 2020-06-08
users/autostart: Rework Startup folder configuration

tomasz1986:
Update the Startup folder configuration for newer versions of Windows.

Firstly, remove the information about downloading Syncing as unrelated
to the actual instructions. We assume that syncthing.exe is already
present in c:\syncthing. Secondly, use Windows Explorer to create a
shortcut directly, without relying on a .bat file to run Syncthing.
Thirdly, add screenshots to show the most important steps visually.
Lastly, remove the note about starting Syncthing with /low priority,
as it does not work with the current default configuration anyway (see
https://github.com/syncthing/syncthing/issues/6597).


users/autostart.rst
```diff
@@ -137,40 +137,38 @@ browser to interact with Syncthing).
...
+   from the Run command (``Win+R``).
+
+#. Right-click empty space in the folder and choose "New", and then
+   "Shortcut".
+
+   |Windows Startup Folder New Shortcut Screenshot|
+
+#. Enter the path to ``syncthing.exe`` in "Type the location of the item:"
+   followed by ``-no-console -no-browser`` (for example ``C:\syncthing\syncthing.exe
+   -no-console -no-browser``)
```
imsodin, line 161:
```suggestion
   -no-console -no-browser``).
```

users/autostart.rst
```diff
@@ -137,40 +137,38 @@ browser to interact with Syncthing).
...
+
+   |Windows Startup Folder Create Shortcut Screenshot|
+
+#. Click "Next".
+#. Click "Finish".
```
imsodin, line 166: This feels redundant with the info in the screenshot (and common sense), but then again I guess it doesn't hurt either

users/autostart.rst
tomasz1986: I clicked on the "Commit suggestion" button, and now we are both listed as the authors. Should I commit the change manually using git instead?

users/autostart.rst
tomasz1986: There is actually a following screen with the "Finish" button, where you can also change the shortcut's name, but it is already set to "syncthing" by default (from `syncthing.exe`), so I skipped the additional explanation.

PR #525, opened 2020-06-18
users/config: Update folder docs

AudriusButkevicius:


AudriusButkevicius:
No idea how to preview it, don't have an RST builder.

users/config.rst
```diff
@@ -193,6 +193,10 @@ Folder Element
         <markerName>.stfolder</markerName>
         <copyOwnershipFromParent>false</copyOwnershipFromParent>
         <modTimeWindowS>0</modTimeWindowS>
+        <maxConcurrentWrites>2</maxConcurrentWrites>
+        <disableFsync>false</disableFsync>
+        <blockPullOrder>standard</blockPullOrder>
+        <copyRangeMethod>standard</copyRangeMethod>
```
tomasz1986, line 199: The same 4 lines should probably be added to the folder element under https://github.com/syncthing/docs/blob/new-docs/users/config.rst#config-file-format (i.e. between the lines 71 and 72).

imsodin:
Preview: https://build.syncthing.net/repository/download/Web_BuildDocumentation/69814:id/html.tar.gz!/index.html

users/config.rst
```diff
@@ -341,6 +349,99 @@ markerName
     Name of a directory or file in the folder root to be used as
     :ref:`marker-faq`. Default is ".stfolder".
 
+copyOwnershipFromParent
+    On Unix systems, tries to copy file/folder ownership from the parent directory (the directory it's located in).
+    Requires running Syncthing as privileged user, or granting it additional capabilities (e.g. CAP_CHOWN on Linux).
+
+modTimeWindowS
+    Allowed modification timestamp difference when comparing local and remote file for equivalence.
```
imsodin, line 357:
```suggestion
    Allowed modification timestamp difference when comparing files for equivalence.
```

Also applies in the scanner (local to local).

users/config.rst
```diff
@@ -341,6 +349,99 @@ markerName
...
+    To be used on systems that have unstable modification timestamps, that might change after being observed after
+    the last write operation. Used in Android only.
+
+maxConcurrentWrites
+    Maximum number of concurrent write operations while syncing. Defaults to 2. Increasing this might increase or
+    decrease disk performance, depending on the underlying storage.
+
+disableFsync
+    Disables committing file operations to disk before recording them in the database.
+    Can lead to data corruption if disabled.
```
imsodin, line 367: Potentially confusing, given the `disableFsync` setting being enabled is dangerous (which disables fsync). 

```suggestion
    Disabling fsync can lead to data corruption.
```

AudriusButkevicius:
looks good to me.

users/config.rst
```diff
@@ -341,6 +349,99 @@ markerName
...
+    Allowed modification timestamp difference when comparing local and remote file for equivalence.
+    To be used on systems that have unstable modification timestamps, that might change after being observed after
+    the last write operation. Used in Android only.
+
+maxConcurrentWrites
+    Maximum number of concurrent write operations while syncing. Defaults to 2. Increasing this might increase or
+    decrease disk performance, depending on the underlying storage.
+
+disableFsync
+    Disables committing file operations to disk before recording them in the database.
```
imsodin, line 366: I'd discourage usage even more, something like this:

```suggestion
    Do not enable this option unless you know exactly what you are doing.
    Disables committing file operations to disk before recording them in the database.
```

And or use the "warning-thingy" (just discovered below).

users/config.rst
```diff
@@ -341,6 +349,99 @@ markerName
...
+    Can lead to data corruption if disabled.
+
+blockPullOrder
+    Order in which parts of a file are downloaded. This option controls how quickly the different parts of the
+    file spread between the connected devices, at the cost of causing strain on the storage.
+
+    Available options:
+
+    standard (default):
+        The file is split into N equal continuous part sequences, where N is the number of connected
```
imsodin, line 376:
```suggestion
        The file is split into N equal continuous sequences, where N is the number of connected
```

users/config.rst
```diff
@@ -341,6 +349,99 @@ markerName
...
+        devices, each device starts downloading it's own sequence, after which it picks other devices
```
imsodin, line 377:
```suggestion
        devices. Each device starts downloading it's own sequence, after which it picks other devices
```

users/config.rst
```diff
@@ -341,6 +349,99 @@ markerName
...
+    random:
+        File parts are downloaded in a random order. Provides great data distribution, but very taxing on
+        spinning disk drives
+
+    inOrder:
+        File parts are downloaded sequentially, from start to finish. Spinning disk drive friendly, but provides
+        no improvements to data distribution.
+
+copyRangeMethod
+    Provides an choice of method for copying data between files. This can be used to optimise copies on network
```
imsodin, line 389:
```suggestion
    Provides a choice of method for copying data between files. This can be used to optimise copies on network
```

users/config.rst
```diff
@@ -341,6 +349,99 @@ markerName
...
+        Available on: Windows
+
+    all
+        Tries all of the copy methods in the following order: ioctl, copy_file_range, sendfile, duplicate_extents,
+        standard.
+
+        Available on: All platforms
+
+        .. warning::
+            Not recommended on Windows as it has not been tested, might be very costly on all other platforms.
```
imsodin, line 443: This could be read as implying it's been tested, regardless of the initial warning. I'd reword to

```suggestion
            Not recommended as it has not been tested. Also might be very costly on all platforms but Windows.
```

AudriusButkevicius: This was tested on non-windows to some extent.

users/config.rst
AudriusButkevicius: I think it's imporant to leave part here, because file is split into parts, so it's important to know its sequences of parts.

AudriusButkevicius:
So I made some comments back, the rest of the changes suggestions look sensible. I'd love to unmerge, reopen, commit suggestions and remerge but I don't think I can do that.

If you feel strongly about these, feel free to follow up.

users/config.rst
imsodin: Ah, now I get what distinction you want to stress here. I'd use "blocks" then e.g.

```suggestion
        The blocks of the file are split into N equal continuous sequences, where N is the number of connected
```

imsodin:
Nah I don't think the changes will change the world, but given I have taken the time to review and you to review the review I'll do a follow-up ;)

PR #526, opened 2020-06-19
users/config: Review of #525

imsodin:
Followup to https://github.com/syncthing/docs/pull/525#pullrequestreview-434214306 - ping @AudriusButkevicius

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -371,7 +371,7 @@ disableFsync
     Disabling fsync can lead to data corruption.
 
 blockPullOrder
-    Order in which parts of a file are downloaded. This option controls how quickly the different parts of the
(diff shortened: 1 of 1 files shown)


users/config.rst
```diff
@@ -373,20 +377,20 @@ blockPullOrder
     Available options:
 
     standard (default):
-        The file is split into N equal continuous part sequences, where N is the number of connected
-        devices, each device starts downloading it's own sequence, after which it picks other devices
-        sequence at random. Provides acceptable data distribution, minimal spinning disk strain.
+        The blocks of a file are split into N equal continuous sequences, where N is the number of connected
```
AudriusButkevicius, line 380: There is no term blocks mentioned before, so I think wedging a new term half way is confusing.

PR #534, opened 2020-07-15
users/stdiscosrv: New HTTP header "X-Client-Port" (ref syncthing#5753)

GermanCoding:
This relates to https://github.com/syncthing/syncthing/pull/5753. From my understanding, discovery reverse proxies should now send this header in order to make fancy TCP headbanging possible.

Corrected the tutorial/example nginx snippets. As for Apache, I have no idea if and how it's possible to add such a custom header.

### DIFF
--- a/users/stdiscosrv.rst
+++ b/users/stdiscosrv.rst
@@ -199,6 +199,10 @@ discovery servers.
 Reverse Proxy Setup
 ~~~~~~~~~~~~~~~~~~~
 
+.. versionadded:: 1.8.0
+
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
The header was changed to X-Client-Port because X-Forwarded-Port means something else (the server port the connection was received on)

GermanCoding:
That's good to hear, I will update things accordingly when I have a minute.

GermanCoding:
Header name is now fixed.

AudriusButkevicius:
Thanks for caring about this :)

PR #540, opened 2020-08-01
users/crashrep: Add note about using custom builds of Syncthing

tomasz1986:
Add a note explaining the inclusion of information such as the user's
computer, domain, and user name in the crash logs. Also, provide a
solution on how not to include those when building Syncthing locally.

### DIFF
--- a/users/crashrep.rst
+++ b/users/crashrep.rst
@@ -39,12 +39,11 @@ The following is an example of a crash report as sent::
     main.syncthingMain(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x49651b4, 0x1, 0x0, 0x0, ...)
             /buildagent/work/github.com/syncthing/syncthing/cmd/syncthing/main.go:873 +0x1bc5
 
-Note that when using a custom compiled build, the names of the domain,
-computer, and user of the machine where Syncthing was built will be
-included in the log as part of the version string. This information is
-essential for the developers to interpret the crash log in context. If
-you compile Synching locally and want to prevent your build from having
-such data embedded, set the ``COMPUTERNAME``, ``USERDOMAIN``, and
-``USERNAME`` variables to ``unknown`` before the compilation.
+
+Note that the username and hostname of the machine where Syncthing was
+built will be included in the crash log as part of the version string.
+This information is essential for the developers to interpret the log in
+context. If you compile Syncthing locally and want to prevent your build
+from having such data embedded, see :ref:`versiontagging`.
 
 For a more details description of the format and how the sending happens, see :ref:`crashrep-dev`.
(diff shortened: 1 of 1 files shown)


calmh:
This sounds to me like if you're using any custom build and it crashes we will report your username etc, which is not the case. I'd say something about the fact that the binary includes information about who built it.

AudriusButkevicius:
I think this should clarify that those things get embedded in the version, and the version gets reported as part of the crash report

tomasz1986:
I think that I have included all the requested changes in the note. Is it fine now?

users/crashrep.rst
```diff
@@ -39,4 +39,12 @@ The following is an example of a crash report as sent::
     main.syncthingMain(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x49651b4, 0x1, 0x0, 0x0, ...)
             /buildagent/work/github.com/syncthing/syncthing/cmd/syncthing/main.go:873 +0x1bc5
 
+Note that when using a custom compiled build, the names of the domain,
```
calmh, line 42: s/when using a custom compiled build,// ?

users/crashrep.rst
```diff
@@ -39,4 +39,12 @@ The following is an example of a crash report as sent::
...
+computer, and user of the machine where Syncthing was built will be
+included in the log as part of the version string. This information is
+essential for the developers to interpret the crash log in context. If
+you compile Synching locally and want to prevent your build from having
+such data embedded, set the ``COMPUTERNAME``, ``USERDOMAIN``, and
```
calmh, line 47: I'm not sure where these variables come from but they don't appear to be what we actually look at


https://github.com/syncthing/syncthing/blob/a85bc1c3a64e33f1c42fe6e1dbeb19a953d01431/build.go#L999

calmh: There should be instructions on tweaking the compiled in metadata on the build page, and we should link to that from here rather than duplicating, I think

calmh: We can refer to https://docs.syncthing.net/dev/building.html#version-tagging by

```
:ref:`versiontagging`
```

tomasz1986: The variables come from https://forum.syncthing.net/t/15417/7. I must admit though that I have not tested them myself, but rather just believed the word :hushed:.

I will try to make all the changes shortly, but let me first actually compile and test everything myself just to make sure that there are no further similar flops on my part.

tomasz1986: I have simplified and rewritten the note to match the terminology used in https://docs.syncthing.net/dev/building.html#version-tagging. I have also tested the compilation, and indeed the variables listed by @calmh are the right ones, while the other ones do not do anything.

PR #551, opened 2020-08-03
refresh-authors: Escape underscores in thanks.txt

tomasz1986:
Underscores need be escaped in RST. If not, the words containing them
are later transformed into links in the HTML.

### DIFF
--- a/refresh-authors.sh
+++ b/refresh-authors.sh
@@ -8,7 +8,4 @@ grep \# AUTHORS > authors-hdr
 cat authors-hdr authors-new > AUTHORS
 rm authors-hdr authors-new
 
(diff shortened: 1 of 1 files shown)


tomasz1986:
I am not sure which operating system the script is being run though. Is `sed` actually available?

calmh:
Yeah it's not well defined so we might try for compatibility. (In practice, macOS or FreeBSD.) The invocation here doesn't work on Mac at least (non-GNU sed). I think simplifying to `... | sed 's/_/\\_/g' > thanks.txt` on the line above should work on all though?

tomasz1986:
I have merged the lines, but I cannot test it as I do not know how to compile/use the `git contributors` tool.

PR #553, opened 2020-08-03
intro/getting-started: Remove outdated scan interval tips

tomasz1986:
Remove the outdated information about scan interval and file watcher.
The current versions of Syncthing use a built-in file watcher, and in
normal circumstances need not tweak the scan interval also, especially
by beginning users.


tomasz1986:
I wanted to do this in a separate commit, but I hope that it is fine to remove this part. In my opinion, there is no need to tweak the default settings, unless one has a valid reason, and especially not by someone who has just installed the program.

Please let me know if you think otherwise, or think that the content should rather stay and/or be updated to match the current versions of Syncthing.

PR #562, opened 2020-08-09
users/firewall: Add info about local discovery concerning routers

RichardUUU:


users/firewall.rst
```diff
@@ -3,6 +3,12 @@
 Firewall Setup
 ==============
 
+Local Discovery
+-------------
+If your devices aren't mobile, you can replace the default ``dynamic`` setting for 
+Addresses with something like: ``tcp://192.168.1.xxx:22000, dynamic``
```
imsodin, line 9: By keeping the `dynamic`, device can still be mobile. A different requirement for this to work is that devices have/get the same local IP all the time.

tomasz1986:
I would also like to add that local ≠ static IP, i.e. local devices may very well use dynamic IPs determined by a DHCP server in the router. This is especially true when talking about home users (in contrary to offices), where people usually keep the default router/OS settings, which normally use dynamic IPs.

I think that it would be good to have at least a short and clear explanation stating why and when someone would potentially want to add static IPs to the configuration.

RichardUUU:
Great comments. Although the process for updating this PR is cryptic to me. Here is some possible language:

Local Discovery
-------------
If you are unable to set up your router/firewall as shown below, and your devices have static IP addresses, you can access them directly by changing the default ``dynamic`` setting for Addresses with something like: ``tcp://192.168.1.xxx:22000, dynamic``

imsodin:
I don't generally use the Github UI, thus tried it myself and did some changes. Please consider them as propositions, if you disagree just change/revert them.

To update the PR in Github itself:
1. Switch to "Files changed" tab and go to the file to be modified.
2. Click the three buttons on the top right corner and select "Edit file".
3. Do your modifications, add a commit message, leave the selection to "Commit directly to the *name* branch." and press "Commit changes" -> done.

imsodin:
Thank you for improving the docs!

PR #564, opened 2020-08-16
dev/issues: Update labels to match GitHub issues (fixes #559)

### LINKED ISSUE #559: dev/issues: Missing labels in Issue Management
https://docs.syncthing.net/dev/issues.html

I am trying to update the labels to match the actual labels used in the GitHub repository. I have added a short explanation to each one of them, but I am not sure whether they are adequate, so please let me know if different words should be used.

However, there seem to be a few labels that seem not to be used anymore, e.g. `pr-merged` last used in 2018. Should these labels remain in the Docs, or should they be removed? Also, the `unreleased` label has not been used at all, and is not listed in the Issues on GitHub either. Should it remain in the Docs?

This is my sketch for now:

```
api
    Issues related to the API and configuration.

bug
    Issues verified to be a bug.

build
    Issues caused by or requiring changes to the build system (scripts
    or Docker image).

dependencies
    Pull requests that update a dependency file.

documentation
    Issues related to the documentation.

enhancement
    New features or improvements of some kind, as opposed to a problem
    (bug).

frozen-due-to-age
    Set automatically on issues when they have been closed untouched
    for a long time, together with being locked for discussion.

good first issue
    Good starting points for new contributors, contained in scope and
    size, and clear about what is the desired outcome.

needs-triage
    New issues needed to be validated.

not-our-bug
    Rare and temporary label used only when we want to keep an issue
    open for visibility, but the real problem is somewhere else.

pr-merged
    Pull requests merged to the main tree.

protocol
    Issues that require a change to the protocol.

skip-changelog
    Issues minor enough not to be included in the changelog.

ui
    Issues related to the graphical UI.

unreleased
    Set on bugs that have never been in a released stable version, i.e.
    a bug introduced in v1.8.0-rc.1 and fixed in v1.8.0-rc.2, and as
    such excluded from the release notes for v1.8.0.

waiting-for-info
    Issues that require more information on how to reproduce.
```

tomasz1986:
Update the labels with their description to match the actual labels used
currently in the issues in the main Syncthing repository. Replace the
obsolete and unused "unreleased" with "skip-changelog", and mark the
"pr-merged" label as legacy kept for historical purposes. For all the
labels, use clear and easy to understand short descriptions.

### DIFF
--- a/dev/issues.rst
+++ b/dev/issues.rst
@@ -17,10 +17,13 @@ error", "Rendered moot by #123" and so on. We don't use the "invalid" or
 "wontfix" labels.
 
 api
-    Issues related to the API and configuration.
+    Issues or pull requests that change public APIs (config, REST). Used
+    to highlight these specifically in the release notes, and ensure we
+    don't miss to bump the minor version.
 
 bug
-    Issues verified to be a bug.
+    A problem with current functionality, as opposed to missing
+    functionality (enhancement).
 
 build
     Issues caused by or requiring changes to the build system (scripts
@@ -67,7 +70,8 @@ ui
     Issues related to the graphical user interface.
 
 waiting-for-info
-    Issues that require more information on how to reproduce.
+    Issues that require more information from the author on how to
+    reproduce.
 
 Milestones
 ----------
--- a/dev/issues.rst
+++ b/dev/issues.rst
@@ -16,38 +16,58 @@ Examples include "Duplicate of #123", "Discovered to be configuration
 error", "Rendered moot by #123" and so on. We don't use the "invalid" or
 "wontfix" labels.
 
+api
+    Issues related to the API and configuration.
+
 bug
-    The issue is a verified bug.
+    Issues verified to be a bug.
 
 build
-    The issue is caused by or requires changes to the build system
-    (scripts or Docker image).
+    Issues caused by or requiring changes to the build system (scripts
+    or Docker image).
+
+dependencies
+    Pull requests that update a dependency file.
+
+documentation
+    Issues related to the documentation.
 
 enhancement
-    This is a new feature or an improvement of some kind, as
-    opposed to a problem (bug).
+    New features or improvements of some kind, as opposed to a problem
+    (bug).
 
 frozen-due-to-age
-    This is set automatically on issues when they have been closed untouched
-    for a long time, together with the issue being locked against
-    discussion.
+    Set automatically on issues when they have been closed untouched
+    for a long time, together with being locked for discussion.
 
-good-first-issue
-    These issues are good starting points for new contributors. They are
-    contained in scope and size, and it should be clear what is the desired
-    outcome.
+good first issue
+    Good starting points for new contributors, contained in scope and
+    size, and clear about what is the desired outcome.
+
+needs-triage
+    New issues needed to be validated.
 
 not-our-bug
     Rare and temporary label used only when we want to keep an issue
     open for visibility, but the real problem is somewhere else.
 
-protocol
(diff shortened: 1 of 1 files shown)


dev/issues.rst
```diff
@@ -16,38 +16,58 @@ Examples include "Duplicate of #123", "Discovered to be configuration
...
+pr-merged
+    Legacy label used in the past for pull requests merged to the main
+    tree.
+
 protocol
-    This requires a change to the protocol.
+    Issues that require a change to the protocol.
+
+skip-changelog
+    Set on bugs that have never been in a released stable version, i.e.
```
AudriusButkevicius, line 62: it's set on PRs I think

dev/issues.rst
```diff
@@ -16,38 +16,58 @@ Examples include "Duplicate of #123", "Discovered to be configuration
...
+
+ui
+    Issues related to the graphical user interface.
 
-unreleased
-    This is set on bugs that have never been in a released stable version -
-    i.e., a bug introduced in v0.14.44-rc.1 and fixed in v0.14.44-rc.2. They
-    are excluded from the release notes for v0.14.44.
+waiting-for-info
+    Issues that require more information on how to reproduce.
```
AudriusButkevicius, line 70: more information from the issue creator

dev/issues.rst
tomasz1986: I have followed https://github.com/syncthing/docs/issues/559#issuecomment-671472241
>And skip-changelog replaced unreleased, i.e. unreleased can go. And skip-changelog is mainly used in the scenario you described on unreleased.

and https://github.com/syncthing/docs/issues/559#issuecomment-671795761:
>And skip-changelog should only be used for bugs introduced in an RC, which don't need to be mentioned in the change log because they weren't present in any previous version

What is the answer then? There have been only five PRs marked with this label, all created last year.

The description itself has literally been taken from the now removed `unreleased` label. I have only updated the Syncthing version numbers in the example.

AudriusButkevicius: You're right, it seems we set it on both, perhaps we should state bugs/PRs as it's clearly either.

calmh: It has defined meaning when set on bugs, as it prevents those bugs from then being mentioned in the changelog. It is meaningless to the automation when set on PRs, as PRs don't go into the changelog at all by themselves. I think the description here is fine, if someone wants to label the corresponding PR like this then it does no harm.

dev/issues.rst
```diff
@@ -16,38 +16,58 @@ Examples include "Duplicate of #123", "Discovered to be configuration
 error", "Rendered moot by #123" and so on. We don't use the "invalid" or
 "wontfix" labels.
 
+api
+    Issues related to the API and configuration.
```
calmh, line 20:
```suggestion
    Issues or pull requests that change public APIs (config, REST). Used to highlight these specifically in the release notes, and ensure we don't miss to bump the minor version.
```

dev/issues.rst
```diff
@@ -16,38 +16,58 @@ Examples include "Duplicate of #123", "Discovered to be configuration
...
+
 bug
-    The issue is a verified bug.
+    Issues verified to be a bug.
```
calmh, line 23:
```suggestion
    A problem with current functionality, as opposed to missing functionality (enhancement).
```

(The "verified" part doesn't apply any more since we use the needs-triage label for that)

dev/issues.rst
tomasz1986: Fixed. I have used "author" to match the GitHub terminology.

calmh (approved):

PR #568, opened 2020-09-21
users/syncing: Explain case-sensitivity principles (fixes #566)

### LINKED ISSUE #566: Explain case insensitive filename handling
There is very little information in the docs about what Syncthing does to avoid or handle case conflicts. The [advanced configuration section](https://docs.syncthing.net/advanced/folder-caseSensitiveFS.html) only says "Syncthing will attempt to detect and prevent case-only file name collisions that can occur on case insensitive systems", but not any scenario or algorithm what is handled and what leads to conflicts.

Thought this might serve as a reminder to document the behavior while it is still fresh in mind after the v1.9.0 release.

acolomb:
State the underlying assumptions and conditions which could lead to a
case conflict.  Advise how to correctly remedy such conflicts and give
some rationale why it's better than pre-v1.9.0 behavior.

Include a warning that the `caseSensitiveFS` advanced option does not
change the basic principles and should thus not be used.  Link to the
corresponding option description.


users/faq.rst
```diff
@@ -64,6 +64,41 @@ The following are *not* synchronized;
...
+
+This cautious behavior tries to save you from possible data loss
+caused by different files overwriting each other's contents. That
+could have happened before version 1.9.0, where the same file would
+erroneously be accessed under two case-differing file names.
+
+Note that the :ref:`caseSensitiveFS advanced option
+<case-sensitive-fs>` is **not** meant to change that basic principle,
+but only provides a performance optimization when the underlying
+filesystem is positively known to be case-sensitive already.
```
acolomb, line 97: If this mention is too prominent here, we could move the paragraph to the option description and add a link to this place instead for a more general overview of the topic.

imsodin:
What made you decide against putting this in https://docs.syncthing.net/users/syncing.html? Personally whenever there's a somewhat suitable place to put new content other than FAQ, I'd put it there. FAQ already is a huge pile of info.

users/faq.rst
imsodin: Please leave it out all-together. There's a tendency of "I have a problem, lets see what this switch does" among a lot of users, so I don't want anyone reading about case-sensitivity (probably because they have a problem) to know about this option. And except for its name it isn't actually related with the info put together here. The only time this option should come into action, is if a linux/unix user on the forum has weird performance issues and we want to be sure it's not because of the filesystem wrapper.

tomasz1986: I personally think that it would be beneficial to add some kind of a warning about this setting to https://docs.syncthing.net/advanced/folder-casesensitivefs. The current explanation does not say clearly enough that the option should never be used in case insensitive file systems.

acolomb:
As [mentioned here](https://github.com/syncthing/docs/issues/566#issuecomment-691162679) I thought it would be too much in proportion to the other paragraphs. But I'm okay with moving it.

users/faq.rst
acolomb: I also vote for using this to extend the advanced config description then. To specifically point the users to where "regular" case sensitivity documentation can be found and state that it has no meaningful use in that regard.

users/syncing.rst
```diff
@@ -79,6 +79,38 @@ synchronize the file. The block lists are compared to build a list of needed
...
+
+In principle Syncthing works with *case-sensitive* paths, meaning that
+`file.txt` and `FILE.txt` denote two independent things. It does not
+support synchronizing the contents of both as if they were somehow
+related.
+
+Some operating systems (e.g. Windows, Mac and Android) assume the
+opposite, treating them as the same file (or directory) and can never
+have both file names simultaneously. They will *preserve* the case
+from file creation, however.
```
imsodin, line 95: I wouldn't talk about what it "supports" or not, that sounds as if this was about feature completeness, when it is a fundamental design choice.

Also I am pretty sure Android is case-sensitive. External SDs used to be FAT, which is case-insensitive, but as far as I remember that hasn't been in use after Android 4.

Last point: What does preserve actually mean in this context? I understand what it means when syncing, but an OS/FS doesn't do that. Likely depending on what it means, is it relevant here? The main point is the sentence before, that on such systems no two files with names differing in case only can exist. I general I'd leave out any information that does not directly help understanding the point to make, otherwise it's just potential for distraction or even confusion.

```suggestion
Syncthing works with *case-sensitive* paths, meaning that
`file.txt` and `FILE.txt` denote two independent files. Some operating systems (e.g. Windows, Mac) or filesystems (e.g. FAT) assume the
opposite, treating them as the same file (or directory) and can never
have both file names simultaneously. They will *preserve* the case
from file creation, however.
```

users/syncing.rst
```diff
@@ -79,6 +79,38 @@ synchronize the file. The block lists are compared to build a list of needed
...
+
+Thus, if two remote devices share these differing case variants with
+your local device, they cannot both be synchronized to such a system
+and will be reported as a case conflict by Syncthing. One of them will
+likely be pulled, and the other one marked with an appropriate error
+message. Which one "wins" is currently not predictable. In order to
+resolve such a case conflict situation, you need to decide on a
+consistent file name and **manually** enforce that across all
+involved, case-insensitive devices.
```
imsodin, line 104: I'd use a simpler scenario here, involving just two case-insensitive devices which happen to have a case-conflict. From support so far that seems the most likely scenario. With three the point about "winning" is unrelated to the problem at hand, and also not really a problem: In most scenarios you'd add one device later than the other, so the synced file will just be the one from the device added first.

And then there's also https://github.com/syncthing/syncthing/issues/6989, which means that for identical files no manual intervention will be necessary any more. That's probably the most common scenario with clusters of case-insensitive devices only. So then the relevant scenario would be case-sensitive with two files differing in case connected to case-insensitive.

I would also mention that the scenario will result in failed items and put the error message verbatim here, such that a search for that message may lead to this content.

users/syncing.rst
tomasz1986: > Also I am pretty sure Android is case-sensitive. External SDs used to be FAT, which is case-insensitive, but as far as I remember that hasn't been in use after Android 4.

SD cards in Android are still (ex)FAT, and thus case insensitive, as far as I know. This is because they are designed to be ejected and used in different devices, which includes usually Windows, and Windows by default does not support any Linux file systems.

users/syncing.rst
acolomb: > I'd use a simpler scenario here, involving just two case-insensitive devices which happen to have a case-conflict. From support so far that seems the most likely scenario.

I wonder how people usually get into this situation. There must have been an unrelated synchronization / copying process before which screwed up file name case. Creating a file locally just before a remote device sends its own version, which it already had a name for before seeing our version? That's a typical conflict scenario anyway, because the contents would likely differ. Otherwise, what sequence of events would be needed to actually provoke a case conflict?

> With three the point about "winning" is unrelated to the problem at hand, and also not really a problem: In most scenarios you'd add one device later than the other, so the synced file will just be the one from the device added first.

They could both have been added long ago, but the local device was maybe just offline. When reconnecting to both conflicting (or not directly linked) remotes at the same time, the important point here is actually that the slightly quicker one will be reflected locally, while the error message from the other one might very well go unnoticed.

> And then there's also [syncthing/syncthing#6989](https://github.com/syncthing/syncthing/issues/6989), which means that for identical files no manual intervention will be necessary any more. That's probably the most common scenario with clusters of case-insensitive devices only. So then the relevant scenario would be case-sensitive with two files differing in case connected to case-insensitive.

Is any work being done on that? For me it's not clear from the issue what the automatic resolution would look like (which case variant will be chosen as "correct"). Basically what @calmh [said there](https://github.com/syncthing/syncthing/issues/6989#issuecomment-693992588), which also gives a rationale for my described three-device conflict above.

> I would also mention that the scenario will result in failed items and put the error message verbatim here, such that a search for that message may lead to this content.

I agree that the error message needs improvement, as it does not clearly point to the problem from a user's perspective. So let's follow up on this when a better message is implemented?

users/syncing.rst
acolomb: > I wouldn't talk about what it "supports" or not, that sounds as if this was about feature completeness, when it is a fundamental design choice.

Stating that it it doesn't support any other view of the world is trying to preempt questions about being able to change that to what many users think is normal coming from the Windows world. I still think it's important to clear up the two names never being considered related.

> Also I am pretty sure Android is case-sensitive. External SDs used to be FAT, which is case-insensitive, but as far as I remember that hasn't been in use after Android 4.

At least on my Android 7.1 device, the file manager will refuse to create a case-conflicting file. The "internal" storage is mounted as fstype `sdcardfs`, whatever that means. I'm not an expert on Android, so this information was taken from some maintainer's statement on an issue or forum discussion - can't find it quickly right now.

> Last point: What does preserve actually mean in this context? I understand what it means when syncing, but an OS/FS doesn't do that. Likely depending on what it means, is it relevant here? The main point is the sentence before, that on such systems no two files with names differing in case only can exist. I general I'd leave out any information that does not directly help understanding the point to make, otherwise it's just potential for distraction or even confusion.

Trying to bring in the technical terminology "case-preserving" for those who might have read about different "levels" of case-sensitivity after growing up with MS-DOS :-) You're right though, it's not really needed here.

users/syncing.rst
imsodin: > I wonder how people usually get into this situation. There must have been an unrelated synchronization / copying process before which screwed up file name case. Creating a file locally just before a remote device sends its own version, which it already had a name for before seeing our version? That's a typical conflict scenario anyway, because the contents would likely differ. Otherwise, what sequence of events would be needed to actually provoke a case conflict?

Someone mentioned using the same program on two devices, and that program apparently creates the same file sometimes in lower, sometimes upper case. In general I don't think much about "why" for case-insensitivity, that just creates headaches, I only think about how to prevent issues that exist (and plenty of weird stuff does exist).

> They could both have been added long ago, but the local device was maybe just offline. When reconnecting to both conflicting (or not directly linked) remotes at the same time, the important point here is actually that the slightly quicker one will be reflected locally, while the error message from the other one might very well go unnoticed.

There are scenarios where it's indeed situation where it's less clear which one wins, however my point is this has nothing to do with case conflicts. The same is true if those two devices modified the same file, producing a "real conflict". That's why I consider it unrelated information and thus I'd cut it.

> Is any work being done on that? For me it's not clear from the issue what the automatic resolution would look like (which case variant will be chosen as "correct"). Basically what @calmh [said there](https://github.com/syncthing/syncthing/issues/6989#issuecomment-693992588), which also gives a rationale for my described three-device conflict above.

I am planning to do it before the next release. I don't see how that comment is related to 3-way scenarios. As to how the automatic resolution would happen:

> We'll have to pick one casing essentially at random but the user apparently isn't picky anyhow or we wouldn't be here.

To make the choice deterministic (i.e. every device makes the same choice), I'd define that string `a` wins over `b`, if `a` > `b`.

acolomb: > Someone mentioned using the same program on two devices, and that program apparently creates the same file sometimes in lower, sometimes upper case. In general I don't think much about "why" for case-insensitivity, that just creates headaches, I only think about how to prevent issues that exist (and plenty of weird stuff does exist).

I'll try to master some of those headaches in search of a simpler example.

> There are scenarios where it's indeed situation where it's less clear which one wins, however my point is this has nothing to do with case conflicts. The same is true if those two devices modified the same file, producing a "real conflict". That's why I consider it unrelated information and thus I'd cut it.

Will case conflicts also result in the "losing" file being pulled with a `.case-conflict-...` suffix?  I assumed not, so the user will end up with only the winning version when they might have preferred the other.  That's the surprise I wanted to explain here.

> To make the choice deterministic (i.e. every device makes the same choice), I'd define that string `a` wins over `b`, if `a` > `b`.

This is definitely non-obvious, so I intend to incorporate it instead, when the implementation is ready. Can be a follow-up change though.

> > https://github.com/syncthing/syncthing/issues/6989#issuecomment-693992588 (emphasis added):
> > There should be some established / documented process for resolving a case conflict that doesn't fall under the above, i.e., when **we join two devices that have files with different contents and different name casing**. If one is older / less desired than the other, delete it. Otherwise rename one of them, etc.
> 
> I don't see how that comment is related to 3-way scenarios.

I thought he was talking about files with *same content*, and specifically about "joining" two *remote* devices.  Maybe just a misunderstanding.

acolomb:
Possible case-conflict scenarios with two devices A (local) and B (remote):

1. Devices are disconnected. Each adds a file locally, then they reconnect.  Both will have their own version and a conflict on the other.

2. Devices are connected.  They both add a file "simultaneously", with differing case.  Same as above, but very unlikely.

3. Playing with ignore / unignore, as described in previous issues. In a way that's "asking for trouble", but should yield the same end result as above.

4. Adding pre-populated folders with existing case conflicts to Syncthing.

I think the now included example explains the first two.  The others are probably not well suited for a documentation example.

If Syncthing does create `.case-conflict-...` files (as Dropbox does for example), that would eradicate the need to explain why an expected file is not even pulled from a remote. Principle of least surprise...

users/syncing.rst
imsodin: I am not interested in discussing these details any further, it requires too much time on to me almost completely irrelevant details - sorry for having engaged the discussion and now cutting it short on my side. I'll just answer the concrete question from the latest comment for completeness:

> Will case conflicts also result in the "losing" file being pulled with a `.case-conflict-...` suffix?

That's not happening now. It's on the table to maybe do it, but nothing is decided or planned on that.

For me to approve this PR, the "winning stuff" has to be removed. However that explicitly means that if anyone else is happy with leaving the "winning" stuff in there, I encourage that person to approve and merge.

acolomb: Sorry if I complicated this matter beyond what is necessary.  Details may seem irrelevant to you, with a deep background on how everything works.  Most users don't have that and will report bugs for "works as designed" behavior.

I'm okay with skipping further discussion.  I requested a review to make sure there are no factually **wrong** details in here, which I understood is the case.  Any correct documentation is better than none.

acolomb:
Would anyone mind if we moved the [conflict handling](https://docs.syncthing.net/users/faq.html#what-if-there-is-a-conflict) FAQ over here as well?  Seems strange to go into details on one special type of conflict, while the more general info is only in a FAQ.

users/syncing.rst
tomasz1986: > > Will case conflicts also result in the "losing" file being pulled with a `.case-conflict-...` suffix?
> 
> That's not happening now. It's on the table to maybe do it, but nothing is decided or planned on that.

Just a side note, but this would be really nice to have though, as in other conflicting cases Syncthing solves the conflicts automatically by renaming the files (or even overwrites them with `<maxConflicts>0</maxConflicts>`), but with a case conflict, the files get stuck waiting for a manual user intervention, which the user may not even be aware of, unless they open the GUI.

Having this would also make the "winning" problem not a big deal anymore, I guess.

acolomb:
@tomasz1986 since you spent some time investigating the corner cases, does this description explain your findings? Is anything missing or too much in your opinion?

users/syncing.rst
```diff
@@ -79,6 +79,39 @@ synchronize the file. The block lists are compared to build a list of needed
...
+In order to resolve such a case conflict situation, you need to decide
+on a consistent file name and **manually** enforce that across all
+involved, case-insensitive devices.
+
+This cautious behavior tries to save you from possible data loss
+caused by different files overwriting each other's contents.  That
+could have happened before version 1.9.0, where the same file would
+erroneously be accessed under two case-differing file names.
+
+All this does not concern the folder "root" path, but only relative
```
tomasz1986, line 112:
```suggestion
All this does not concern the folder path itself, but only relative
```
I would avoid using the word "root", as this whole case-sensitivity question is mostly related to Windows, and the concept of "root" does not exist there, i.e. a typical Windows users does not understand what a "root path" is.

acolomb: Thanks for the review, much appreciated.  There are at least eight uses of the term "root path", "folder root" or similar in the docs already.  I tried to align with that, as I had no better alternative idea.  Will remove the quotes though, for consistency.

The "double space after full stop" is a historic debate with associated scientific research, leaning toward double spaces at least for monospaced fonts. Code and RST markup is usually displayed in such a way.  I mainly go by the [GNU Coding Standards](https://www.gnu.org/prep/standards/standards.html#Comments) and enjoy the benefit of automatic re-wrapping in Emacs.  So yes, on purpose :-)

acolomb:
Merging this in its current form, as there were no further objections after @imsodin decided to limit his effort (no offense).

Having any (correct) documentation is better than none, but I'm open to further suggestions for follow-up PRs. Thanks for all reviews!

PR #576, opened 2020-10-12
Add caddy2 configuration file

rjpruitt16:
https://github.com/syncthing/syncthing/issues/6992

### DIFF
--- a/users/reverseproxy.rst
+++ b/users/reverseproxy.rst
@@ -68,6 +68,16 @@ Caddy
         write none
         header none
     }
+    
+Caddy
+~~~~~
+
+.. code-block:: none
+
+    :2016
+
+    reverse_proxy 127.0.0.1:8384
+
 
 Folder Configuration
 --------------------
--- a/users/reverseproxy.rst
+++ b/users/reverseproxy.rst
@@ -58,7 +58,7 @@ Caddy
 ~~~~~
 
 .. code-block:: none
-    
+
     proxy /syncthing localhost:8384 {
         transparent
     }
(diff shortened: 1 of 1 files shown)


hildebro (approved):
Can confirm this works on a clean installation.

calmh:
I think for this to be useful it should at minimum map the UI to a subpath like /syncthing, as the Caddy 1 example does. It should maybe also not set the port to 2016.

rjpruitt16:
Should the port be deleted. I thought it would convey more information if set. I could also put example.com.

rjpruitt16:
:)

imsodin:
> Should the port be deleted. I thought it would convey more information if set. I could also put example.com.

I don't know caddy, but looking at the example for v1 and Jakobs response above: Yes, to get it in line with the other example.

Also there's now two identical "Caddy" subtitles. Either put both examples under one subtitle and specify them as v1 and v2 in text or have "Caddy v1" and "Caddy v2" (except that Caddy might not use "v1" notation but something else, use whatever is common).

rjpruitt16:
any more feedback

users/reverseproxy.rst
```diff
@@ -68,6 +68,13 @@ Caddy
         write none
         header none
     }
+    
+    caddy2
```
imsodin, line 72: That's part of the code block - probably shouldn't be, right?

rjpruitt16: I wanted to follow your advice of combining caddy into 1 subtitle. I took a second look, and it did look a little weird. I separated them into different subtitles

PR #579, opened 2020-10-20
users/config: Clarify config and GUI input difference for listen address

DavidFair:
Clarifies that configuration files should use multiple XML tags rather
than a command seperated list which is GUI specific.

This caught me out earlier after following the docs and required some digging through the forums to figure it out.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -872,7 +872,7 @@ setLowPriority
 Listen Addresses
 ^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
So to me it seems like the meaning hasn't changed at all, so I am not sure how it makes it more clear.

DavidFair:
The original states:
>`you *can* have multiple tags` 

So I went ahead (skimming admittedly) and used commas as *can* implied that it would accept other forms such as commas used in the GUI too

AudriusButkevicius:
Is there a reason why you removed the space between the addresses?

DavidFair:
An unintentional formatting change from my local .rst rules - let me ammend my commit ....

PR #580, opened 2020-10-22
rest: Document new /rest/config endpoints

imsodin:
Documents https://github.com/syncthing/syncthing/pull/7001

The descriptions of the different endpoints are a bit repetitive, even after collecting similar ones together. The main purpose is to mention all available endpoints and which methods they support. Maybe just enlisting all the options as keywords, without actual text would do just as well (?).

I purposely didn't include any examples: Those are just bound to get out-of-date and the user may as well just try it out to see what it looks like.

### DIFF
--- a/rest/config.rst
+++ b/rest/config.rst
@@ -1,0 +1,57 @@
+Config Endpoints
+================
+
+These endpoints facilitate access and modification of the :ref:`configuration
+<config>` in a granular way. Config sent to the endpoints must be in the same
+format as returned by the corresponding GET request. When posting the
+configuration succeeds, the posted configuration is immediately applied, except
+for changes that require a restart. Query :ref:`rest-config-insync` to check if
+a restart is required.
+
+For all endpoints supporting ``PATCH``, it takes the existing config and
+unmarshals the given JSON object on top of it. This means all child objects will
+replace the existing objects, not extend them. For example for
+``RawListenAddresses` in options, which is an array of strings, sending
+``{RawListenAddresses: ["tcp://10.0.0.2"]`` will replace all existing listen
+addresses.
+
+.. note::
+   Changing config in quick succession can be problematic. The changes may not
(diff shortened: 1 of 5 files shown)


rest/config.rst
```diff
@@ -0,0 +1,57 @@
...
+For all endpoints supporting ``PATCH``, it takes the existing config and
+unmarshals the given JSON object on top of it. This means all child objects will
+replace the existing objects, not extend them. For example for
+``RawListenAddresses` in options, which is an array of strings, sending
+``{RawListenAddresses: ["tcp://10.0.0.2"]`` will replace all existing listen
+addresses.
+
+.. note::
+   Changing config in quick succession can be problematic. The changes may not
+   be applied in order, which might change the outcome. It is advicabable to
```
calmh, line 20: We hold some sort of config lock on the API side, and we wait for the config change to commit (right?) so I think this should really be fine and not require a warning? I mean if this is the case it feels like a bug...

The problem we do have is that if you do two concurrent changes there's no way to know which gets the lock and which one has to wait, so the order is undefined -- but that's because they were concurrent, if you just do your config changes serially I think it should be OK at any rate

rest/config.rst
```diff
@@ -0,0 +1,57 @@
+Config Endpoints
```
calmh, line 1: Should be an `.. addedversion:` or something here I think

calmh (commented):
writing docs good ❤️

rest/config.rst
imsodin: Indeed we do have that lock. Mentioning this, I wasn't really thinking about that lock enough and rechecked if I missed it somewhere: https://github.com/syncthing/syncthing/pull/7053. I removed the note alltogether because I think it's common sense that you shouldn't do config changes that might conflict concurrently.

calmh (approved):

PR #583, opened 2020-10-27
Add FPM to requirements for building debs

DavidFair:
Adds FPM into the list of requirements when building a .deb package.

This was missing and required me to dig through previous Github issues
to make sure that it was the correct package before installing it
locally.

### DIFF
--- a/dev/building.rst
+++ b/dev/building.rst
@@ -31,6 +31,9 @@ Prerequisites
 -  The latest stable version of Go. Earlier releases may work, but we recommend
    always using the latest stable version. At the time of writing this is **Go 1.14**.
 -  Git
+-  If you want to build Debian packages FPM is required. See FPM's
+   `installation information <https://fpm.readthedocs.io/en/latest/installing.html>`__.
+
 
 If you're not already a Go developer, the easiest way to get going
 is to download the latest version of Go as instructed in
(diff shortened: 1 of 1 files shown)


dev/building.rst
```diff
@@ -31,6 +31,9 @@ Prerequisites
 -  The latest stable version of Go. Earlier releases may work, but we recommend
    always using the latest stable version. At the time of writing this is **Go 1.14**.
 -  Git
+-  If you are building packages FPM is required, installation information
+   `here. <https://fpm.readthedocs.io/en/latest/installing.html>`__
```
calmh, line 35:
```suggestion
-  If you want to build Debian packages using our scripts FPM is required, see their
   `installation information <https://fpm.readthedocs.io/en/latest/installing.html>`__.
```

dev/building.rst
```diff
@@ -112,6 +115,10 @@ The following ``build.go`` subcommands and options exist.
 ``go run build.go test``
   Runs the tests.
 
+``go run build.go deb``
+  Creates a Syncthing deb installer in the current directory. Requires FPM
```
calmh, line 119:
```suggestion
  Creates a Debian package in the current directory. Requires FPM
```

calmh (commented):
Minor tweaks, but good point to note this down

DavidFair:
Sorry about the delay - and thanks for the suggestions I think it's a lot clearer.

I've amended my commit if you want to check again. Just to make you aware, one difference is instead of "see their installation instructions" which felt a bit clunky, I've used "See FPM's installation instructions" instead. 
However, feel free to come back to me as I'm happy to take more suggestions

calmh (approved):

calmh:
👍

PR #584, opened 2020-10-30
users/config: Improve modTimeWindowS explanation

tomasz1986:
Reword the explanation to make it more specific. Add an example of a
file system (FAT), where the setting is applicable. Add a warning about
the necessecity of setting the value before adding files to the folder.
Explicitly mention the default values depending on the OS.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -354,9 +354,12 @@ copyOwnershipFromParent
     Requires running Syncthing as privileged user, or granting it additional capabilities (e.g. CAP_CHOWN on Linux).
 
 modTimeWindowS
-    Allowed modification timestamp difference when comparing files for equivalence.
-    To be used on systems that have unstable modification timestamps, that might change after being observed after
-    the last write operation. Used in Android only.
+    Allowed modification timestamp difference when comparing files for
+    equivalence. To be used on file systems which have unstable or
+    imprecise modification timestamps (e.g. FAT) that might change after
+    being recorded during the last write operation. Must be set before
+    adding files to the folder in order to take effect. Usually defaults
+    to 2 on Android, and always to 0 elsewhere.
 
 maxConcurrentWrites
     Maximum number of concurrent write operations while syncing. Defaults to 2. Increasing this might increase or
--- a/users/config.rst
+++ b/users/config.rst
@@ -356,10 +356,10 @@ copyOwnershipFromParent
 modTimeWindowS
     Allowed modification timestamp difference when comparing files for
     equivalence. To be used on file systems which have unstable or
-    imprecise modification timestamps (e.g. FAT) that might change after
-    being recorded during the last write operation. Must be set before
-    adding files to the folder in order to take effect. Usually defaults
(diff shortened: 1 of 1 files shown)


users/config.rst
```diff
@@ -354,9 +354,12 @@ copyOwnershipFromParent
...
 
 modTimeWindowS
-    Allowed modification timestamp difference when comparing files for equivalence.
-    To be used on systems that have unstable modification timestamps, that might change after being observed after
-    the last write operation. Used in Android only.
+    Allowed modification timestamp difference when comparing files for
+    equivalence. To be used on file systems which have unstable or
+    imprecise modification timestamps (e.g. FAT) that might change after
+    being recorded during the last write operation. Must be set before
+    adding files to the folder in order to take effect. Usually defaults
```
calmh, line 361: I know the context of this addition, but I don't think "must be set before adding files to the folder to take effect" increases understanding much here... Any config change only affects what happens after it is made, obviously, but this somehow implies it won't work if set on an existing folder.

Perhaps something in the first sentence about the comparison being used to decide whether a file has changed and needs to be rehashed? Then it might be more clear that changing this option doesn't un-rehash and un-change the file...

tomasz1986: Yeah, I know that the sentence is vague. I could not come up with anything better.

What do you think about something like this?

```Must be set before adding files to the folder in order to take effect on those files.```

Otherwise I will try to think about how to rewrite the sentences to explain the point better.

Comment (calmh) [reply users/config.rst:361]: I don't particularly like that either, because it makes it sound like the value is somehow locked in per file once we've seen it the first time... while in reality it simply applies each and every time we look at a file and consider the modification timestamp.

Comment (tomasz1986) [reply users/config.rst]: I reworded the explanation a little bit, and also removed the controversial line. I agree that it will be better to explain this behaviour with regards to RO folders in general rather than having separate explanations under each setting.

File: users/config.rst [hunk 2]
```diff
@@ -354,9 +354,12 @@ copyOwnershipFromParent
     Requires running Syncthing as privileged user, or granting it additional capabilities (e.g. CAP_CHOWN on Linux).
 
 modTimeWindowS
-    Allowed modification timestamp difference when comparing files for equivalence.
-    To be used on systems that have unstable modification timestamps, that might change after being observed after
-    the last write operation. Used in Android only.
+    Allowed modification timestamp difference when comparing files for
+    equivalence. To be used on file systems which have unstable or
+    imprecise modification timestamps that might change after being
```
imsodin, line 359: Imprecise timestamps are not a problem, only unstable ones. For an attempted explanation see: https://forum.syncthing.net/t/why-default-modtimewindows-to-2-for-fat-only-on-android/16256/2

calmh (approved):

imsodin (approved):

imsodin:
Why is there the "(#582)" in the commit message? I don't see the connection.

tomasz1986:
> Why is there the "(#582)" in the commit message? I don't see the connection.

There is no connection, just a mistake on my part. Fixed!

PR #588, opened 2020-11-15
css: Add margins between list items with paragraphs

tomasz1986:
The default theme removes margins after last paragraph inside a list
item, but this removes any space between two adjacent list items. Thus,
override the default CSS keeping the default margin of 1em, and only
remove the margin in the last list item instead.


tomasz1986:
Before:

image

After:

image

PR #595, opened 2020-12-29
users: Document advanced config option remoteGUIport.

acolomb:
Implemented in https://github.com/syncthing/syncthing/pull/7017.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -81,7 +81,7 @@ The following shows an example of the default configuration file (IDs will diffe
             <maxSendKbps>0</maxSendKbps>
             <maxRecvKbps>0</maxRecvKbps>
             <maxRequestKiB>0</maxRequestKiB>
-            <remoteGUIPort>8384</remoteGUIPort>
+            <remoteGUIPort>0</remoteGUIPort>
         </device>
         <gui enabled="true" tls="false" debugging="false">
             <address>127.0.0.1:8384</address>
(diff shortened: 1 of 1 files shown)


tomasz1986:
Just for the record, this is going to be available from v1.12.1.

users/config.rst
```diff
@@ -81,6 +81,7 @@ The following shows an example of the default configuration file (IDs will diffe
             <maxSendKbps>0</maxSendKbps>
             <maxRecvKbps>0</maxRecvKbps>
             <maxRequestKiB>0</maxRequestKiB>
+            <remoteGUIPort>8384</remoteGUIPort>
```
tomasz1986, line 84: The value should be

```
<remoteGUIPort>0</remoteGUIPort>
```

here, as this is supposed to show the default configuration. The default value is `0`.

users/config.rst
```diff
@@ -408,6 +409,7 @@ Device Element
         <maxSendKbps>0</maxSendKbps>
         <maxRecvKbps>0</maxRecvKbps>
         <maxRequestKiB>0</maxRequestKiB>
+        <remoteGUIPort>8384</remoteGUIPort>
```
tomasz1986, line 412: I would change this to `<remoteGUIPort>0</remoteGUIPort>`, but also add `<remoteGUIPort>8384</remoteGUIPort>` to the next device config below. This way we have both the default and a custom values shown.

acolomb:
Thanks @tomasz1986 for the review.  Both points changed as suggested.

As v1.12.1 will be tagged in a couple of days, this should be merged before then, right?

tomasz1986:
> As v1.12.1 will be tagged in a couple of days, this should be merged before then, right?

The setting is already in the RC, so I would say yes 🙂.

tomasz1986 (approved):

PR #605, opened 2021-02-11
users/faq: Add a warning about mixing nested folders with Introducer

tomasz1986:
Add a warning to the FAQ about using the Introducer feature together
with nested folders, as that can lead to Syncthing automatically sharing
both the nested folder itself and its parent between the same devices.

Even though Syncthing can deal with such situations and still properly
synchronise the files, it leads to an unnecessary overhead and creates
temporary files, which hang around for 24 hours (by default) for no
purpose.

Ref: https://github.com/syncthing/syncthing/issues/7338

### DIFF
--- a/users/faq-parts/usage.rst
+++ b/users/faq-parts/usage.rst
@@ -27,11 +27,16 @@ See the previous question.
 Am I able to nest shared folders in Syncthing?
 ----------------------------------------------
 
-Sharing a folder that is within an already shared folder is possible, but it has
-its caveats. What you must absolutely avoid are circular shares. This is just
-one example, there may be other undesired effects. Nesting shared folders is not
-supported, recommended or coded for, but it can be done successfully when you
-know what you're doing - you have been warned.
+Sharing a folder that is within an already shared folder is possible,
+but it has its caveats. What you must absolutely avoid are circular
+shares. You should also be careful when using :ref:`introducer`, which
+can lead to sharing both the nested folder and its parent between the
+same devices. These are just two examples, there may be other undesired
+effects. Nesting shared folders is not supported, recommended or coded
+for, but it can be done successfully when you know what you're doing -
+you have been warned.
+
+
 
 How do I rename/move a synced folder?
 -------------------------------------


imsodin:
I personally wouldn't add any text here. The only message to bring across is "nested devices are potentially dangerous - tread with care". Adding another bit of complexity might just either confuse people or even let them think that as they don't use the introducer, they are fine. And where do you draw the line: What about auto-accepting folders?

tomasz1986:
Well, I think that it would be good to have the known consequences listed somewhere though. This is the only place in the Docs, where nested folders are actually discussed, so I have added this information here.

> let them think that as they don't use the introducer, they are fine.

The very next sentence is `These are just two examples, there may be other undesired effects.`, so I would not really think so 😉. But again, I know that FAQ may not be the best location for a more detailed explanation, yet there is no other right now.

imsodin:
I remember having this discussion when we last re-worked this FAQ entry: It is exactly not about an attempt to list possible problems. Which your disclaimer sentence exemplifies - so why do it anyway? It is about making the user aware, that it has potential to cause problems and they need to think about it themselves.  
Anyway, it's just my opinion, and not a strong one - if anyone else thinks this is super helpful they can approve and merge.

tomasz1986:
Yeah, I am not sure myself now. I had originally thought about adding a note under https://docs.syncthing.net/users/introducer.html#introducer, but then there are other things like Versioning, etc., which do not play nicely with nested folders.

On the one hand, I agree that one possible approach is to simply not list any specific cases, stick to the general warning, and let the users decide whether they want to take the risk. On the other hand, it is really difficult to predict all the consequences, especially by new users, and once they hit them, it is usually too late, i.e. something has already got ruined in the process.

imsodin:
> [...] On the other hand, it is really difficult to predict all potential consequences, especially by new users, and once they hit them, it is usually too late, i.e. something has already got ruined in the process.

That's what the main message is aimed at:

> Nesting shared folders is not supported, recommended or coded
for, but it can be done successfully when you know what you're doing -
you have been warned.

Maybe make it bold if you think it isn't clear enough yet :)

tomasz1986:
> Maybe make it bold if you think it isn't clear enough yet :)

The problem, at least in my experience, is that people generally ignore such general warnings. I would say that the current one could be made a little stronger. I will think about how to achieve it.

Let us close this one then.

PR #606, opened 2021-02-11
users/versioning: Improve external versioning example script for Windows

tomasz1986:
Enable UTF-8 support, as otherwise the script will fail with non-native
code pages, e.g. when using non-Latin characters in the English version
of Windows, etc. Use quotes with the "set" command or the script will be
unable to deal with funky characters like "&" in file and folder names.
Use "md" instead of "mkdir", because it is shorter. Lastly, use "rem"
instead of "::" for comments, as this is the only way to comment that
is recommended by Microsoft (and "::" is just a hack in itself).

### DIFF
--- a/users/versioning.rst
+++ b/users/versioning.rst
@@ -145,21 +145,24 @@ behavior as mentioned above. I created the following script and saved it as
 
     @echo off
 
-    :: We need command extensions for mkdir to create intermediate folders in one go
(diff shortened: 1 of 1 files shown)


tomasz1986:
There are still some more wacky characters in addition to `&` that will make the script fail (despite the quotes), but I cannot remember them right now. I will do another PR if they come to my mind later.

imsodin (approved):

PR #607, opened 2021-02-12
users/config: Add ID to each config option for easy referencing

tomasz1986:
Assign an ID to each config option, so that we can easily refer with an
URL to a specific one when needed. Use more specific IDs than before in
order to prevent conflicts between same options under different elements,
e.g. folder-id and device-id should be separate.

### DIFF
--- a/_static/custom.css
+++ b/_static/custom.css
@@ -77,3 +77,8 @@ div.contents {
 div.contents li > p:last-child {
   margin-bottom: inherit;
 }
+
+/* Add spacing between each setting in users/config */
+dl {
+  margin-top: 1em;
+}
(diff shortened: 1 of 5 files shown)


tomasz1986:
The CSS change is required to make everything look the same as before, because Sphinx formats each element as a separate `dl`. I have tried hard to make it add IDs to the existing `dt`s instead, but it just does not seem possible with RST, i.e. it is impossible to add an ID to a specific element and also give the ID a different value than the element's text.

st-review:
:robot: *beep boop*

AudriusButkevicius:
Perhaps I should have left it closed, or do you plan on finishing this?

tomasz1986:
> Perhaps I should have left it closed, or do you plan on finishing this?

I do, but probably only in late June/July, when I have more free time to spend. Shouldn't the stalled pull requests re-open themselves if there are new commits anyway?

AudriusButkevicius:
Idk, but I guess I'll close it and we will see.

acolomb:
Superseded by #710.

PR #608, opened 2021-02-12
users/config: Add a warning to ignoreDelete

tomasz1986:
Add a warning similar to disableFsync clearly stating that this option
should be used at one's own risk.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -332,7 +332,7 @@ order
 
 ignoreDelete
     .. warning::
-        This has irreversible consequences - use at your own risk.
+        Enabling this is highly not recommended - use at your own risk.
 
     When set to true, this device will pretend not to see instructions to
     delete files from other devices.
--- a/users/config.rst
+++ b/users/config.rst
@@ -334,6 +334,9 @@ ignoreDelete
     When set to true, this device will pretend not to see instructions to
     delete files from other devices.
 
+    .. warning::
+        This has irreversible consequences - use at your own risk.
+
 scanProgressIntervalS
     The interval with which scan progress information is sent to the GUI. Zero
     means the default value (two seconds).
--- a/users/config.rst
+++ b/users/config.rst
@@ -331,11 +331,11 @@ order
     the 1 KB becomes known to the pulling device.
 
 ignoreDelete
-    When set to true, this device will pretend not to see instructions to
-    delete files from other devices.
-
     .. warning::
         This has irreversible consequences - use at your own risk.
+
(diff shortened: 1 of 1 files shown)


imsodin:
I'd add something explicitly saying "not recommended to use" or just replace the entire description with a link to https://docs.syncthing.net/advanced/folder-ignoredelete.html, which already has a warning.

tomasz1986:
Done. I think that having the warning here is useful though, so that people can see it immediately and hopefully get discouraged from using it without having to go deeper into the Docs for a more detailed explanation.

AudriusButkevicius:
Tbh, I would have made it more scary, saying this will corrupt the database and lead to permanently out of sync status etc.

tomasz1986:
> Tbh, I would have made it more scary, saying this will corrupt the database and lead to permanently out of sync status etc.

I think that this may be too long for the general config page, but could go into warning at https://docs.syncthing.net/advanced/folder-ignoredelete.html. I will try to add it shortly.

users/config.rst
```diff
@@ -331,6 +331,9 @@ order
     the 1 KB becomes known to the pulling device.
 
 ignoreDelete
+    .. warning::
+        Enabling this is highly not recommended - use at your own risk.
```
AudriusButkevicius, line 335: Enabling this will willingly corrupt the database and lead to permanently out of sync state - use at your own risk

imsodin:
I am all for making it super scary - however where exactly does db corruption come into this? I am only aware of  making the device out-of-sync forever, not of any db corruption.

AudriusButkevicius:
It's corrupt in terms of state of the world as you are missing delete entries, and potentially have gaps in the sequences?

imsodin:
You have the full view of the world, you just choose to ignore the deletions when pulling. Thus leaving you out-of-sync on those items compared to the world.

AudriusButkevicius:
Huh, I went looking for the code in the protocol package where it used drop the fileinfos before they even reached the model, and to my surprise it's no longer there.

It seems the deletes are not dropped anymore, they are just ignored in the Need part, which I assume means that if you disable the flag, things should go back to normal?

imsodin:
Basically yes.

tomasz1986:
This information does make the new PR (https://github.com/syncthing/docs/pull/611) kind of irrelevant, right? I mean, if the database is not getting corrupted, and the state is reversible, then ultimately what I wrote there is simply not true :smirk:.

AudriusButkevicius:
Agree.

PR #612, opened 2021-02-14
users/versioning: Add a PowerShell example to External Versioning

tomasz1986:
Add two code snippets that use PowerShell in order to move versioned
files directly to the Recycle Bin, as if they have been deleted using
the Windows Explorer.

### DIFF
--- a/users/versioning.rst
+++ b/users/versioning.rst
@@ -142,8 +142,11 @@ The script will then move the file in question to
 ``~/.trashcan/docs/letter.txt``, replacing any previous version of that letter
 that may already have been there.
 
-Example for Windows
-~~~~~~~~~~~~~~~~~~~
+Examples for Windows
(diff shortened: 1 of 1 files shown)


tomasz1986:
Tested under both Windows 7 and 10, and also with funky and multilingual filenames.

tomasz1986:
I have modified the commit to include the script in full, following the existing Batch example, instead of an embedded command. If someone has a need to embed the script directly in Syncthing, they probably know how to change the formatting to fit the commands in a single line, etc.

The first script is very basic, as it simply sends everything to the Recycle Bin. This, however, can easily end up with thousands versions of the same file, if the file is being modified often. That is why I have also added an alternative version that only sends actually deleted files to the Bin, and permanently deletes modified ones.

calmh (approved):

PR #616, opened 2021-02-23
fix: new /rest/config/insync api path

evrins:
fix new /rest/config/insync api path

https://docs.syncthing.net/rest/system-config-insync-get.html 

should use /rest/config/insync instead

### DIFF
--- a/rest/config.rst
+++ b/rest/config.rst
@@ -26,7 +26,7 @@ addresses.
 
 .. _rest-config-insync:
 
-/rest/system/config/insync
+/rest/config/insync
 --------------------------
 
 ``GET`` returns whether the config is in sync, i.e. whether the running configuration is


calmh:
The new path exists, but the old path is still used and mentioned in a lot of other places. I'm not sure we should just switch to the new one here. (Or even if the new one should exist under this name at all...  Since it's new, it might have been something more descriptive like `/rest/config/restart-required` or something.)

imsodin:
This PR just changes a typo of mine when introducing the new `/rest/config/...` endpoints - the `/rest/config/insync` endpoint already exists unfortunately. If we want to switch to something more descriptive like the proposed `/rest/config/restart-required` that needs to happen on the go side of things too. Given how new `/rest/config/insync` is I'd just not mention it in the docs at all then.

imsodin:
@evrins I added a `/rest/config/restart-required` endpoint. Could you please change this PR to use that one instead.

st-review:
:robot: *beep boop*

PR #617, opened 2021-02-25
Mention QUIC related port

bt90:


users/firewall.rst
```diff
@@ -45,11 +45,12 @@ Local Firewall
 --------------
 
 If your PC has a local firewall, you will need to open the following ports for
-incoming and outgoing traffic:
+incoming and outgoing traffic, assuming that you configured port **22000** in
```
calmh, line 48: Since this port isn't visible in the config by default, maybe hedge this like the previous mention ("unless you changed it from the default" or something)?

bt90: I moved it to a note

calmh (approved):

PR #618, opened 2021-02-25
Describe QUIC address config

bt90:


users/config.rst
```diff
@@ -498,33 +498,34 @@ From following child elements at least one ``address`` child must exist.
...
     Contains an address or host name to use when attempting to connect to this device.
     Entries other than ``dynamic`` must be prefixed with ``tcp://`` (dual-stack),
-    ``tcp4://`` (IPv4 only) or ``tcp6://`` (IPv6 only). Note that IP addresses need
-    not use tcp4/tcp6; these are optional. Accepted formats are:
+    ``tcp4://`` (IPv4 only), ``tcp6://`` (IPv6 only), ``quic://`` (dual-stack),
+    ``quic4://`` (IPv4 only) or ``quic6://`` (IPv6 only).
+    Note that IP addresses need not use tcp4/tcp6/quic4/quic6; these are optional. Accepted formats are:
 
-    IPv4 address (``tcp://192.0.2.42``)
+    IPv4 address (``tcp://192.0.2.42`` / ``quic://192.0.2.42``)
```
calmh, line 505: I don't think we need to repeat quic for all of these, since we just listed all the schemes above. Perhaps add a note to say explicitly that any of the supported schemes can be used in place of tcp?

AudriusButkevicius: Likewise, we have relays, which is probably something more interesting to configure directly, yet we don't see the need to copy that for every address variation.

calmh (approved):

PR #620, opened 2021-03-01
advanced/option-usebadger: Deprecate and warn against using Badger

tomasz1986:
Badger support is being removed as of Syncthing 1.14.x. For this reason,
add a note about the environment variable being deprecatd, and also warn
against using it.

### DIFF
--- a/advanced/option-usebadger.rst
+++ b/advanced/option-usebadger.rst
@@ -7,7 +7,8 @@ Environment variable USE_BADGER
 *Removed in version 1.14.0.*
 
 .. warning:: Badger experiment has been removed and the environment
-   variable is no longer supported.
+   variable is no longer supported. Version 1.14.0 is that last one that
+   allows switching back from Badger to LevelDB.
 
 Syncthing traditionally uses the LevelDB database (`syndtr/goleveldb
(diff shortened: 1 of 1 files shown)


imsodin:
It's actually been deprecated in v1.12.0 (https://github.com/syncthing/syncthing/pull/7089) and what happens now is the ability to switch back from badger to leveldb is also removed. So I'd either remove the documentation fully, or state that v1.14.0 is the last version that allows switching back. Though really that's a hypothetical scenario: No-one is using the experiment, definitely not someone that's on some old version and never updates.

tomasz1986:
What does actually happen if someone is using Badger and tries to upgrade to 1.14.0? Does Syncthing not start at all and panics?

imsodin:
Ignores the badger db, i.e. creates a new (level)db (or uses an old, already existing one).

tomasz1986:
> So I'd either remove the documentation fully, or state that v1.14.0 is the last version that allows switching back.

I have fixed the explanation. I would personally keep it, at least until Badger has been removed from the code completely, but also if there are other database implementations added in the future, I think that at least parts of the text could be simply reused.

calmh:
Badger has been completely removed, and the docs should be as well.

calmh:
Removed.

PR #626, opened 2021-03-11
user/config: Fix incorrect listen address

imsodin:
I fixed the incorrect listen address, discovered in https://forum.syncthing.net/t/syncthing-v1-14-0-sometimes-fails-to-start-address-already-in-use-ubuntu-20-04/16481/4 by @bt90 . I also added a note saying to **not** use the config examples given directly.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -64,7 +64,14 @@ The database contains the following files:
 Config File Format
 ------------------
 
-The following shows an example of the default configuration file (IDs will differ):
+The following shows an example of a default configuration file (IDs will differ):
+
+
+.. note::
+   The config examples are present for illustration. Do **not** copy them
+   entirely to use as your config. They are likely out-of-date and the values
+   might not correspond any more to the defaults.
+
 
 .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


users/config.rst
```diff
@@ -64,7 +64,14 @@ The database contains the following files:
...
 ------------------
 
-The following shows an example of the default configuration file (IDs will differ):
+The following shows an example of a default configuration file (IDs will differ):
+
+
+.. note::
+   The config examples are present for illustration. Do **not** copy them
+   entirely to use as your config. They are likely out-of-date and the values
+   might not correspond any more to the defaults.
```
AudriusButkevicius, line 73: The "any more" part feels grammatically incorrect and not sure if it adds any value (if its out of date, its out of date).

AudriusButkevicius (approved):

users/config.rst
tomasz1986: How about "the values may no longer correspond to the defaults"?

imsodin, line 73:
```suggestion
   may no longer correspond to the defaults.
```

PR #630, opened 2021-04-04
users/contrib: Add another GNOME Shell extension

2nv2u:
Added another shell extension

### DIFF
--- a/users/contrib.rst
+++ b/users/contrib.rst
@@ -54,6 +54,10 @@ Linux
 
   A GNOME Shell extension displaying a Syncthing status icon in the top bar.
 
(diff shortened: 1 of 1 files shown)


users/contrib.rst
```diff
@@ -54,6 +54,10 @@ Linux
 
   A GNOME Shell extension displaying a Syncthing status icon in the top bar.
 
+- `Syncthing Indicator <https://extensions.gnome.org/extension/1070/syncthing-indicator/>`_
+
+  Shell indicator for starting, monitoring and controlling the Syncthing daemon using SystemD.
```
imsodin, line 59:
```suggestion
  A GNOME Shell indicator for starting, monitoring and controlling the Syncthing daemon using SystemD.
```

PR #633, opened 2021-04-16
Add information about '.' being a valid marker name.

jtagcat:
https://forum.syncthing.net/t/read-only-send-only/16581/4

### DIFF
--- a/users/faq-parts/usage.rst
+++ b/users/faq-parts/usage.rst
@@ -22,7 +22,7 @@ you know will always exist in the folder.
 I really hate the ``.stfolder`` directory, can I remove it?
 -----------------------------------------------------------
 
-See the previous question.
+See the previous question. It is heavily advised against, but `.` is a valid marker (alias) name as well.
 
 Am I able to nest shared folders in Syncthing?
 ----------------------------------------------


imsodin (changes requested):
So far I am happy to pretend I never heard about this, but I personally don't want this documented. Firstly because the fewer people do it the better, secondly if people do it, I think it's a bit less likely that they shoot themselves in the foot if they needed to figure it out on their own or find some reference on the forum and lastly because if we ever decide to close this loophole (that's what it is), having it documented makes it harder.

tomasz1986:
Yeah, even though I am the one mentioning the "hack" in the linked post, this probably belongs to the forum and power users, who are into experimentation and such.

By the way, I haven't tested it, but I would guess that `..` will work too, won't it?

jtagcat:
> So far I am happy to pretend I never heard about this, but I personally don't want this documented. Firstly because the fewer people do it the better, secondly if people do it, I think it's a bit less likely that they shoot themselves in the foot if they needed to figure it out on their own or find some reference on the forum and lastly because if we ever decide to close this loophole (that's what it is), having it documented makes it harder.

Honestly, I'm on the same opinion. At the same time I'd prefer 'advanced dangerous docs' to 'out of context forum post found with search copy-pasta'.

Of course, closing the loophole, and changing it to an option under the advanced menu would be preferred.

PR #636, opened 2021-04-23
users/faq: Update 'Global Changes' -> 'Recent Changes' button name

sainaen:
This is a follow-up to the change requested in syncthing/syncthing#4326, merged in v0.14.41.


imsodin (approved):
Thanks!

A follow-up to v0.14.41 :D  
(Just in case it's not obvious, I am not making fun of you, but of us, the project, respectively the situation).

PR #639, opened 2021-05-08
users: Add a rough architecture overview to the man page.

acolomb:
Insert one paragraph after the overview description to give a starting point for CLI users reading the man page and trying to discover the GUI.

This could be useful for people installing the DEB package for example, then seeing only an installed daemon binary and a man page.  This will point them to the Web UI in case they missed the Getting Started Guide. As suggested [on the forum](https://forum.syncthing.net/t/great-technology-bad-user-interface/16781/4).

### DIFF
--- a/specs/untrusted.rst
+++ b/specs/untrusted.rst
@@ -1,9 +1,9 @@
-.. _untrusted:
-
 .. warning::
    This feature is currently in public testing and not yet recommended to be
    used for important data. Related UI controls are hidden by a feature flag -
    check the release notes for information on how to test it.
+
+.. _untrusted:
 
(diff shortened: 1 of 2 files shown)


acolomb:
Note there were warnings when building the HTML docs, not related to this change.  They are fixed in the second commit:

```
syncthing-docs/specs/bep-v1.rst:345: WARNING: Failed to create a cross reference. A title or caption not found: untrusted
syncthing-docs/users/untrusted.rst:158: WARNING: Failed to create a cross reference. A title or caption not found: untrusted
```

calmh (approved):

PR #640, opened 2021-05-14
users/untrusted: Fix grammatical error

ar-maged:


users/untrusted.rst
```diff
@@ -129,7 +129,7 @@ Caveats
 -------
 
 The untrusted device must not be able to compare encrypted data between
-different files and detect if data is equal, This means that blocks cannot as
+different files and detect if data is equal, this means that blocks cannot as
```
imsodin, line 132:
```suggestion
different files and detect if data is equal. This means that blocks cannot as
```

?

ar-maged: Even better 👍

imsodin:
Thanks.

PR #645, opened 2021-05-26
rest: Add DELETE methods for cluster/pending endpoints.

acolomb:
Proposed API to fix https://github.com/syncthing/syncthing/issues/7700.

Slighly improves the description of `PendingFoldersChanged` event along the way.

### DIFF
--- a/events/pendingfolderschanged.rst
+++ b/events/pendingfolderschanged.rst
@@ -7,9 +7,9 @@ PendingFoldersChanged
 
 Emitted when pending folders were added / updated (offered by some
 device, but not shared to them) or removed (folder ignored, dismissed
-or added or no longer offered from the remote device).  An entry
-without a ``device`` attribute means that the folder is no longer
-pending for any device.
+or added or no longer offered from the remote device).  A removed
+entry without a ``deviceID`` attribute means that the folder is no
+longer pending for any device.
 
 .. code-block:: json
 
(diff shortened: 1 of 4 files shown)


rest/cluster-pending-folders-delete.rst
```diff
@@ -0,0 +1,19 @@
...
+====================================
+
+.. versionadded:: 1.18.0
+
+Remove records about a pending folder, without ignoring it in the
+configuration.  Valid values for the ``folder`` and ``device``
+parameters are those from the corresponding
+:doc:`/rest/cluster-pending-folders-get` endpoint.  The ``device``
+parameter is optional and affects announcements of this folder from
+*any* device if omitted.
```
imsodin, line 11: Second part feels backwards: Either "and affects ... from the given device or *any* device if omitted." or just "and omitting it removes this folder from all devices."

rest/cluster-pending-folders-delete.rst
```diff
@@ -0,0 +1,19 @@
+DELETE /rest/cluster/pending/folders
+====================================
+
+.. versionadded:: 1.18.0
+
+Remove records about a pending folder, without ignoring it in the
+configuration.  Valid values for the ``folder`` and ``device``
```
imsodin, line 7: The ignoring in config part makes sense given the history, but seems out of context like this. I'd not mention it at all, or then add a full blown explanation that you should ignore devices instead, if you don't ever want to get pending entry again.

events/pendingfolderschanged.rst
```diff
@@ -6,8 +6,10 @@ PendingFoldersChanged
 .. versionadded:: 1.14.0
 
 Emitted when pending folders were added / updated (offered by some
-device, but not shared to them) or removed (folder ignored or added or
-no longer offered from the remote device).
+device, but not shared to them) or removed (folder ignored, dismissed
+or added or no longer offered from the remote device).  An entry
+without a ``device`` attribute means that the folder is no longer
+pending for any device.
```
imsodin, line 12: What is this about? I.e. why is the device removed respectively why is it any different than the existing "remove-events"?

imsodin (commented):
Endpoints look good, just some minor comments/questions.

events/pendingfolderschanged.rst
acolomb: This is a shortcut for when a pending folder is ignored locally.  Any pending entries with that folder ID should then be discarded, so instead of enumerating them all, only one entry with the `folderID` attribute is emitted.  That is handled in the GUI by removing a whole nested sub-object.

rest/cluster-pending-folders-delete.rst
acolomb: Committed the first suggestion.

rest/cluster-pending-folders-delete.rst
acolomb: Applied both suggestions.

acolomb:
Please check if all review comments have been addressed in a satisfactory way.  Ready to merge from my side.

imsodin (approved):

PR #648, opened 2021-06-01
Add section on setting up initial sync

roseeng:
As suggested on the forum (https://forum.syncthing.net/t/helping-first-sync-by-manual-copy/16858/6), here is a suggested contribution to the FAQ about "Can I help initial sync by copying files manually?".

Feel free to suggest or perform edits.

/Göran

### DIFF
--- a/users/faq-parts/usage.rst
+++ b/users/faq-parts/usage.rst
@@ -90,11 +90,11 @@ rebuild on remote ends, and then delete from A.
 Can I help initial sync by copying files manually?
 --------------------------------------------------
 
-If you have a large folder that you want to keep in sync over a not-so-fast network, and you have the possibility to move all files to the remote site in a faster manner, here is a procedure to follow to make sure that you don't confude Syncthing.
+If you have a large folder that you want to keep in sync over a not-so-fast network, and you have the possibility to move all files to the remote site in a faster manner, here is a procedure to follow to make sure that you don't confuse Syncthing.
 
 Create the folder on site 1, but don't share it with anyone.
-Create the folder on site 2, and this time copy the Folder Id from the folder on site 1. Normally you would always use the value that Syncthing supplies to you since it must be unique, but in this case we want the folders to be considered the same.
-Copy the files from site 1 to site 2 using normal file copy.
+Create the folder on site 2, and this time copy the Folder Id from the folder on site 1. Normally you would always use the value that Syncthing supplies to you - since it must be unique - but in this case we want the folders to be considered the same.
+Copy the files from site 1 to site 2 using normal file copy. If this takes a long time (perhaps travelling to site 2), it could be a good idea to make sure that the files on site 1 is not updated while you do this.
 
 After that rescan the folders on both sides so that Syncthing knows all the files. Then share them. Syncthing will exchange file information but existing files (or parts) will not be transferred.
 
--- a/users/faq-parts/usage.rst
+++ b/users/faq-parts/usage.rst
@@ -87,6 +87,19 @@ block) from A, and then as A gets rescanned remove the files from A.
 A workaround would be to copy first from A to B, rescan B, wait for B to
 rebuild on remote ends, and then delete from A.
 
+Can I help initial sync by copying files manually?
+--------------------------------------------------
+
+If you have a large folder that you want to keep in sync over a not-so-fast network, and you have the possibility to move all files to the remote site in a faster manner, here is a procedure to follow to make sure that you don't confude Syncthing.
+
+Create the folder on site 1, but don't share it with anyone.
+Create the folder on site 2, and this time copy the Folder Id from the folder on site 1. Normally you would always use the value that Syncthing supplies to you since it must be unique, but in this case we want the folders to be considered the same.
+Copy the files from site 1 to site 2 using normal file copy.
+
+After that rescan the folders on both sides so that Syncthing knows all the files. Then share them. Syncthing will exchange file information but existing files (or parts) will not be transferred.
+
+Always make sure that if you add many files to a folder that is used by Syncthing, that the folders are either paused or Syncthing is not running. Then run rescan.
+
 Is Syncthing my ideal backup application?
 -----------------------------------------
 
--- a/users/faq-parts/usage.rst
+++ b/users/faq-parts/usage.rst
@@ -92,9 +92,9 @@ Can I help initial sync by copying files manually?
 
 If you have a large folder that you want to keep in sync over a not-so-fast network, and you have the possibility to move all files to the remote site in a faster manner, here is a procedure to follow to make sure that you don't confuse Syncthing.
 
-Create the folder on site 1, but don't share it with anyone.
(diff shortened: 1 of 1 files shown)


users/faq-parts/usage.rst
```diff
@@ -87,6 +87,19 @@ block) from A, and then as A gets rescanned remove the files from A.
...
 rebuild on remote ends, and then delete from A.
 
+Can I help initial sync by copying files manually?
+--------------------------------------------------
+
+If you have a large folder that you want to keep in sync over a not-so-fast network, and you have the possibility to move all files to the remote site in a faster manner, here is a procedure to follow to make sure that you don't confuse Syncthing.
+
+- Create the folder on site 1, but don't share it with anyone.
+- Create the folder on site 2, and this time copy the Folder Id from the folder on site 1. Normally you would always use the value that Syncthing supplies to you - since it must be unique - but in this case we want the folders to be considered the same.
+- Copy the files from site 1 to site 2 using normal file copy. If this takes a long time (perhaps travelling to site 2), it could be a good idea to make sure that the files on site 1 is not updated while you do this.
```
calmh, line 97:
```suggestion
- Copy the files from site 1 to site 2 using normal file copy. If this takes a long time (perhaps travelling to site 2), it could be a good idea to make sure that the files on site 1 is not updated while you do this.
- Create the folder on site 2, and this time copy the Folder ID from the folder on site 1. Normally you would always use the value that Syncthing supplies to you - since it must be unique - but in this case we want the folders to be considered the same.
```

(switching order here so the files are in place when the folder is created)

users/faq-parts/usage.rst
```diff
@@ -87,6 +87,19 @@ block) from A, and then as A gets rescanned remove the files from A.
...
+
+After that rescan the folders on both sides so that Syncthing knows all the files. Then share them. Syncthing will exchange file information but existing files (or parts) will not be transferred.
```
calmh, line 99:
```suggestion
Now share the folder with the other device, on both sides. Syncthing will exchange file information but existing files (or parts) will not be transferred.
```

users/faq-parts/usage.rst
```diff
@@ -87,6 +87,19 @@ block) from A, and then as A gets rescanned remove the files from A.
...
+
+Always make sure that if you add many files to a folder that is used by Syncthing, that the folders are either paused or Syncthing is not running. Then run rescan.
+
```
calmh, line 102:
```suggestion
```

I don't think this is necessary (and anyway unrelated to the subject at hand).

roseeng: I agree, this was actually just copied from the reply I got on the forum

users/faq-parts/usage.rst
```diff
@@ -87,6 +87,17 @@ block) from A, and then as A gets rescanned remove the files from A.
 A workaround would be to copy first from A to B, rescan B, wait for B to
 rebuild on remote ends, and then delete from A.
 
+Can I help initial sync by copying files manually?
+--------------------------------------------------
+
+If you have a large folder that you want to keep in sync over a not-so-fast network, and you have the possibility to move all files to the remote site in a faster manner, here is a procedure to follow to make sure that you don't confuse Syncthing.
```
imsodin, line 93: I don't see why Syncthing should get confused - this is more about the user getting confused :)  
I'd just remove that bit:

```suggestion
If you have a large folder that you want to keep in sync over a not-so-fast network, and you have the possibility to move all files to the remote site in a faster manner, here is a procedure to follow:
```

roseeng:
There are a couple of discussions on the forum now that might have benefited from having this in the docs... is there anything you want me to add/change?

users/faq-parts/usage.rst
```diff
@@ -87,6 +87,17 @@ block) from A, and then as A gets rescanned remove the files from A.
 A workaround would be to copy first from A to B, rescan B, wait for B to
 rebuild on remote ends, and then delete from A.
 
+Can I help initial sync by copying files manually?
+--------------------------------------------------
+
+If you have a large folder that you want to keep in sync over a not-so-fast network, and you have the possibility to move all files to the remote site in a faster manner, here is a procedure to follow:
+
+- Create the folder on site 1, but don't share it with anyone.
```
tomasz1986, line 95:
```suggestion
- Create the folder on device A, but don't share it with device B yet.
```

users/faq-parts/usage.rst
```diff
@@ -87,6 +87,17 @@ block) from A, and then as A gets rescanned remove the files from A.
...
+- Copy the files from site 1 to site 2 using normal file copy. If this takes a long time (perhaps travelling to site 2), it could be a good idea to make sure that the files on site 1 is not updated while you do this.
```
tomasz1986, line 96:
```suggestion
- Copy the files from device A to device B using regular file copy. If this takes a long time (perhaps requiring travelling to device B physically), it may be a good idea to make sure that the files on device A are not updated while you are doing this.
```

users/faq-parts/usage.rst
```diff
@@ -87,6 +87,17 @@ block) from A, and then as A gets rescanned remove the files from A.
...
+- Create the folder on site 2, and this time copy the Folder ID from the folder on site 1. Normally you would always use the value that Syncthing supplies to you - since it must be unique - but in this case we want the folders to be considered the same.
```
tomasz1986, line 97:
```suggestion
- Create the folder on device B, and this time copy the folder ID from the folder on device A. Normally, you would always use the value that Syncthing supplies to you – since it must be unique – but in this case we want the folders to be considered the same.
```

users/faq-parts/usage.rst
```diff
@@ -87,6 +87,17 @@ block) from A, and then as A gets rescanned remove the files from A.
...
+
+Now share the folder with the other device, on both sides. Syncthing will exchange file information but existing files (or parts) will not be transferred.
```
tomasz1986, line 99:
```suggestion
Now share the folder with the other device, on both sides. Syncthing will exchange file information, but existing files will not be transferred.
```

users/faq-parts/usage.rst
tomasz1986: I don't think we use "sites" to refer to "devices" anywhere, so I'd suggest to stick to "device" instead.

users/faq-parts/usage.rst
```diff
@@ -87,6 +87,17 @@ block) from A, and then as A gets rescanned remove the files from A.
 A workaround would be to copy first from A to B, rescan B, wait for B to
 rebuild on remote ends, and then delete from A.
 
+Can I help initial sync by copying files manually?
+--------------------------------------------------
+
+If you have a large folder that you want to keep in sync over a not-so-fast network, and you have the possibility to move all files to the remote site in a faster manner, here is a procedure to follow:
```
tomasz1986, line 93:
```suggestion
If you have a large folder that you want to keep in sync over a not-so-fast network, and you have the possibility to move all files to the remote device in a faster manner, here is a procedure to follow:
```

I'm sorry, I missed this one instance. Otherwise, everything looks fine for me.

PR #650, opened 2021-06-05
users/untrusted: Info about decrypting data

imsodin:
Adds a notice that the folder id is also part of the encryption key and a section about decrypting data.

### DIFF
--- a/users/untrusted.rst
+++ b/users/untrusted.rst
@@ -146,11 +146,11 @@ the changed block is transferred as usual.
 Decrypting data
 ---------------
 
-Say you recovered a backup of an untrusted device. To get at the plaintext data,
-you can either set up Syncthing to share the data as a receive-only folder to
-a trusted device as explained above. It's also possibly to directly decrypt the
-files without involving sharing, running the following command pointed at the
-root of the encrypted folder:
+Say you recovered a backup or lost the Syncthing database on an untrusted
+device, and need to get the plaintext data. To do that there's a command to
+convert the encrypted data to plaintext on disk. Meaning you should only do tath
+on a trusted device. To decrypt the data, run the following command pointed at
+the root of the encrypted folder:
 
 .. code-block:: bash
 
--- a/users/untrusted.rst
+++ b/users/untrusted.rst
@@ -8,6 +8,13 @@ It is possible to set a password on a folder when it's shared with another
 device. Data sent will be encrypted by this password, and data received will
 be decrypted by the same password.
 
+.. note::
+
+    The data is encrypted using both the password and the folder ID. You
+    shouldn't ever need the folder ID as it's stored in the folder marker,
+    however to be on the safe side make sure you store both securily and
+    reliably.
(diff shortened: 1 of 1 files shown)


users/untrusted.rst
```diff
@@ -136,6 +143,19 @@ and receive the data for the new file over the network. However if you have a bi
 e.g. ``video.mp4``, and you modify just a part of it (e.g. video metadata), only
 the changed block is transferred as usual.
 
+Decrypting data
+---------------
+
+Say you recovered a backup of an untrusted device. To get at the plaintext data,
+you can either set up Syncthing to share the data as a receive-only folder to
+a trusted device as explained above. It's also possibly to directly decrypt the
```
calmh, line 151: This part isn't implemented yet as the scanner doesn't import the encrypted fileinfos. Currently offline decryption is the only option if the database is gone.

users/untrusted.rst
```diff
@@ -8,6 +8,13 @@ It is possible to set a password on a folder when it's shared with another
 device. Data sent will be encrypted by this password, and data received will
 be decrypted by the same password.
 
+.. note::
+
+    The data is encrypted using both the password and the folder ID. You
+    shouldn't ever need the folder ID as it's stored in the folder marker,
+    however to be on the safe side make sure you store both securily and
```
calmh, line 15:
```suggestion
    however to be on the safe side make sure you store both securely and
```

users/untrusted.rst
```diff
@@ -146,11 +146,11 @@ the changed block is transferred as usual.
...
 ---------------
 
-Say you recovered a backup of an untrusted device. To get at the plaintext data,
-you can either set up Syncthing to share the data as a receive-only folder to
-a trusted device as explained above. It's also possibly to directly decrypt the
-files without involving sharing, running the following command pointed at the
-root of the encrypted folder:
+Say you recovered a backup or lost the Syncthing database on an untrusted
+device, and need to get the plaintext data. To do that there's a command to
+convert the encrypted data to plaintext on disk. Meaning you should only do tath
```
calmh, line 151:
```suggestion
convert the encrypted data to plaintext on disk. Meaning you should only do that
```

users/untrusted.rst
```diff
@@ -136,6 +143,19 @@ and receive the data for the new file over the network. However if you have a bi
...
+
+Say you recovered a backup or lost the Syncthing database on an untrusted
+device, and need to get the plaintext data. To do that there's a command to
+convert the encrypted data to plaintext on disk. Meaning you should only do tath
+on a trusted device. To decrypt the data, run the following command pointed at
+the root of the encrypted folder:
+
+.. code-block:: bash
+
+    $ syncthing decrypt <path>
```
calmh, line 157:
```suggestion
    $ syncthing decrypt --to <plaintext-destination> --password <your-password> <encrypted-folder-path>
```

calmh (approved):

PR #653, opened 2021-07-08
Adding shout outs to Restic and Rclone

iamh2o:
as backup options.


imsodin:
This isn't about backup and/or cloud storage though, but direct, local, filesystem based copy/sync. Thus I feel those examples would be misleading (Syncthing doesn't directly work with cloud storage and isn't a backup solution) - nothing against the projects.

iamh2o:
valid-

PR #654, opened 2021-07-16
users/faq-parts: Replace fsync with Möbius sync

rasa:
The fsync() app is no longer on Apple's App Store, and hasn't been updated in years.
Möbius Sync is on the App Store, and appears to be actively developed.
For details, see the FAQ item titled "Is Möbius Sync open source?' at 
https://www.mobiussync.com/faq/

### DIFF
--- a/users/contrib.rst
+++ b/users/contrib.rst
@@ -67,9 +67,8 @@ iOS
 
 - `Möbius Sync <https://www.mobiussync.com/>`_
 
-  Möbius Sync is an alternative implementation of Syncthing (using the same network
-  protocol) and uses a somewhat outdated fork of Syncthing's code at its core. The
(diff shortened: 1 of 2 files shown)


calmh:
It's not an alternative implementation though; fsync was, but Möbius isn't. It's a skin/wrapper like the syncthing-android app.

rasa:
@calmh I updated the docs per your comment.

tomasz1986:
Is there actually a consensus regarding recommending a closed source application in the official documentation? I personally think that it is better than nothing, but many may have much stronger opinions on this matter.

As a matter of fact, all the other solutions that are recommended currently seem to be free and open source.

calmh:
Taking the above into account, my current thinking is indeed "no".

PR #660, opened 2021-08-08
users/faq-parts: Windows allows [ in filenames

rasa:
See https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions


calmh (commented):
Frankly I'd remove the entire sentence. It seems to say that escaping isn't necessary because the characters in question aren't allowed in Windows file names, but that's not really the case.

rasa:
@calmh I removed the sentence.

calmh (approved):

PR #664, opened 2021-08-15
usage: Add title

rasa:
It's the only page where the link to it isn't in the page's title.


calmh:
I think this is somewhat intentional, as this page is also what's used as `man syncthing`

PR #665, opened 2021-08-15
usage: minor tweaks and opinionated changes

rasa:


rasa:
This statement
```
A workaround would be to copy first from A to B, rescan B, wait for B to
 rebuild on remote ends, and then delete from A.
 ```
Is not clear to me. It seems it should be 
`to rebuild the files on its end`
or
`for its rebuilding to end`
But I'm not sure of the author's intent here.

calmh:
> to rebuild the files on its end

this, i.e., to copy the files on the remote side

calmh (approved):

rasa:
@calmh Fixed, per your suggestion

calmh (approved):

PR #668, opened 2021-08-16
users/config: consistency and readability tweaks

rasa:


rasa:
Re zeros vs zeroes, see https://english.stackexchange.com/questions/3824/what-is-the-plural-form-of-zero

users/config.rst
```diff
@@ -338,61 +338,61 @@ order
...
+        Enabling this is highly discouraged - use at your own risk. You have been warned.
 
     When set to true, this device will pretend not to see instructions to
     delete files from other devices.
 
 scanProgressIntervalS
-    The interval with which scan progress information is sent to the GUI. Zero
-    means the default value (two seconds).
+    The interval in seconds with which scan progress information is sent to the GUI. Setting to ``0``
+    will force the system to use the default value of two.
```
tomasz1986, line 349: This sounds rather lengthy. Is the "force the system" part really necessary? How about simply "will use the default value of two"?

rasa: I agree, it sounds wordy, and "force" is heavy-handed. But just saying `0` will use `2` sounds non-sensical. It's best to spell things out in this case, as not everyone has English as their primary language.

I changed "force" to "cause". Is that better? If not, we'll simplify as you suggest.

tomasz1986: > as not everyone has English as their primary language.

Personally, I'm not sure about the "system" part. Do you mean the operating system or Syncthing? Less technical users (regardless of their native tongue) will likely have no idea what "system" means here.

rasa: Agreed. I changed it to Syncthing to be clear.

calmh (approved):

PR #671, opened 2021-08-16
folder-uselargeblocks: Grammar tweaks

rasa:


advanced/folder-uselargeblocks.rst
```diff
@@ -19,9 +19,9 @@ useLargeBlocks
 of blocks for files larger than 256 MiB. When enabled, the file will be
 indexed and transferred using blocks larger than the standard 128 KiB. This
 results in a smaller block list and thus lower overhead. The larger block
-sizes are powers of two from from 128 KiB up to 16 MiB. Syncthing will
+sizes are powers of two from 128 KiB up to 16 MiB. Syncthing will
 attempt to select a block size to keep the number of blocks in a file
-between 1000 and 2000, using the largest and smallest block size accordingly
+between 1,000 and 2,000, using the largest and smallest block size accordingly
```
imsodin, line 24: No need for thousands separators here in my opinion. Those are customary only for >=5 digits and I don't like using either dots or commas in any case. Both are also used as decimal points in some locales. A (thin) space is the international standard.

imsodin (changes requested):
Thanks! Once comment about the thousand separators.

imsodin (approved):

PR #673, opened 2021-08-16
device-allowednetworks: Switch to devices plural

rasa:


advanced/device-allowednetworks.rst
```diff
@@ -5,19 +5,19 @@ allowedNetworks
 
 ``allowedNetworks`` is an advanced device setting that affects connection
 handling. The default is for this setting to be empty, indicating that there
-is no restriction on the allowed networks for a device.
+are no restrictions on the allowed networks for a device.
 
 By setting this to a comma separated list of networks, connections to the
 given device will be limited to those networks. The networks refer to the
-address of the *remote* device, not the network that the local device is
+address range of *remote* devices, not the network that the local device is
```
imsodin, line 12: `allowedNetworks` is an option specific to one single device, same twice below.

imsodin (changes requested):

advanced/device-allowednetworks.rst
rasa: I know, plurals of plurals is confusing.

I tried to improve the wording, but now it's even more verbose. Perhaps we just fix the is/are, and leave the rest alone, as it's probably fine as is.

imsodin: I think the current version shows there's still a misconception in place here: There's one `allowedNetworks` for each remote device. It does not limit connections to remote devices with those addresses, it restricts connections to the particular remote device that is being configured to the given addresses. Or maybe I am completely misunderstanding the text.

rasa: Gotcha. I completely misunderstood the setting. Should of read the source. I'll close this and move on, as it's probably fine as is.

PR #676, opened 2021-08-18
ldap: Improve grammar

rasa:


users/ldap.rst
```diff
@@ -6,11 +6,11 @@ source. Typical examples include Microsoft Active Directory and OpenLDAP /
 OpenDirectory servers.
 
 To enable LDAP authentication some configuration options must be set in the
-advanced config editor. Under Actions, Advanced expand the LDAP tab. The
-following options are available:
+advanced config editor. Under Actions, selecting Advanced displays the LDAP tab. 
+The following options are available:
```
tomasz1986, line 11: What's wrong with "expand" here? We are literally expanding the tab (which is displayed shrunken by default).

PS It is "grammar" 😉.

rasa: Selecting Advanced displays (or exposes, or makes visible) the LDAP tab. It doesn't really expand it, as the tab isn't visible at all, until you click Advanced.

I *always* spell grammar wrong. Go figure.

tomasz1986: As far as I understand the original instructions, the point is that you have to click the tab in order to display all the options, hence "(you) expand the LDAP tab".

rasa: Understood. I changed it back to expands. I think it still needs "selecting", though.

tomasz1986: I think that there is a little misunderstanding here though. The original sentence uses an imperative form of "expand". It doesn't say that advanced "expands" the LDAP tab, but rather that "(you) expand the LDAP tab". I guess that the comma causes some confusion here, but how about

```
Under "Actions > Advanced", expand the LDAP tab.
```
?

rasa: That's perfect.

PR #679, opened 2021-09-09
rest: Dont use default as folder ID example

jtagcat:
Folder ID-s usually look more like `ab1c2-def3g`. Changing to `default` to `ab1c2-def3g` makes more sense.

Correct me if I'm wrong.


imsodin (approved):

tomasz1986:
For the record, the Default folder's ID is in fact `default` 😉.

PR #689, opened 2021-10-30
build: Update deps for netlify

imsodin:


imsodin:
@calmh Looks like we are on an ancient netlify build image, and I don't have access to change that. Could you bump that to something recent when you get the chance (to move to a newer python at some point before 3.7 is EOL).

calmh:
@imsodin yeah I wasn't aware it was locked to something, but indeed that was the case. Update from Ubuntu 16.04 to Ubuntu 20.04.

calmh:
Deploy now fails, looks like we need Python 3.8 instead of 3.7

PR #691, opened 2021-10-30
build: Use default python from netlify

imsodin:


calmh:
Default is apparently 2.7 👯 

I just merged a PR with 3.8 instead

imsodin:
Default to an EOL python version - nice one netlify o.O

calmh:
Well, that is what you get with `apt install python; python --version` in Ubuntu 20.04, so I guess this one is on Ubuntu (or Debian)

wweich:
Ubuntu 20.04 "ships" with Python3. python is (by default) still the alias for Python2, so you can't run `python` on a fresh Ubuntu 20.04, but only `python3`. AFAIK this is because most python software developed for python3 uses python3 as env and most software which uses python as env still wants (and can only run with) python2.

calmh:
Apparently so.

```
root@ff46f6c186e6:/# apt install python
...
Note, selecting 'python-is-python2' instead of 'python'
```

(the Docker image ships with neither Python installed though)

PR #692, opened 2021-11-03
Document missing CLI option flags and use double-dash consistently.

acolomb:
The command-line help output lists all options formatted with `--double-dashes`.  Although Syncthing actually accepts both single and double dash forms, I think keeping consistency between CLI help and docs helps avoid confusion.

The same changes could probably be done for `stdiscosrv` and `strelaysrv`, as well as the `build.go` script documentation.  They seem to accept double dashes as well, but print the single dash in their help output.

Newly documented options:
 * `--log-max-files`
 * `--log-max-size`
 * `--allow-newer-config`
 * `--no-upgrade`
 * `--gui-apikey`

One important fix is in the `syncthing` command synopsis, where `[device-id]` was looking like an optional argument to be specified, because of the missing dash(es).

### DIFF
--- a/users/syncthing.rst
+++ b/users/syncthing.rst
@@ -13,11 +13,11 @@ Synopsis
               [--generate=<dir>] [--gui-address=<address>] [--gui-apikey=<key>]
               [--home=<dir> | --config=<dir> --data=<dir>]
               [--logfile=<filename>] [--logflags=<flags>]
-	      [--log-max-files=<num>] [--log-max-size=<num>]
+              [--log-max-files=<num>] [--log-max-size=<num>]
               [--no-browser] [--no-console] [--no-restart] [--paths] [--paused]
               [--reset-database] [--reset-deltas] [--unpaused] [--allow-newer-config]
-	      [--upgrade] [--no-upgrade] [--upgrade-check] [--upgrade-to=<url>]
-	      [--verbose] [--version]
+              [--upgrade] [--no-upgrade] [--upgrade-check] [--upgrade-to=<url>]
+              [--verbose] [--version]
 
 Description
 -----------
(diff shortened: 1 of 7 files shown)


users/syncthing.rst
```diff
@@ -9,13 +9,15 @@ Synopsis
...
-              [-upgrade-check] [-upgrade-to=<url>] [-verbose] [-version]
+    syncthing [--audit] [--auditfile=<file|-|-->] [--browser-only] [--device-id]
+              [--generate=<dir>] [--gui-address=<address>] [--gui-apikey=<key>]
+              [--home=<dir> | --config=<dir> --data=<dir>]
+              [--logfile=<filename>] [--logflags=<flags>]
+              [--log-max-files=<num>] [--log-max-size=<num>]
+              [--no-browser] [--no-console] [--no-restart] [--paths] [--paused]
+              [--reset-database] [--reset-deltas] [--unpaused] [--allow-newer-config]
+              [--upgrade] [--no-upgrade] [--upgrade-check] [--upgrade-to=<url>]
+              [--verbose] [--version]
```
imsodin, line 20: This is still legacy usage, the new way would be using `syncthing cli`, and display options/help generated from that cli here. Might be possible to adapt `syncthing --help` output (and `syncthing serve --help`), possibly even auto-generating it. And having docs for the other sub-commands, but that's clearly out-of-scope here :)

imsodin (approved):
Thanks.

users/syncthing.rst
acolomb: Actually I'm working on at least a manual take at documenting the other subcommands right now.  That's when I found some inconsistencies, see https://github.com/syncthing/syncthing/pull/8037.

Auto-generating stuff would be neat, and AFAICT the urfave/cli package supports that approach and stuff like bash-completions as well.  But as long as the repositories are separated, that automation step will be really hard I suppose.  So focusing on fixups right now.

My main motivation and starting point btw. was having any mention of `syncthing cli` at all in the docs, pointing to its self-explanatory help system at the very least.

acolomb: Why do you say this is "legacy" usage?  It's still 100 % valid for the `serve` command?

acolomb: Ping @imsodin ?

PR #693, opened 2021-11-03
users/syncthing: Explain the different CLI modes based on sub-commands.

acolomb:
Add a section about the general concept and include "decrypt" and "cli" in the syncthing command synopsis.  Top-level entry points into the "cli" hierarchy are documented in a separate included snippet.

Hopefully something like that can be auto-generated in the distant future to make sure an up-to-date listing is included in the documentation.

Add options used exclusively in the `decrpyt` mode to the overall options list.

### DIFF
--- a/includes/cli-commands.rst
+++ b/includes/cli-commands.rst
@@ -1,7 +1,7 @@
-The available sub-commands are grouped into several nested hierarchies and
-dynamically generated from the running Syncthing instance.  On every level, the
-``--help`` option lists the available properties, actions and commands for the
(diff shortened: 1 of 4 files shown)


acolomb:
I wonder what will happen to the reference into `untrusted.rst` when this gets converted to a manpage?

acolomb:
Superseded by #694.

acolomb:
Actually since syncthing/syncthing#8043 is not merged for the current release candidate cycle, this could still be included without waiting for #694.

PR #694, opened 2021-11-16
users/syncthing: Document the generate subcommand and its options

acolomb:
This is based on -- and includes the changes from -- #693.

Concerns the new subcommand added in https://github.com/syncthing/syncthing/pull/8043.

### DIFF
--- a/users/syncthing.rst
+++ b/users/syncthing.rst
@@ -21,6 +21,12 @@ Synopsis
               [--upgrade] [--no-upgrade] [--upgrade-check] [--upgrade-to=<url>]
               [--verbose] [--version] [--help] [--debug-*]
 
+    syncthing generate
+              [--home=<dir> | --config=<dir>]
+              [--gui-user=<username>] [--gui-password=<password|->]
+              [--no-default-folder] [--no-console]
+              [--help]
+
     syncthing decrypt (--to=<dir> | --verify-only)
               [--password=<pw>] [--folder-id=<id>] [--token-path=<file>]
               [--continue] [--verbose] [--version] [--help]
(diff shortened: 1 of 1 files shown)


users/syncthing.rst
```diff
@@ -202,6 +268,29 @@ Exit codes over 125 are usually returned by the shell/binary loader/default
 signal handler. Exit codes over 128+N on Unix usually represent the signal which
 caused the process to exit. For example, ``128 + 9 (SIGKILL) = 137``.
 
+Sub-Commands
+------------
+
+The command line syntax actually supports different modes of operation through
+several sub-commands, specified as the first argument.  The default one is
+``serve`` and can therefore be omitted.
```
imsodin, line 276: That's just for backwards compatibility, I don't think it should be documented and thus it's use encoured.

imsodin (commented):
Great, thanks a lot for writing this up!

A (possible) nit: I think the common way is to write it as a single word "subcommand", not "sub-command". I am not a native English speaker, so not an authority on this.

users/syncthing.rst
acolomb: There are tons of examples out there in the Internets which don't include `serve`, where the documentation would not explain why they work at all.

```suggestion
several sub-commands, specified as the first argument.  The default one is
``serve``, assumed when none is given.
```

No longer a suggestion to do so, but still explains what will happen.  Will commit a better wording...

acolomb:
> Great, thanks a lot for writing this up!
> 
> A (possible) nit: I think the common way is to write it as a single word "subcommand", not "sub-command". I am not a native English speaker, so not an authority on this.

About three times as many Google hits on "subcommand", so that seems preferred.

imsodin (approved):
Looks good to me. If someone with good English-Fu would go over this, that would be nice - us German-speakers have a nack for making English more complicated than necessary :D

imsodin (changes requested):
Actually this shouldn't be merged until the generate command is released.

acolomb:
> Actually this shouldn't be merged until the generate command is released.

It did end up in [v1.18.5-rc.2](https://github.com/syncthing/syncthing/releases/tag/v1.18.5-rc.2), maybe by accident?  I don't think it's very intrusive so having a shorter RC cycle for this feature is probably ok.  And it was my fault for not looking at the RC timing before merging the branch.  Anyway, just a heads up that this docs PR should be merged unless you think https://github.com/syncthing/syncthing/pull/8043 should be pulled back.

acolomb:
Hm. v1.18.5 was released and tagged here on the docs repo. Okay to merge @imsodin, so it ends up in the next release tag at least?

imsodin:
Yeah right, forgot it made it into the RC. I don't think we do docs tagging, or do we? So it should be available in a few moments.

acolomb:
The tags are created (automatically?) when the main repo has a release.  Just that the web site deployment doesn't care about it at all and just uses `main` I suppose.

PR #695, opened 2021-11-17
config.rst: Improve wording

camoz:


camoz:
Please review carefully!

From my understanding exactly one address must be present. If that's not the case and there can be multiple, then change the wording to `At least one ...`. Anyhow, I think the wording should be made more explicit.

calmh (approved):

PR #697, opened 2021-11-18
getting-started.rst: Update installation instructions

camoz:
On common linux distros like Debian/Ubuntu/Fedora/Archlinux/etc., it is best practice to install and maintain software packages via the local package manager. Reflect this in the installation guide.

### DIFF
--- a/intro/getting-started.rst
+++ b/intro/getting-started.rst
@@ -40,7 +40,9 @@ Syncthing
 ~~~~~~~~~
 
 Grab the `latest release`_ of Syncthing for your operating system and unpack
-it (if you use Debian/Ubuntu you may want to use the `Debian repository`_). 
+it. If you are running Linux, it is best to install Syncthing via the local
+package manager of your Linux distribution. If you use Debian/Ubuntu and want a
+more up-to-date version of Syncthing, you can use the `Debian repository`_.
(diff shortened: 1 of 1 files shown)


calmh:
Let's not make a judgement on what's "best" here. We mention the alternatives.

camoz:
Sorry for the late response. How about if we change the docs from 

>  If you are running Linux, it is best to install Syncthing

to

>  If you are running Linux, you can install Syncthing

? Then it's completely neutral. Right now it's not neutral, since installation via the local package manager is not mentioned anywhere, despite that being the standard way to install software on linux distributions.

Some reference on why maintaining local software via the local package manager is generally the preferred way to do it:

- https://drewdevault.com/2019/12/09/Developers-shouldnt-distribute.html
- https://drewdevault.com/2021/09/27/Let-distros-do-their-job.html

Imagine what would happen if more and more software was distributed not via the central package management system, but via app-specific update mechanisms. E.g. software breakage could happen at any point, when an app decides to update quietly in the background. Instead, it's safer to schedule system updates, when one has set aside a little bit of time to deal with potential breakage. There are some more good points in the blog posts above.

AudriusButkevicius:
We do not recommend using the distro installations, as they are often lagging behind quite badly causing all sorts of pain.

PR #700, opened 2021-11-24
Update FAQ on iOS client

jakob:
While reading the FAQ I stumbled across an outdated section.

- I removed mention of fsync() which seems to be abandoned
- added mention of Mobius Sync, which is currently available on the iOS App Store

I'm not sure if you want to mention 3rd party software in your docs, but considering that Resilio Sync gets a mention maybe Mobius Sync (which actually works with Syncthing) should also get a mention?

### DIFF
--- a/users/faq-parts/general.rst
+++ b/users/faq-parts/general.rst
@@ -79,13 +79,17 @@ Sync uses an undocumented, closed protocol with unknown security properties.
 
 .. [#resiliosync] https://en.wikipedia.org/wiki/Resilio_Sync
 
-Why is there no iOS client?
----------------------------
+Is there an iOS client?
+-----------------------
 
-There is an alternative implementation of Syncthing (using the same network
-protocol) called ``fsync()``. There are no plans by the current Syncthing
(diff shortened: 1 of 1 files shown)


calmh (approved):

calmh:
Thanks. I made a slight wording change to my preference but the substance is good 👍

jakob:
You are good with words! I was looking for a way to describe Mobius without calling it a hack, but couldn't come up with anything.

rasa:
I'm [thankful](https://github.com/syncthing/docs/pull/654#issuecomment-900273792).

tomasz1986:
> I'm [thankful](https://github.com/syncthing/docs/pull/654#issuecomment-900273792).

Yeah. I think it would at least be beneficial to underline that the program is closed source (as "commercial" doesn't necessary equal being so).

rasa:
@tomasz1986 And your comment is helpful too:
> Möbius Sync also requires the user to pay if they want to use more than 20 MB of storage

PR #704, opened 2021-12-22
add systemd relay methods, outline deb package locations

tepozoa:


imsodin (commented):
Thanks for your contribution to the docs!

I think the heading structure isn't quire right and requires also some structuring changes to the existing content. All the new sections are sub-sections of the "Setting Up" section. You could e.g. remove the "Debian/Ubuntu Package" title and add the explanation about the user setup etc. that happened automatically directly under "Setting Up", telling the reader that they can skip the instructions to create a user in that case. Also the systemd instructions could just directly be appended to the paragraph that says "use the recommended procedure", as a (/the prevalent) example option. 

For the "443 with unprivileged" instructions, I'd add a new header for the top parts, maybe called "Port forwarding", thus making it clear that that's one option and systemd anther.

imsodin:
I hope the closing wasn't a reaction to my feedback. I really like your contribution (very well understandable and to the point). My feedback just aimed at integrating it a bit more into the existing structure. You may integrate them, or you of course may also disagree ;)

PR #705, opened 2022-01-02
users/contrib: Updated Syncthing for Home Assistant OS

jaegerschnitzel:
changed Syncthing for Home Assistant OS to better maintained addon.


imsodin (approved):
The removed repo has a notice that it's owner isn't active anymore, maintenance options at best -> lgtm.

PR #708, opened 2022-01-09
rest: Remove bogus fields from connections API endpoint.

acolomb:
Remove fields from the /rest/system/connections element "total" which
are not populated and misleading.  Removed in upcoming version.


calmh (approved):

acolomb:
So the version info should be correct with v1.19.0?  It is technically a change in API, that was my assumption at least.

PR #709, opened 2022-01-09
rest, users: Update configuration to version 35.

acolomb:
Concerns the following fields changed in recent versions:

* encryptionPassword (added in folder)
* cleanupInterval, fsPath, fsType (added in folder versioning)
* caseSensitiveFS, junctionsAsDirs (added in folder)
* ignoredFolder (moved from configuration to device element)
* untrusted (added under device)
* unackedNotificationID (previously not in example)
* defaultFolderPath (moved from options to defaults)
* announceLANAddresses (added in options)
* sendFullIndexOnUpgrade (added in options)
* connectionLimitEnough (added in options)
* connectionLimitMax (added in options)
* insecureAllowOldTLSVersions (added in options)
* remoteIgnoredDevice (renamed from ignoredDevice under configuration)

Added the whole defaults section with an appropriate short
description.

Not all added options in the example have a description yet.

Change the GET /rest/system/config response to mostly match the
example XML listed in the config.xml description (except for skipping
device IDs).

### DIFF
--- a/rest/system-config-get.rst
+++ b/rest/system-config-get.rst
@@ -10,22 +10,23 @@ Returns the current configuration.
 .. code-block:: json
 
     {
-      "version": 30,
+      "version": 35,
       "folders": [
 	{
-	  "id": "GXWxf-3zgnU",
-	  "label": "MyFolder",
+	  "id": "default",
+	  "label": "Default Folder",
 	  "filesystemType": "basic",
 	  "path": "...",
 	  "type": "sendreceive",
 	  "devices": [
 	    {
 	      "deviceID": "...",
-	      "introducedBy": ""
+	      "introducedBy": "",
+	      "encryptionPassword": ""
 	    }
 	  ],
-	  "rescanIntervalS": 60,
-	  "fsWatcherEnabled": false,
+	  "rescanIntervalS": 3600,
+	  "fsWatcherEnabled": true,
 	  "fsWatcherDelayS": 10,
 	  "ignorePerms": false,
 	  "autoNormalize": true,
@@ -34,10 +35,11 @@ Returns the current configuration.
 	    "unit": "%"
 	  },
 	  "versioning": {
-	    "type": "simple",
-	    "params": {
-	      "keep": "5"
-	    }
+	    "type": "",
+	    "params": {},
+	    "cleanupIntervalS": 3600,
+	    "fsPath": "",
+	    "fsType": "basic"
 	  },
 	  "copiers": 0,
 	  "pullerMaxPendingKiB": 0,
@@ -46,14 +48,20 @@ Returns the current configuration.
 	  "ignoreDelete": false,
 	  "scanProgressIntervalS": 0,
 	  "pullerPauseS": 0,
-	  "maxConflicts": 10,
+	  "maxConflicts": -1,
 	  "disableSparseFiles": false,
 	  "disableTempIndexes": false,
 	  "paused": false,
 	  "weakHashThresholdPct": 25,
 	  "markerName": ".stfolder",
 	  "copyOwnershipFromParent": false,
-	  "modTimeWindowS": 0
+	  "modTimeWindowS": 0,
+	  "maxConcurrentWrites": 2,
+	  "disableFsync": false,
+	  "blockPullOrder": "standard",
+	  "copyRangeMethod": "standard",
+	  "caseSensitiveFS": false,
+	  "junctionsAsDirs": true
 	}
       ],
       "devices": [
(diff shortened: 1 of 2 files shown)


users/config.rst
```diff
@@ -228,6 +290,8 @@ Folder Element
         <disableFsync>false</disableFsync>
         <blockPullOrder>standard</blockPullOrder>
         <copyRangeMethod>standard</copyRangeMethod>
+        <caseSensitiveFS>false</caseSensitiveFS>
+        <junctionsAsDirs>true</junctionsAsDirs>
```
tomasz1986, line 294:
```suggestion
        <junctionsAsDirs>false</junctionsAsDirs>
```

Sticking to the convention of using default values here, shouldn't this say `false`? The default for `junctionsAsDirs` is `false`.

users/config.rst
```diff
@@ -206,9 +262,15 @@ Folder Element
 
     <folder id="default" label="Default Folder" path="/Users/jb/Sync/" type="sendreceive" rescanIntervalS="3600" fsWatcherEnabled="true" fsWatcherDelayS="10" ignorePerms="false" autoNormalize="true">
         <filesystemType>basic</filesystemType>
-        <device id="3LT2GA5-CQI4XJM-WTZ264P-MLOGMHL-MCRLDNT-MZV4RD3-KA745CL-OGAERQZ"></device>
+        <device id="S7UKX27-GI7ZTXS-GC6RKUA-7AJGZ44-C6NAYEB-HSKTJQK-KJHU2NO-CWV7EQW" introducedBy="">
```
tomasz1986, line 265: Is there any particular reason for changing the existing device IDs (and API keys too, see below)?

users/config.rst
```diff
@@ -907,6 +980,79 @@ setLowPriority
...
+    attribute is meaningless in this context.
+
+folder
+    Template for a ``folder`` element, with the same internal structure.  Any
+    fields here will be used for a newly added shared folder.  The ``id``
+    attribute is meaningless in this context.
+
+    The UI will propose to create new folders at the path given by the ``path``
+    attribute (used to be ``defaultFolderPath`` under ``options``).  It also
+    applies for automatically accepted folders from a remote device.
```
tomasz1986, line 1051:
```suggestion
    The UI will propose to create new folders at the path given in the ``path``
    attribute (used to be ``defaultFolderPath`` under ``options``).  It also
    applies to folders automatically accepted from a remote device.
```

users/config.rst
acolomb: Done.  I have no clue how it ended up there, because I just migrated the existing config.xml content. from the example.

users/config.rst
acolomb: Yes.  I don't have access to the device certificate originally used to generate the examples.  So it generated a new certificate locally and that one's device ID always got added as a "new" (local) device everywhere.  To avoid fussing around each time I regenerate the XML, I just decided to switch over to the new ID.

Nobody will care I suppose, it's just an example for the encoded format.

users/config.rst
acolomb: Applied, thanks.

rest/system-config-get.rst
```diff
@@ -153,7 +173,94 @@ Returns the current configuration.
...
+	"connectionLimitEnough": 0,
+	"connectionLimitMax": 0,
+	"insecureAllowOldTLSVersions": false
       },
-      "remoteIgnoredDevices": []
+      "remoteIgnoredDevices": [
+	{
+	  "time": "2022-01-09T20:02:01Z",
+	  "deviceID": "...",
+	  "name": "bugger",
```
calmh, line 188:
```suggestion
	  "name": "...",
```

rest/system-config-get.rst
```diff
@@ -114,14 +133,14 @@ Returns the current configuration.
...
-	"startBrowser": false,
+	"startBrowser": true,
 	"natEnabled": true,
 	"natLeaseMinutes": 60,
 	"natRenewalMinutes": 30,
 	"natTimeoutSeconds": 10,
-	"urAccepted": -1,
-	"urSeen": 2,
-	"urUniqueId": "",
+	"urAccepted": 3,
```
calmh, line 141:
```suggestion
	"urAccepted": 0,
```

acolomb: Applied.

rest/system-config-get.rst
acolomb: Applied.

acolomb:
Note I have started to add some more missing config element docs, even if just with a TODO admonition.  But that work is based on a rather big branch of mine which adds proper link and cross-reference targets to each config option.  So stay tuned for that and please be patient about other missing entries ;-)

calmh:
LGTM. About the changing device IDs, if we want to use something stable moving forward we could use the test setup in https://github.com/syncthing/syncthing/tree/main/test where there are fixed public certificates / device IDs.

acolomb:
Good to know there are some testing certificates available.  Will remember that for the next change, hopefully.

Issue #442, opened 2019-03-06
Configuration option "autoAcceptFolders" for devices isn't documented


dhs-rec:
Reading the configuration via REST API call (and also in the UI), I see in the output that each device has an "autoAcceptFolders" option, set to false. However, I couldn't find it in the documentation.

So what exactly happens if I set this to true?
- Where are automatically accepted folders stored?
- What are their names?

AudriusButkevicius:
They are stored at the default path, and name I think is folder name advertised by remote peer, if that is taken then its folder name plus folder id.

AudriusButkevicius:
Feel free to update the docs :)

dhs-rec:
Well, if you only "think" what it does, how should I know? Do you really want me to update the docs with some half-baked knowledge?

AudriusButkevicius:
For the directory name, it tries folder label, if that exists already on the filesystem it uses the id, if the id exists as well, it doesn't do anything and just logs that it failed.

If you already have a syncthing folder with that id, it just shares it, oppose to creating a new one.

stamepicmorg:
[from sources](https://github.com/syncthing/syncthing/blob/6db8dc33f2a346e5828a6bc92b3e70e66d853494/lib/model/model.go)
```
// handleAutoAccepts handles adding and sharing folders for devices that have
// AutoAcceptFolders set to true.
```

-----------

bump!

AudriusButkevicius:
The documentation is community maintained, so instead of bumping, I suggest you open a PR improving it.
I am more than happy to help with the wording and explanation of how it works in a PR.

stamepicmorg:
> The documentation is community maintained, so instead of bumping, I suggest you open a PR improving it.

it's all very cool. already written above - do I need to learn the "go" language to read the sources and write the documentation?
there is an undocumented thing that the developers are silent about. what to do in this case?

AudriusButkevicius:
The docs repo (the repo you are commenting on) is not go code, it's .rst which is widely documented format for documentation.
There is enough prior art of rst in this repo to deduce how to add a sentence about this setting.

What this setting does is already discussed in this ticket. A pull request adding a description does can be further iterated on, if it does not reflect reality of what the flag does.

Issue #375, opened 2018-02-14
Document new auto accept folders option


imsodin:
See syncthing/syncthing#4551

Issue #310, opened 2017-03-22
"Ignore permissions" should be documented


zertrin:
I would like to better understand how activating this option works (especially what happens when I activate it only on one device).

zertrin:
After reading the previous issue, I did as suggested a google search and found this https://forum.syncthing.net/t/v0-8-10-ignore-permissions/263

This information ought to be available in the docs IMHO.

AudriusButkevicius:
As the issue template explains, please use the [forum](https://forum.syncthing.net/) for support.
Please use the search on the forum and the closed issue tracker for simillar issues before starting a new thread. 
If this issue is later proven to be a bug, we will reopen this issue or open a new one.

calmh:
This is the docs repo, and a legitimate documentation request.

imsodin:
It is documented in https://docs.syncthing.net/users/config.html#folder-element, but from the point of view of the config file not the web UI and the explanation is minimal (e.g. not covering the specific question of the OP). Also as #302 points out the search is very limited, I wasn't able to find via the search, I just had to know where to look (even though "Configuration" -> "Folder Element" should be rather straight forward).

AudriusButkevicius:
Sorry, my bad, was pretty sure its the code repo.

bean5:
Old and documented as stated above. Can close, @calmh ?

PR #610, opened 2021-02-12
users/config: Update to v33, add missing options and ref links (fixes #310, fixes #375, fixes #442)

tomasz1986:
1. Update the config to v33, so that it matches the config.xml used in
   Syncthing v1.13.x. Make sure that all the elements listed on the page
   are exact copies of the config displayed at the top, so that it will
   easier to update them in the future. Also, remove the second instance
   of device example, as it is not really necessary for all the options
   are explained in details below anyhow.

2. Add missing options with short explanations, so that the config page
   is now complete, matching Syncthing v1.13.x.

3. Add direct links to those config options that have their own pages in
   the Docs. This way, it is possible to quickly navigate there from the
   config page. In the process, fix or add references in the advanced
   options' pages using more explicit wording, so that it is less likely
   they will overlap with other refs. Also, rename the advanced options
   RST file when necessary.

Signed-off-by: Tomasz Wilczyński <twilczynski@naver.com>

PR #710, opened 2022-01-09
users/config: Structured option documentation.

acolomb:
This is effectively a rewrite of #610 by @tomasz1986.  A completely different approach though, which uses the formal mechanisms of Sphinx through a custom plugin implementing an additional "directive" and "role" in RST parlor.  This gives us high quality permanent links to individual config options, even with highlighting on the link target anchor (option name).  All this without having to repeat oneself in the RST documents, as with the old approach where each option needed a separate label with a hopefully matching name.  And it works easily for cross-referencing any option internally from other documents.

For example, in the RST document:

```
.. stconf:option:: folder.fsWatcherDelayS

    Blah blah, only relevant if :stconf:opt:`fsWatcherDelayEnabled` is on.
```

This gives us an external link target (when deployed) https://docs.syncthing.net/users/config.html#config-option-folder.fswatcherdelays.  The internal references work as shown in the RST example.

This is based on the commits of #709 because, well I work things as they come along :-)  It also adds all (?) missing configuration options I could find in the protobuf definitions.  **Please help getting those documented where possible**, so we don't leave lots of TODOs behind after this PR!

**Reviewing is best done each commit by itself.** They include logically separate changes and have some more explanations in the commit messages.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -1210,7 +1210,7 @@ The ``options`` element contains all other global configuration options.
 
 .. option:: options.sendFullIndexOnUpgrade
 
-    Controls whether all index data is resent when an upgrade is detected,
+    Controls whether all index data is resent when an upgrade has happened,
     equivalent to starting Syncthing with :option:`--reset-deltas`.  This used
     to be the default behavior in older versions, but is mainly useful as a
     troubleshooting step and causes high database churn. The default is now
--- a/users/config.rst
+++ b/users/config.rst
@@ -344,12 +344,10 @@ element:
 
     If set to ``true``, this detects changes to files in the folder and scans them.
 
-.. _fsWatcherDelayS:
-
 .. option:: folder.fsWatcherDelayS
 
     The duration during which changes detected are accumulated, before a scan is
-    scheduled (only takes effect if ``fsWatcherEnabled`` is set to ``true``).
+    scheduled (only takes effect if :opt:`fsWatcherEnabled` is set to ``true``).
 
 .. option:: folder.ignorePerms
 
@@ -1069,7 +1067,6 @@ The ``options`` element contains all other global configuration options.
     maintain NAT mapping. Default is ``24`` and you can set it to ``0`` to
     disable contacting STUN servers.
 
-.. _set-low-priority:
 
 .. option:: options.setLowPriority
 
@@ -1154,7 +1151,7 @@ be present in the ``defaults`` element:
     applies to folders automatically accepted from a remote device.
 
     Even sharing with other remote devices can be done in the template by
-    including the appropriate ``device`` element underneath.
+    including the appropriate :opt:`folder.device` element underneath.
 
 .. _listen-addresses:
 
--- a/users/config.rst
+++ b/users/config.rst
@@ -346,6 +346,10 @@ element:
         The folder is in "receive only" mode -- it will not propagate
         changes to other devices.
 
+    ``receiveencrypted``
+        Must be used on untrusted devices, where the data cannot be decrypted
+        because no folder password was entered.  See :doc:`untrusted`.
+
 .. option:: folder.rescanIntervalS
 
     The rescan interval, in seconds. Can be set to ``0`` to disable when external
@@ -751,8 +755,9 @@ From the following child elements at least one ``address`` child must exist.
 
     This boolean value marks a particular device as untrusted, which disallows
     ever sharing any unencrypted data with it.  Every folder shared with that
-    device then needs an encryption password set.  Refer to the detailed
-    explanation under :doc:`untrusted`.
+    device then needs an encryption password set, or must already be of the
+    "receive encrypted" type locally.  Refer to the detailed explanation under
+    :doc:`untrusted`.
 
 
 GUI Element
--- a/users/config.rst
+++ b/users/config.rst
@@ -420,8 +420,9 @@ The following child elements may exist:
 
 .. option:: folder.pullerMaxPendingKiB
 
-    The number of pullers is automatically adjusted based on this desired amount
-    of outstanding request data.
+    Controls when we stop sending requests to other devices once we’ve got this
+    much unserved requests.  The number of pullers is automatically adjusted
+    based on this desired amount of outstanding request data.
 
 .. option:: folder.order
 
@@ -563,8 +564,8 @@ The following child elements may exist:
 
 .. option:: folder.junctionsAsDirs
 
-    .. todo:: Describe element!
-    .. bool                               follow_junctions           = 34 [(ext.goname) = "JunctionsAsDirs", (ext.xml) = "junctionsAsDirs", (ext.json) = "junctionsAsDirs"];
+    NTFS directory junctions can be treated as ordinary directories, if this is
+    set to ``true``.
 
 
 Device Element
@@ -1210,13 +1211,16 @@ The ``options`` element contains all other global configuration options.
 
 .. option:: options.announceLANAddresses
 
-    .. todo:: Describe element!
-    .. bool            announce_lan_addresses                   = 48 [(ext.goname)= "AnnounceLANAddresses", (ext.xml) = "announceLANAddresses", (ext.json) = "announceLANAddresses", (ext.default) = "true"];
+    Enable (the default) or disable announcing private (RFC1918) LAN IP
+    addresses to global discovery.
 
 .. option:: options.sendFullIndexOnUpgrade
 
-    .. todo:: Describe element!
-    .. bool            send_full_index_on_upgrade               = 49;
+    Controls whether all index data is resent when an upgrade is detected,
+    equivalent to starting Syncthing with :option:`--reset-deltas`.  This used
+    to be the default behavior in previous versions, but is mainly useful as a
+    troubleshooting step and causes high database churn. The default is now
+    ``false``.
 
 .. option:: options.featureFlag
 
--- a/users/config.rst
+++ b/users/config.rst
@@ -384,7 +384,7 @@ The following child elements may exist:
     Specifies a versioning configuration.
 
     .. seealso::
-	:ref:`versioning`
+	:doc:`versioning`
 
 .. option:: folder.copiers
 	    folder.pullers
@@ -591,7 +591,7 @@ element:
     should copy their list of devices per folder when connecting.
 
     .. seealso::
-	:ref:`introducer`
+	:doc:`introducer`
 
 .. option:: device.skipIntroductionRemovals
 
--- a/users/config.rst
+++ b/users/config.rst
@@ -414,7 +414,7 @@ The following child elements may exist:
 	    folder.hashers
 
     The number of copier and hasher routines to use, or ``0`` for the
-    system determined optimums. These are low level performance options for
+    system determined optimums. These are low-level performance options for
     advanced users only; do not change unless requested to or you've actually
     read and understood the code yourself. :)
 
@@ -564,8 +564,8 @@ The following child elements may exist:
 
 .. option:: folder.junctionsAsDirs
 
-    NTFS directory junctions can be treated as ordinary directories, if this is
-    set to ``true``.
+    NTFS directory junctions are treated as ordinary directories, if this is set
+    to ``true``.
 
 
 Device Element
@@ -714,8 +714,9 @@ From the following child elements at least one ``address`` child must exist.
     and synced locally under the :opt:`default path <defaults.folder>`.  For the
     folder name, Syncthing tries to use the label from the remote device, and if
     the same label already exists, it then tries to use the folder's ID.  If
-    that exists as well, it just logs an error.  A local folder already added
-    with the same ID will just be shared rather than created separately.
+    that exists as well, the folder is just offered to accept manually.  A local
+    folder already added with the same ID will just be shared rather than
+    created separately.
 
 .. option:: device.maxSendKbps
 
@@ -1211,14 +1212,14 @@ The ``options`` element contains all other global configuration options.
 
     Controls whether all index data is resent when an upgrade is detected,
     equivalent to starting Syncthing with :option:`--reset-deltas`.  This used
-    to be the default behavior in previous versions, but is mainly useful as a
+    to be the default behavior in older versions, but is mainly useful as a
     troubleshooting step and causes high database churn. The default is now
     ``false``.
 
 .. option:: options.featureFlag
 
     Feature flags are simple strings that, when added to the configuration, may
-    unleash unfished or still-in-development features to allow early user
+    unleash unfinished or still-in-development features to allow early user
     testing.  Any supported value will be separately announced with the feature,
     so that regular users do not enable it by accident.
 
@@ -1238,8 +1239,8 @@ The ``options`` element contains all other global configuration options.
 
 .. option:: options.insecureAllowOldTLSVersions
 
-    Only for compatibility with old devices, as detailed in
-    :doc:`/advanced/option-insecure-allow-old-tls-versions`.
+    Only for compatibility with old versions of Syncthing on remote devices, as
+    detailed in :doc:`/advanced/option-insecure-allow-old-tls-versions`.
 
 
 Defaults Element
--- a/users/config.rst
+++ b/users/config.rst
@@ -383,8 +383,8 @@ The following child elements may exist:
 
     Specifies a versioning configuration.
 
-.. seealso::
-    :ref:`versioning`
+    .. seealso::
+	:ref:`versioning`
 
 .. option:: folder.copiers
 	    folder.pullers
@@ -590,8 +590,8 @@ element:
     Set to true if this device should be trusted as an introducer, i.e. we
     should copy their list of devices per folder when connecting.
 
-.. seealso::
-    :ref:`introducer`
+    .. seealso::
+	:ref:`introducer`
 
 .. option:: device.skipIntroductionRemovals
 
--- a/users/config.rst
+++ b/users/config.rst
@@ -430,8 +430,8 @@ The following child elements may exist:
 
 .. option:: folder.order
 
-    The order in which needed files should be pulled from the cluster.
-    The possibles values are:
+    The order in which needed files should be pulled from the cluster.  It has
+    no effect when the folder type is "send only".  The possibles values are:
 
     ``random`` (default)
         Pull files in random order. This optimizes for balancing resources among
--- a/users/config.rst
+++ b/users/config.rst
@@ -314,9 +314,8 @@ element:
 
 .. option:: folder.filesystemType
 
-    .. todo:: Document available filesystem types.
-    .. FILESYSTEM_TYPE_BASIC
-    .. FILESYSTEM_TYPE_FAKE
+    The internal file system implementation used to access this folder, detailed
+    in a :doc:`separate chapter </advanced/folder-filesystem-type>`.
 
 .. option:: folder.path
     :mandatory:
@@ -359,11 +358,17 @@ element:
 
 .. option:: folder.ignorePerms
 
-    True if the folder should ignore permissions.
+    If ``true``, files originating from this folder will be announced to remote
+    devices with the "no permission bits" flag.  The remote devices will use
+    whatever their default permission setting is when creating the files.  The
+    primary use case is for file systems that do not support permissions, such
+    as FAT, or environments where changing permissions is impossible.
 
 .. option:: folder.autoNormalize
 
-    Automatically correct UTF-8 normalization errors found in file names.
+    Automatically correct UTF-8 normalization errors found in file names.  The
+    mechanism and how to set it up is described in a :doc:`separate chapter
+    </advanced/folder-autonormalize>`.
 
 The following child elements may exist:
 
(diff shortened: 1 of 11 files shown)


tomasz1986:
Thanks for doing all the dirty work!

Firstly, I'd like to comment that it will probably be easier to work on this one if https://github.com/syncthing/docs/pull/709 has been accepted first, just to clean up the overlapping changes.

> **Please help getting those documented where possible**

I only wanted to say that can find very detailed explanations for some of the options in their respective commits or PRs. For instance, `filesystemType` is very well documented under https://github.com/syncthing/syncthing/commit/d27463268de5644318fe8bf501278cfb0f2d3bfc. Many of the shorter explanation could probably be almost directly copied and pasted into the config page in the Docs. Some of the longer ones may likely need to have their own respective subpages in the advanced section though.

calmh:
Super cool! Looking at it visually I think the option headings could use some spacing, especially when close to a callout of some sort, but structurally this is sweet.

Screenshot 2022-01-10 at 10 19 47

acolomb:
Yeah, definitely some visual cleanups required, but I haven't investigated the CSS side yet.  Wanted to get feedback on the general idea and implementation first.  Thanks for that!

acolomb:
Fixed the CSS now to look better.

tomasz1986:
Just something that has come to my mind, but with so many potential link target name changes, have you verified for sure that there are no hard-coded links in the Web GUI and/or Syncthing main website that could possible break with this commit?

acolomb:
It's actually only three link targets that have been removed.  The only one apparently linked to from the Web GUI is not actually pointing to the target, but just the HTML file (`versioning.html`).  So no breakage expected.

It will get interesting when we start cleaning up and different Syncthing versions point to the same "master" version of the docs repository, where targets may disappear over time.  Again raises the question if a version of the docs should be built and deployed for each release tag there.  The we could start linking to version-specific docs from the main application.

acolomb:
I'm having a hard time finding any (power-)user-compatible description at all (including commit messages, PR descriptions) about some options.  Need help with those, preferably from whoever wrote the feature (and forgot to make a docs PR :wink: ).

* announceLANAddresses
* sendFullIndexOnUpgrade
* featureFlag
* pullerMaxPendingKiB
* junctionsAsDirs

acolomb:
Otherwise I think this is ready for a (thorough) review now.

I've seen a lot of warning messages when doing a partial rebuild with Sphinx.  That is caused by the new extension finding duplicates in the automatically cached lookup data between build runs.  Not sure how to handle that properly, but it's only cosmetic.

tomasz1986:
> * announceLANAddresses

From https://forum.syncthing.net/t/syncthing-v1-10-0: `This release adds the config option announceLANAddresses to enable (the default) or disable announcing private (RFC1918) LAN IP addresses to global discovery.`

> * sendFullIndexOnUpgrade

From: https://forum.syncthing.net/t/syncthing-v1-11-0-v1-11-1: `This release adds the sendFullIndexOnUpgrade option to control whether all index data is resent when an upgrade is detected, equivalent to starting Syncthing with --reset-deltas. This (sendFullIndexOnUpgrade=true) used to be the behavior in previous versions, but is mainly useful as a troubleshooting step and causes high database churn. The new default is false.`

> * pullerMaxPendingKiB

From https://forum.syncthing.net/t/help-optimizing-for-10gbps-lan/17101/8: `pullerMaxPendingKiB and maxConcurrentIncomingRequestKib are related to how much data we agree to have in flight. Namely pullerMaxPendingKiB controls when we stop sending requests to other devices once we’ve got this much unserved requests.`

> * junctionsAsDirs

It's this PR: https://github.com/syncthing/syncthing/pull/6907. There is a comment in the code saying that `NTFS directory junctions can be treated as ordinary directories`.

acolomb:
Thanks @tomasz1986 ,you're way better than the GitHub search function obviously. That's one place I didn't look. Will incorporate them once back at the desk.

tomasz1986:
No problem. The forum is actually a real treasure of useful information. I like to browse and read through some of the older stuff (i.e. 2014 and up) as a way to relax 🙂.

tomasz1986:
Just a side note, but this commit fixes #310, fixes #375, and fixes #442 (and possibly even more issues?). I'd add all of these to the commit title.

acolomb:
Last remaining options to be documented:
* stunKeepaliveStartS
* stunKeepaliveMinS (these two replaced stunKeepaliveSeconds)
* featureFlag

> Just a side note, but this commit fixes #310, fixes #375, and fixes #442 (and possibly even more issues?). I'd add all of these to the commit title.

I'm not sure we should consider #310 fixed?  It's just a minimal description of what the option does, the concept from a user's perspective is still not explained anywhere.

I linked the other two issues.  No need to have it in the commit title, as there is no such automation as on the main repo.

tomasz1986:
> I linked the other two issues. No need to have it in the commit title, as there is no such automation as on the main repo.

Yeah, although I personally find it useful to have commits and issues clearly linked like that. You can then view the commit history (or use the search) and immediately see which issues a specific commit is related to without even opening it.

acolomb:
Thanks @tomasz1986 for the pointers.  I added some text for the remaining options.

> Then, `stunKeepaliveStartS` and `stunKeepaliveMinS`, related to QUIC, were introduced in [syncthing/syncthing@e714df0](https://github.com/syncthing/syncthing/commit/e714df013f49a0fbfc84bc4b1a954b25bb0f9747) much later on 2019-05-29. There are some comments in the code about the two options, but I haven't been able to find any clear-cut sentence that could be used as a proper description.

At least someone had complained before.  However no issue was recorded, not to mention docs updated: :-(

https://forum.syncthing.net/t/docs-stunkeepaliveseconds-is-now-stunkeepalivestarts-stunkeepalivemins/13736

> I think one important thing to note is that multiple `featureflag` elements are accepted in `config.xml`.

I kept the text intentionally vague because these are intended precisely to only have brave and experienced users test possible features, after reading any available announcement and warnings about them.

Final review, please and ready for merging from my point of view.

users/config.rst
```diff
@@ -298,100 +305,141 @@ One or more ``folder`` elements must be present in the file. Each element
...
+    .. seealso::
+	:doc:`versioning`
+
+.. option:: folder.copiers
+	    folder.hashers
 
-copiers, pullers, hashers
-    The number of copier, puller and hasher routines to use, or ``0`` for the
+    The number of copier and hasher routines to use, or ``0`` for the
     system determined optimums. These are low level performance options for
```
tomasz1986, line 417:
```suggestion
    system determined optimums. These are low-level performance options for
```

users/config.rst
```diff
@@ -298,100 +305,141 @@ One or more ``folder`` elements must be present in the file. Each element
...
-order
+.. option:: folder.pullerMaxPendingKiB
+
+    Controls when we stop sending requests to other devices once we’ve got this
+    much unserved requests.  The number of pullers is automatically adjusted
+    based on this desired amount of outstanding request data.
+
+.. option:: folder.order
+
     The order in which needed files should be pulled from the cluster.
```
tomasz1986, line 429:
```suggestion
    The order in which needed files should be pulled from remote devices.
```

users/config.rst
```diff
@@ -298,100 +305,141 @@ One or more ``folder`` elements must be present in the file. Each element
...
     The possibles values are:
 
-    random
+    ``random`` (default)
         Pull files in random order. This optimizes for balancing resources among
         the devices in a cluster.
```
tomasz1986, line 434: Same as above, but I'm not sure what to change "cluster" to here. Maybe "among the connected devices" or "among the devices sharing the same folder"?

users/config.rst
```diff
@@ -400,97 +448,125 @@ order
...
+
+.. option:: folder.caseSensitiveFS
+
+    Affects performance by disabling the extra safety checks for case
+    insensitive filesystems.  The mechanism and how to set it up is described in
+    a :doc:`separate chapter </advanced/folder-caseSensitiveFS>`.
+
+.. option:: folder.junctionsAsDirs
+
+    NTFS directory junctions can be treated as ordinary directories, if this is
```
tomasz1986, line 567:
```suggestion
    NTFS directory junctions are treated as ordinary directories, if this is
```

users/config.rst
```diff
@@ -612,37 +698,61 @@ address
...
+    The mechanism is described in detail in a :doc:`separate chapter
+    </advanced/device-allowednetworks>`).
+
+.. option:: device.autoAcceptFolders
+
+    If ``true``, folders shared from this remote device are automatically added
+    and synced locally under the :opt:`default path <defaults.folder>`.  For the
+    folder name, Syncthing tries to use the label from the remote device, and if
+    the same label already exists, it then tries to use the folder's ID.  If
+    that exists as well, it just logs an error.  A local folder already added
```
tomasz1986, line 717: I think it actually asks to add the folder manually in such a case? Does it still log an error? It's been a while since I tested these, but I don't remember seeing any actual errors. I may be wrong, of course, but if possible, please re-check this one.

users/config.rst
```diff
@@ -612,37 +698,61 @@ address
...
     address which is currently used for synchronization.  Only the TCP port is
     exchanged for the value specified here.  Note that any port forwarding or
     firewall settings need to be done manually and the link will probably not
     work for link-local IPv6 addresses because of modern browser limitations.
 
+.. option:: device.untrusted
+
+    This boolean value marks a particular device as untrusted, which disallows
+    ever sharing any unencrypted data with it.  Every folder shared with that
+    device then needs an encryption password set.  Refer to the detailed
```
tomasz1986, line 753: I think you can also share receive encrypted folders with such devices, can't you? Those don't need to have a password set on the device that shares them.

users/config.rst
```diff
@@ -980,6 +1173,75 @@ setLowPriority
     disable this behavior, for example to control process priority yourself
     as part of launching Syncthing, set this option to ``false``.
 
+.. option:: options.maxFolderConcurrency
+
+    This option controls how many folders may concurrently be in I/O-intensive
+    operations such as syncing or scanning.  The mechanism is described in
```
tomasz1986, line 1179: Just a side note, as I know this text comes from the already existing subpage, but in most cases syncing is not I/O intensive, especially outside of a local LAN, where bandwidth is almost always the limit, not the I/O. Even on a local LAN, I have personally not seen a case of I/O being the problem. It has always been the connection itself.

Only a digression; nothing to do here.

users/config.rst
```diff
@@ -980,6 +1173,75 @@ setLowPriority
...
+
+.. option:: options.crashReportingURL
+
+    Server URL where :doc:`automatic crash reports <crashrep>` will be sent if
+    enabled.
+
+.. option:: options.crashReportingEnabled
+
+    Switch to opt out from the :doc:`automatic crash reporting <crashrep>`
+    feature to avoid Syncthing "phoning home" on serious troubles.  Defaults to
```
tomasz1986, line 1190: I'd suggest to make this one explicit, i.e. replace "phoning home" with "sending panic logs" (with a `ref` to https://docs.syncthing.net/users/crashrep.html).

users/config.rst
```diff
@@ -980,6 +1173,75 @@ setLowPriority
...
+
+    Controls how Syncthing uses the backend key-value database that stores the
+    index data and other persistent data it needs.  The available options and
+    implications are explained in a :doc:`separate chapter
+    </advanced/option-database-tuning>`.
+
+.. option:: options.maxConcurrentIncomingRequestKiB
+
+    This limits how many bytes we have "in the air" in the form of response data
+    being read and processed.
```
tomasz1986, line 1203: I think this one needs to be explained more, in more "human language", if possible. Right now, I can't understand it, let alone any even less technical person.

users/config.rst
```diff
@@ -980,6 +1173,75 @@ setLowPriority
...
+
+.. option:: options.announceLANAddresses
+
+    Enable (the default) or disable announcing private (RFC1918) LAN IP
+    addresses to global discovery.
```
tomasz1986, line 1208: Maybe add a link to https://datatracker.ietf.org/doc/html/rfc1918?

users/config.rst
```diff
@@ -980,6 +1173,75 @@ setLowPriority
...
+
+.. option:: options.sendFullIndexOnUpgrade
+
+    Controls whether all index data is resent when an upgrade is detected,
```
tomasz1986, line 1212:
```suggestion
    Controls whether all index data is resent when an upgrade has happened,
```

users/config.rst
```diff
@@ -980,6 +1173,75 @@ setLowPriority
...
+    equivalent to starting Syncthing with :option:`--reset-deltas`.  This used
+    to be the default behavior in previous versions, but is mainly useful as a
```
tomasz1986, line 1214:
```suggestion
    to be the default behavior in older versions, but is mainly useful as a
```
It's been quite a while, so I'd suggest "older".

users/config.rst
```diff
@@ -980,6 +1173,75 @@ setLowPriority
...
+    troubleshooting step and causes high database churn. The default is now
+    ``false``.
+
+.. option:: options.featureFlag
+
+    Feature flags are simple strings that, when added to the configuration, may
+    unleash unfished or still-in-development features to allow early user
```
tomasz1986, line 1221:
```suggestion
    unleash unfinished or still-in-development features to allow early user
```

users/config.rst
```diff
@@ -980,6 +1173,75 @@ setLowPriority
...
+.. option:: options.connectionLimitMax
+
+    The maximum number of connections which we will allow in total, zero meaning
+    no limit.  Affects incoming connections and prevents attempting outgoing
+    connections.  The mechanism is described in detail in a :doc:`separate
+    chapter </advanced/option-connection-limits>`.
+
+.. option:: options.insecureAllowOldTLSVersions
+
+    Only for compatibility with old devices, as detailed in
```
tomasz1986, line 1241:
```suggestion
    Only for compatibility with old versions of Syncthing, as detailed in
```

users/config.rst
acolomb: What's wrong with the word "cluster"?  It's technically correct, though not a technical term: "a collection of things somehow tied together".  If there's no obvious better term, let's just leave it at that.

Note that we're looking at the detailed documentation for an XML file format here.  As such we don't necessarily need to aim for Average Joe to understand everything.

users/config.rst
acolomb: You're right.  I just checked the code again and as far as I can tell it just gets the normal treatment.  The log message is just at INFO level.

users/config.rst
acolomb: I don't know. I followed what was described on the [dedicated page](https://docs.syncthing.net/users/untrusted.html#gui):

> If you never want to sync any unencrypted data to a particular device, you can mark it as untrusted in the advanced tab of the device edit dialog. This will prevent a connection to that device if you forget to set a password on any folder shared with it.

users/config.rst
acolomb: It already is a ref to that page.  "Phoning home" is an established term AFAICT, which should trigger anybody who's sensitive to such behavior in a piece of software.  That was exactly my goal.

users/config.rst
acolomb: Me neither.  I'm living off what I could find as my understanding of Syncthing's code is also limited.  Whoever has time / energy to improve it, please do so.  But let's keep that in a separate PR.  My goal was to have completeness and not introduce additional TODO boxes.

users/config.rst
acolomb: Honestly, I'd avoid it.  Who knows when it will break and linking to it raises the expectations regarding how closely the standard was followed in Syncthing's code.  Which is probably rather "pragmatic" :wink:

users/config.rst
acolomb: Is that an upgrade of the local Syncthing executable?  If so, your suggestion makes sense.  Though if it also includes upgrades remote devices, then "detect" is probably appropriate.

users/config.rst
tomasz1986: I treat "cluster" similarly to "node" and "repository". All are present in the English dictionary, but in reality they're not used by the general population (when it comes to computers and such). As for Syncthing, "nodes" were replaced by "devices", "repositories" by "folders" a few years ago. "Cluster" was left out in the process.

I just think it's beneficial to use simpler terms unless there is no other option. While the average Joe is unlikely to come and explore the docs on his own, he is very often sent here from the forum, so the easier to understand the explanations are, the better.

users/config.rst
acolomb: Thanks, lol!

users/config.rst
tomasz1986: "Phoning home" is usually used with a negative connotation though. I don't think you'll find any other piece of software calling its own data or crash collection with this particular term.

users/config.rst
tomasz1986: It is the local executable. I had this option enabled for a few months, and I also remember the same behaviour when it always took place after upgrading Syncthing.

Technically, it doesn't even have to be an "upgrade". Simply replacing the executable with a different one (e.g. a custom-built or an older version) will trigger the index exchange.

users/config.rst
tomasz1986: Yeah, there will be exceptions. Still, when people hear about "phoning home", it's usually closer to something like https://techsupportalert.com/how-to-stop-windows-10-from-phoning-home or https://news.softpedia.com/news/samsung-phones-said-to-come-with-chinese-spyware-phoning-home-528768.shtml, which implicitly indicates nefarious activities, so I'd still suggest to use a more "neutral" expression here.

users/config.rst
imsodin: The intention is definitely that you can share an already encrypted folder with an untrusted device. I haven't thought about that when implementing though, so not sure it works. I'd consider it a bug though if it didn't.

users/config.rst
imsodin: I agree with Tomasz here: We want people to give a choise, but we also want people to send crash reports. Just the neutral statement he proposed (to prevent Syncthing from sending crash reports) is more appropriate and even has more meaning (phoning home can mean basically anything).

users/config.rst
tomasz1986: I've just tested this. I can confirm you can share an already encrypted folder with an untrusted device. The only caveat is that the folder type must be manually adjusted to "receive encrypted" on the untrusted side, but other than that, everything seems to be working as expected.

acolomb: So, any suggestion what changes are needed in the text?

users/config.rst
acolomb: Done.

users/config.rst
acolomb: I committed a possible solution.  It also documents the folder.type = `receiveencrypted`, which was previously missed.

users/config.rst
acolomb: Well, "node" is closely related to graphs and "repository" is technical, both rather uncommon words.  

> "Cluster" was left out in the process.

And I think it's still ok to use.

tomasz1986: "Cluster", at least regarding computing, is on the same level as the other two, really. I mean, there's a reason why cloud/sync solutions aimed at the general public (e.g. Dropbox, Google Drive, OneDrive, etc.) don't use such terminology in their documentation, unless it's from the get-go very clearly directed strictly at developers and/or power users.

I'm not saying this needs to be addressed in this commit, but the question remains whether the documentation (and wording in general) should be accessible to all users of Syncthing, including non-technical people, or only targeted at techies and such. I'd also keep in mind that we're in the general "usage" section of the Docs here, not "development" or even "advanced".

acolomb: 1. Dropbox etc. use a client-server model, where "cluster" is not as meaningful as for a decentralized system.
2. This page explains how to interpret entries in an XML file.  Possibly will get linked to from the "advanced" configuration panel.  It's directly strictly at developers and power users, otherwise we should definitely get rid of the large XML quotes :wink:

Glad to hear you don't insist on this discussion being finished within this PR.

tomasz1986: > Dropbox etc. use a client-server model

Dropbox is a little more complex than that, e.g. it also supports direct sync between LAN devices without downloading files from the central server, but let's not go into such details here.

> This page explains how to interpret entries in an XML file.  Possibly will get linked to from the "advanced" configuration panel.

Yeah, this is a different problem, I'd say mainly about a discrepancy between the Web GUI and the Docs. The two are structured differently, and what's clearly divided in the GUI between "basic" and "advanced" configuration, in the Docs is all simply presented at the same page. As such, it becomes the source of information for both power users and beginners. I don't think there's an easy solution to this problem, but I personally just always prefer to stick to using language that is easy and understandable by everyone (which includes beginners, of course, but also non-native English speakers, especially since the Docs are not available in other languages yet).

acolomb: Actually there are a few help links from the regular "Settings" dialog which lead directly to the detailed sub-pages here (e. g. GUI listen address, Automatic upgrades).  But others link to the config description, which could be improved.  There simply isn't a "novice-compatible" help topic for everything.  Feel free to write up more stuff in more easily understood language and submit it.  The distinction itself makes sense to me, because no regular user wants to scroll through XML code.

users/config.rst
```diff
@@ -298,100 +305,145 @@ One or more ``folder`` elements must be present in the file. Each element
...
-order
+.. option:: folder.pullerMaxPendingKiB
+
+    Controls when we stop sending requests to other devices once we’ve got this
+    much unserved requests.  The number of pullers is automatically adjusted
+    based on this desired amount of outstanding request data.
+
+.. option:: folder.order
+
     The order in which needed files should be pulled from the cluster.
```
tomasz1986, line 433:
```suggestion
    The order in which needed files should be pulled from the cluster. 
    It has no effect when the folder type is "send only".
```

tomasz1986: Just one quick addition I'd like to suggest (actually taken from https://github.com/syncthing/docs/pull/532).

acolomb:
Added the missing link from #715.

Any more reviews / objections?

acolomb:
Merging in light of the upcoming v1.19.0 release, as there have been no objections raised for a week.

PR #711, opened 2022-01-12
Generate a versions.json (ref syncthing/syncthing#8107)

calmh:


acolomb:
Wow. That was quick. Thanks :-)

PR #713, opened 2022-01-18
Redirect to version-specific subdirectory from query parameter.

acolomb:
This implements the URL redirection part of https://github.com/syncthing/syncthing/issues/8107 within the browser.  The same change would need to be applied to all pre-rendered historic versions already deployed to the webserver, if it's supposed to work consistently.

I have some code ready for a version picker widget, which reuses this functionality.  But let's look at the basics first.

Include a custom JavaScript snippet in the built HTML version.  It
checks the ?version=... query parameter and heuristically finds a best
match among the deployed versions list in JSON format.

If a match is found, the browser URL is reconstructed to prepend the
matched version string as a path component.  If the previous first
path component matches a version tag from the list, that is removed
first.  The query parameters get discarded, and the fragment
identifier (#anchor) appended again.  This changed URL is used for a
silent redirect.

The version matching heuristic accepts any version string starting
with a complete known version tag, checking numerically higher ones
first.  For example if v0.12.10 matches first, v0.12.1 is not even
considered anymore.  Incomplete matches are accepted as long as they
only differ after the last dot: v0.13.10 applies to any
v0.13.* (higher or lower) as long as there is no better, direct match.

Any other string in the query parameter leads to the same path with
any known version string in the top-level component removed.  That
should be the default, fall-back version.

### DIFF
--- a/_static/version_redirect.js
+++ b/_static/version_redirect.js
@@ -1,0 +1,82 @@
+var collator = new Intl.Collator(undefined, {numeric: true, sensitivity: 'base'});
+
+const VERSIONS_LIST = "/versions.json";
+
+const getVersions = $.getJSON(VERSIONS_LIST).then(function (data) {
+    // Start with highest version number, using natural sorting
+    data.entries.sort(collator.compare).reverse();
+    return data.entries;
+});
+
+function findBestVersion(version, available) {
+    var bestVersion = '';
+    available.some(function (candidate) {
+        if (version.startsWith(candidate)) {
+            // Direct prefix match
+            bestVersion = candidate;
+            return true;
+        }
+        if (collator.compare(candidate, version) < 0) {
+            // Available version is numerically lower than requested
+            if (version.startsWith(candidate.slice(0, candidate.lastIndexOf('.')))) {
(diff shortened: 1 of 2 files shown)


acolomb:
Use e.g. the following URL for testing:
`https://deploy-preview-713--syncthing-docs.netlify.app/users/config?version=v1.3.1#description`

calmh:
Looks reasonable to me. I don't think we need to worry about historical versions that aren't built/tagged, nor really `/v1.2.3/foo?version=v2.3.4` (shouldn't happen unless someone is manually typing it in?)...

acolomb:
> Looks reasonable to me. I don't think we need to worry about historical versions that aren't built/tagged, nor really `/v1.2.3/foo?version=v2.3.4` (shouldn't happen unless someone is manually typing it in?)...

It will happen when a version picker is added, because that will live under the `/v1.2.3/` hierarchy as well.

acolomb:
Of course the missing tags won't be a problem in the future because they are automated as well within the release procedure?

One more argument to merge the two repositories ;-)

acolomb:
Any objections to merging this?

PR #715, opened 2022-01-21
users/faq: Update info on syncing Windows junctions

tomasz1986:
users/faq: Update info on syncing Windows junctions

Separate hard links and Windows junctions, adding more detailed
information on the latter, also mentioning the requirement to enable
their synchronisation on a per-folder basis.

Ref: https://github.com/syncthing/syncthing/commit/ee445e35a0547ccfa7d239ff558819db9ea94348


acolomb:
A link to the config option would have been great. But of course only works after #710 is merged.

PR #720, opened 2022-01-22
users/versioning: Use lower case in Windows batch example

tomasz1986:
Adjust the case of commands and variables used in the external
versioning example to follow the code used by Microsoft in their
examples.

Ref: https://docs.microsoft.com/windows-server/administration/windows-commands/windows-commands

### DIFF
--- a/users/versioning.rst
+++ b/users/versioning.rst
@@ -170,21 +170,21 @@ behavior as mentioned above. I created the following script and saved it as
     chcp 65001
 
     rem We need command extensions for md to create intermediate folders in one go
-    setlocal EnableExtensions
+    setlocal enableextensions
 
     rem Where I want my versions stored
-    set "VERSIONS_PATH=%USERPROFILE%\.trashcan"
+    set "versions_path=%USERPROFILE%\.trashcan"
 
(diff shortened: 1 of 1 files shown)


users/versioning.rst
```diff
@@ -170,21 +170,21 @@ behavior as mentioned above. I created the following script and saved it as
     chcp 65001
 
     rem We need command extensions for md to create intermediate folders in one go
-    setlocal EnableExtensions
+    setlocal enableextensions
```
imsodin, line 173: Oh nice, the best of all the cASev_AriantS: `novisualseparationnomatterhowmanywordswesticktogether` (only trumped by `SCREAMINGNOVISUALSEPARATION`).

imsodin (approved):
¯\_(ツ)_/¯

(Sincerely thanks a lot for working on the docs! Just that for this particular change I don't have much more to say than that :) ).

users/versioning.rst
tomasz1986: Yeah, although to be fair, these old commands are usually short enough not make any real difference regardless of the case. Note that environment variables still use UPPERCASE 😉.

PR #721, opened 2022-01-23
Add version picker box at the top

acolomb:
Add a header note allowing to pick a specific version.  It is populated from the `versions.json` file and redirects to the specified version when changed.

Screenshot 2022-01-23 at 15-33-41 Welcome to Syncthing’s documentation — Syncthing v1 documentation

To-Do
-----

* [x] Preset the current version in the picker
* [x] When merged, replicate this into all the historic versions at https://github.com/syncthing/docs-pre-rendered

### DIFF
--- a/_static/version_redirect.js
+++ b/_static/version_redirect.js
@@ -51,18 +51,22 @@ function stripVersionPath(path, versions) {
     return path;
 }
 
-function redirectToPath(newPath) {
+function redirectToPath(newPath, keepHistory) {
     const fragment = window.location.href.indexOf('#');
     if (fragment != -1) {
         newPath += window.location.href.slice(fragment);
     }
 
     if (newPath && newPath != window.location.pathname) {
-        window.location.replace(newPath);
+        if (keepHistory) {
+            window.location.assign(newPath);
+        } else {
+            window.location.replace(newPath);
+        }
     }
 }
 
-function redirectToVersion(target, available) {
+function redirectToVersion(target, available, keepHistory) {
     const tail = stripVersionPath(window.location.pathname, available + [target]);
 
     var newPath = '';
@@ -72,7 +76,7 @@ function redirectToVersion(target, available) {
     if (tail) {
         newPath += tail;
     }
-    redirectToPath(newPath);
+    redirectToPath(newPath, keepHistory);
 }
 
 function setVersionPickerOptions() {
@@ -99,7 +103,7 @@ function setVersionPickerOptions() {
 function pickVersion() {
     getVersions.then(function (available) {
         const targetVersion = document.getElementById('version-picker').value;
-        redirectToVersion(targetVersion, available);
+        redirectToVersion(targetVersion, available, true);
     });
 }
 
@@ -111,7 +115,7 @@ const versionParam = urlParams.get('version');
 if (versionParam) {
     getVersions.then(function (available) {
         const useVersion = findBestVersion(versionParam, available);
-        redirectToVersion(useVersion, available);
+        redirectToVersion(useVersion, available, false);
     });
 }
 
--- a/_static/version_redirect.js
+++ b/_static/version_redirect.js
@@ -1,5 +1,6 @@
 var collator = new Intl.Collator(undefined, {numeric: true, sensitivity: 'base'});
 
+// Override MIME type to avoid "invalid XML" warnings when using $.getJSON()
 $.ajaxSetup({beforeSend: function (xhr) {
     if (xhr.overrideMimeType) {
         xhr.overrideMimeType("application/json");
@@ -41,11 +42,14 @@ function findBestVersion(version, available) {
     return bestVersion;
 }
 
-function stripVersionPath(path, versions) {
+function splitVersionPath(path, versions) {
+    // Find end of first path component, disregarding leading slash
     var slash = path.indexOf('/', 1);
     if (slash != -1) {
-        if (versions.indexOf(path.slice(1, slash)) != -1) {
-            return [path.slice(1, slash), path.slice(slash)];
+        var firstComponent = path.slice(1, slash);
+        if (versions.indexOf(firstComponent) != -1) {
+            // Component is a valid known version path, split it off
+            return [firstComponent, path.slice(slash)];
         }
     }
     return ['', path];
@@ -67,7 +71,7 @@ function redirectToPath(newPath, keepHistory) {
 }
 
 function redirectToVersion(target, available, keepHistory) {
-    const tail = stripVersionPath(window.location.pathname, available + [target])[1];
+    const tail = splitVersionPath(window.location.pathname, available + [target])[1];
 
     var newPath = '';
     if (target) {
@@ -84,7 +88,7 @@ function setVersionPickerOptions() {
         var items = [
             '<option value="">latest</option>'
         ];
-        var current = stripVersionPath(window.location.pathname, available)[0];
+        var current = splitVersionPath(window.location.pathname, available)[0];
         $.each(available, function (key, val) {
             var item = '<option value="' + val + '"';
             if (val == current) item += ' selected';
--- a/_static/version_redirect.js
+++ b/_static/version_redirect.js
@@ -69,6 +69,34 @@ function redirectToVersion(target, available) {
     redirectToPath(newPath);
 }
 
+function setVersionPickerOptions() {
+    getVersions.then(function (available) {
+        var items = [
+            '<option value="">latest</option>'
+        ];
+        $.each(available, function (key, val) {
+            items.push('<option value="' + val + '">' + val + '</option>');
+        });
+        let sel = document.createElement('select');
+        sel.setAttribute('id', 'version-picker');
+        sel.onchange = pickVersion;
+        sel.innerHTML = items.join('');
+        let note = document.createElement('div');
+        note.classList.add('admonition', 'note');
+        note.append('Browsing documentation for version: ');
+        note.append(sel);
+        var doc = document.getElementsByClassName('document')[0];
+        doc.prepend(note);
+    });
+}
+
+function pickVersion() {
+    getVersions.then(function (available) {
+        const targetVersion = document.getElementById('version-picker').value;
+        redirectToVersion(targetVersion, available);
+    });
+}
+
 
 const urlParams = new URLSearchParams(window.location.search);
 const versionParam = urlParams.get('version');
@@ -80,3 +108,5 @@ if (versionParam) {
         redirectToVersion(useVersion, available);
     });
 }
+
+window.addEventListener('DOMContentLoaded', setVersionPickerOptions);
--- a/_static/version_redirect.js
+++ b/_static/version_redirect.js
@@ -1,5 +1,11 @@
 var collator = new Intl.Collator(undefined, {numeric: true, sensitivity: 'base'});
 
+$.ajaxSetup({beforeSend: function (xhr) {
+    if (xhr.overrideMimeType) {
+        xhr.overrideMimeType("application/json");
+    }
+}});
+
 const VERSIONS_LIST = "/versions.json";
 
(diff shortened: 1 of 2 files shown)


acolomb:
So far this is mainly to get feedback on the design / visual integration.

The picker works, but of course is not integrated into all historic versions yet.  I've experimented on that, and it seems the easiest way is to rebuild all of them after patching in the changes to `conf.py` and the added `version_redirect.js` each time.  I [created a draft](https://github.com/acolomb/syncthing-docs-pre-rendered/commits/inject-version-picker) `Makefile` for this purpose, but as my build environment seems rather different, it results in a truckload of changes to every version.  Happy to get feedback / help on that aspect.  We should postpone actually doing that until this one is merged though.

tomasz1986:
> So far this is mainly to get feedback on the design / visual integration.

I've had a quick look at the preview. One potential problem I can think of is its potential integration with older versions of the docs that use a completely different layout, e.g. https://deploy-preview-721--syncthing-docs.netlify.app/v0.11.8/. Have you thought about moving the bar to the very top of the browser window, maybe taking the full width in the process? After that, it could also possibly be stuck to the top, but then it will always cover a part of the window, probably getting annoying very quickly.

The current script also doesn't work in IE11, because of its very limited support for `let` and `const` that the code is filled in with. I may try to do something about it later in case you either don't feel like it or have no access to the browser.

acolomb:
Thanks for looking into it.  I thought it looked awful covering the whole window width, but that's subjective.  Just using the `admonition note` classes is very convenient, but has the side-effect of padding around the box.  Haven't checked what it looks like in older versions, but will do that soon.  Frankly I don't care much about the looks of those very old versions as long as it works at all.  Whoever would want to browse them? :wink: 

Regarding the IE 11 stuff, I'm really reluctant to keep supporting old this crappy browser if it causes headaches.  Some things I looked up and decided to use a compatible alternative (such as `String.startsWith()`), but especially with the basic JS language constructs, I just don't have an overview of what works.  `const` seems like a really useful thing, but in this context it's easy to check manually what code touches a variable.  I don't even grasp the difference why `let` is needed next to `var`, just copy-pasted it from some partial problem solutions on the webs.  So if you feel like helping out on that front, be my guest.  If not, we'll just diregard IE 11 shortcomings I suppose.

tomasz1986:
From the screenshot above, I'd say the picker actually seems to look and integrate better in v0.11.13 and it does in the current one… The yellowish background stands out quite a bit, to be frank.

I think I've found a solution to the IE error, but is there any simple way to test the picker locally? If I do the standard `make html`, the generated page is loaded without it (regardless of the browser).

acolomb:
I use https://ltworf.github.io/weborf/ as a mini-webserver to serve it locally.  The problem with browsing local builds is that each file URL is considered a separate "domain" in the browsers for security reasons.  Hence it will not load the `versions.json` file and fail gracefully by not showing anything.

acolomb:
> I created a draft `Makefile` [in my fork](https://github.com/acolomb/syncthing-docs-pre-rendered/commits/inject-version-picker) for this purpose, but as my build environment seems rather different, it results in a truckload of changes to every version. Happy to get feedback / help on that aspect.

I switched to using `./docker-build.sh`, but that environment still seems to generate different contents than what is committed at https://github.com/syncthing/syncthing-docs-pre-rendered.  What did you use to build @calmh?

calmh:
I used whatever happens to be the local sphinx on my computer, because I'm lazy and the output looked reasonable. We used to use `./docker-build.sh` or something similar to build with the kastelo/sphinx-build image, and I guess may do something like that again in the future. The Standard build is the one done by Netlify, based on requirements.txt and netlify-build.sh...

tomasz1986:
Just a suggestion, but how about making the font size a little smaller, e.g. to be the same as the sidebar, which is `14px`? The reason for this is make the version picker seem a little less important, as right now, with the huge font and the yellow background, it looks as it it was the most important element on the page 😛.

The font of the selector also doesn't match the font used for the rest of the text. I think it'd look nicer if the two were identical. I've checked the CSS and the simplest solution may be just to apply `font: inherit` to the `select` element.

acolomb:
These styling issues are a can of worms.  I added the `font: inherit` attribute, but font size, background color etc. is so much dependent on the theme in use that I don't want to spend time rebuilding over and over just to make it look pretty in every version.  Testing this is time-consuming, as every change needs to be committed, a patch generated, and then every historic version rebuilt with the patch applied.  Then commit all those changes back into the (local) pre-rendered repo in order to deploy it together with the docs repo build again.  Not fun.

Thus I stayed as close as possible to standard Sphinx / docutils styling classes that should work on any theme.  If anything, we can optimize on the currently used theme, but only using future-proof standard CSS classes please.

tomasz1986:
Yeah, without using any absolute values, the font size needs to be set to `90%` to match the size used in the sidebar. The only standard Sphinx class that would fit seems to be `related`. It does work fine in the current theme when applied to the `div`.

acolomb:
I added a small hack to make the deploy preview work on Netlify, by pointing it at my fork of the docs-pre-rendered repo.  It contains all historic versions rebuilt with the version picker patch applied.  Have fun testing!

(Yeah, it doesn't actually show the current version yet. Still undecided whether hinging that on the first path component is a good idea.)

tomasz1986:
I've checked the hacked preview, and I'd say the biggest problem with the selector seems to be that after each switch, it keeps resetting the selection to `latest`. This means that the text always says `Browsing documentation for version: latest` instead of displaying the actual version.

acolomb:
> Now, looking again at it, I guess this may be exactly what you meant here :sweat_smile:.

Exactly.

As for the 90% font size, that stuff really doesn't belong into a JS file, because it's hard to find where it came from.  Seems too much hassle for so little gain.

tomasz1986:
> As for the 90% font size, that stuff really doesn't belong into a JS file, because it's hard to find where it came from. Seems too much hassle for so little gain.

Is there a better place to place it in though? It could go to a CSS file (possibly with an intention to only apply to the current theme), but then we're left with static CSS code paired with a dynamically generated content, which is bad practice in general, so…	

The font's now got the same size as the `body` text, but for me it still appears somehow "larger" and more "in your face", possibly due to the its top position and the yellow background, hence the idea to make it a little smaller :wink:.

acolomb:
> Is there a better place to place it in though? It could go to a CSS file (possibly with an intention to only apply to the current theme), but then we're left with static CSS code paired with a dynamically generated content, which is bad practice in general, so…

Adding any of this to the historic docs means adding a `<script>` tag to every single HTML file.  I took a shortcut to do that by rebuilding all of them, but might just as well limit the final commit to only this one-line change (per file, per version).  If we wanted to add some CSS as well, that would be another patch line per file, per version.  To keep things simple, I want to limit changes to the absolutely necessary.

That said, we can of course adjust the styling of this for future docs versions.  Should not introduce a total visual break  though, so the note looks roughly the same everywhere (except for the ancient theme).  The fact it's yellow comes from our `custom.css` overrides by the way.  That's also been in use for many versions, so not something I'd like to touch.  We can use `.hint` instead of `.note`, which gives a grayish background.

acolomb:
@tomasz1986 what do you think of the gray box instead of yellow?

tomasz1986:
I'd say it's much better this way 🙂. Same functionality, but less noticeable and more out of the way.

acolomb:
I'm still a bit reluctant to cater for ancient browsers given this is a change that will be replicated to many historic versions.  What would we really lose on those browsers?  The ability to switch versions and the redirecting feature (which is already merged, in the non-IE11-compatible form).  Users of this browser will always land on a valid and up-to-date documentation version.  Historians trying to scroll through ancient docs will be out of luck or must manually change the URL.  People on old Syncthing versions will land on the HEAD version anyway.  Users of v(1.19.0)+1 will land there as well, instead of the specific version page.  Once that version is so outdated that the HEAD docs don't make sense anymore, I hope that IE11 will have finally ceased to exist.  What do others think on this matter @syncthing/maintainers?

This is good to go from my side.  We'll first need to merge https://github.com/syncthing/docs-pre-rendered/pull/2, then revert the temporary "TESTING ONLY" change here and finally squash-merge this one.

tomasz1986:
> I hope that IE11 will have finally ceased to exist.

I always mention IE11, because it's the one that I've got to use (e.g. for some ancient government and education sites). However, there's also a growing problem of old Android devices that are stuck on outdated browsers with no future updates possible.

Anyhow, as for the IE11 compatibility, I'm going to give it a shot later. This should in no way be an obstacle to merge this commit now 🙂.

_static/version_redirect.js
```diff
@@ -1,5 +1,11 @@
 var collator = new Intl.Collator(undefined, {numeric: true, sensitivity: 'base'});
 
+$.ajaxSetup({beforeSend: function (xhr) {
+    if (xhr.overrideMimeType) {
+        xhr.overrideMimeType("application/json");
+    }
+}});
```
calmh, line 7: What's this? (Might deserve a comment)

_static/version_redirect.js
```diff
@@ -39,25 +45,29 @@ function stripVersionPath(path, versions) {
     var slash = path.indexOf('/', 1);
     if (slash != -1) {
         if (versions.indexOf(path.slice(1, slash)) != -1) {
-            path = path.slice(slash);
+            return [path.slice(1, slash), path.slice(slash)];
```
calmh, line 48: Is this now a `split` function rather than a `strip`? This could also do with a comment, because I'm a bit lost among the slicing

acolomb: Right, this function has grown beyond its original purpose, so better to rename and comment.

_static/version_redirect.js
```diff
@@ -66,7 +76,40 @@ function redirectToVersion(target, available) {
...
+        var items = [
+            '<option value="">latest</option>'
+        ];
+        var current = stripVersionPath(window.location.pathname, available)[0];
+        $.each(available, function (key, val) {
+            var item = '<option value="' + val + '"';
+            if (val == current) item += ' selected';
+            item += '>' + val + '</option>';
+            items.push(item);
+        });
```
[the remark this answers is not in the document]
acolomb: That becomes quite unreadable IMHO, as even the first (fixed) option line needs at least four to five JS function calls to replicate.  I can try it if you think it's worth the trouble.  This approach with `innerHTML` was taken from some examples on the webz, and I think it's fine as long as there is no nesting of HTML elements within the items.

_static/version_redirect.js
acolomb: Trying to avoid "invalid XML" warnings on some browsers, notably Firefox.  Comment added.

acolomb:
@calmh are you okay with the looks and implementation of this now? Then I could go ahead and mirror it to the docs-pre-rendered repo later today.

acolomb:
Merged after not receiving any more objections ;-)

tomasz1986:
I'm sorry for a late comment, but I'm browsing the Docs right now, and I must say that the functionality is great, but there is one problem. As the picker is always added to the page last, it causes the whole page to jump. This happens constantly upon opening new pages when navigating through the Docs.
 c
The problem may not be noticeable when browsing locally or if you've got a super fast Internet connection, but here on my side the jumping is very noticeable, and I must say it gets annoying very quickly. Would it be perhaps possible to load the version picker first and only then fill in the rest of the content?

acolomb:
Fair point. I initially had the HTML implemented directly into the template, so it became part of each page's generated code. But that was a more invasive change during rebuilding all the historic versions. And the intention was to show the picker only when it actually works, which we know only after the versions.json file has been correctly parsed. That happens asynchronously and is delayed until after the whole page has loaded so the place where we insert the HTML is guaranteed to exist.

I have some ideas how to fix it, but it will take some time to test and keep the diffs reasonably small.

acolomb:
Improved in #732 with hopefully less jumping.

PR #725, opened 2022-02-04
Update Caddy v2 configuration docs to fix 'host check error'

jeroen7s:
Using the current Caddy configuration shown in the documentation, users get a host check error trying to access Syncthing's web-ui on their servers.
Adding these changes fixes that

### DIFF
--- a/users/reverseproxy.rst
+++ b/users/reverseproxy.rst
@@ -77,7 +77,10 @@ Caddy v2
 
     handle_path /syncthing/* {
         uri strip_prefix /syncthing
-        reverse_proxy http://localhost:8384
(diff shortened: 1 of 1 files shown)


st-review:
:robot: *beep boop*

calmh:
Apparently we never looked at this; sorry! I think this is fine. The host check error is because we're trying to protect the user when Syncthing is supposedly only listening on localhost, but in this case it's presumably open to the world via the proxy. Hence circumventing the error seems valid... (Someone else can correct my thinking if this is wrong.)

jeroen7s:
No problem, thanks for looking at it and merging :+1:

PR #727, opened 2022-02-06
rest: Document missing endpoints

acolomb:
Add documentation for REST API endpoints that were either incomplete or missing at all.  Consolidate the paging mechanism in the top-levelt REST page.  Clean up some cross-reference links and styling along the way.  **See single commit messages for details and links to the introducing PRs.**

The following endpoints still have a TODO note and should be completed after getting additional info here:

* [x] `/rest/db/localchanged` (what does the `version` field tell us?)
* [x] `/rest/debug/*`
* [x] `/meta.js` (not really REST)
* [x] `/qr/` (not really REST)


rest/db-localchanged-get.rst
```diff
@@ -0,0 +1,36 @@
+GET /rest/db/localchanged
+=========================
+
+.. versionadded:: 0.14.55
+
+Takes one mandatory parameter, ``folder``, and returns the list of files which
+were changed locally in a receive-only folder.  Thus they differ from the global
+stateand could be reverted by pulling from remote devices again, see
```
imsodin, line 8:
```suggestion
state and could be reverted by pulling from remote devices again, see
```

rest/db-need-get.rst
```diff
@@ -20,8 +20,8 @@ section in the response will have 10 items. If you issue a request query with
 2 items. If you issue a query for ``page=3`` and ``perpage=10``, you will only
 have the last 10 items of the ``rest`` section.
 
-In all these calls, ``total`` will be 30 to indicate the total number of
-available items.
+.. note:: Return format changed in version 0.14.43, removing the ``total`` count
+          attribute.
```
imsodin, line 24: Nice archealogy-work :)

rest/debug.rst
```diff
@@ -0,0 +1,40 @@
...
+-------------------
+
+.. todo:: No documentation for this endpoint yet!
+
+
+/rest/debug/heapprof
+--------------------
+
+.. todo:: No documentation for this endpoint yet!
+
```
imsodin, line 29: Link to https://docs.syncthing.net/users/profiling.html might suffice here.

rest/folder-versions-get.rst
```diff
@@ -0,0 +1,41 @@
...
+	  "modTime": "2022-02-06T20:44:13+01:00",
+	  "size": 4
+	},
+	{
+	  "versionTime": "2022-02-06T20:44:20+01:00",
+	  "modTime": "2021-01-14T13:21:16+01:00",
+	  "size": 4
+	}
+      ]
+    }
```
imsodin, line 41: Indentation is messed up here (the code-formatting might take care of that?)

imsodin (commented):
Thanks a lot for this!

Just a few minor comments.

>  /rest/db/localchanged (what does the version field tell us?)

What version field?

acolomb:
> > /rest/db/localchanged (what does the version field tell us?)
> 
> What version field?

The one in `message FileInfoTruncated`, which ends up in each file entry.  The other attributes seem obvious enough to me, but the average API consumer might not have any clue what a version vector is and what they would do with the field.

rest/cluster-pending-folders-delete.rst
```diff
@@ -3,21 +3,20 @@ DELETE /rest/cluster/pending/folders
...
-Valid values for the ``folder`` and ``device`` parameters are those
-from the corresponding :doc:`/rest/cluster-pending-folders-get`
-endpoint.  The ``device`` parameter is optional and affects
-announcements of this folder from the given device, or from *any*
-device if omitted.
+Remove records about a pending folder announced from a remote device.  Valid
+values for the ``folder`` and ``device`` parameters are those from the
+corresponding :doc:`cluster-pending-folders-get` endpoint.  The ``device``
+parameter is optional and affects announcements of this folder from the given
+device, or from *any* device if omitted.
```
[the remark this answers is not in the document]
acolomb: I wouldn't exactly say we have a never-ending stream of docs contributions anyway.  Dunno why GitHub doesn't highlight it here, but there is a change in the link actually (dropping /rest/).  It's not a functional change, but I tried to harmonize the links and add missing ones.  Only when touching stuff anyway is when I usually go about reflowing as well.  I'd like to keep that practice, but of course not for random passages of text--that would be tedious to review.

Might do some clean-up in one go at some point.  My auto-formatter would be Emacs' text reflowing function with a limit of 80 chars per line then.  But there is the odd situation where I even overrule it by own judgment, so strict auto-formatting wouldn't be better IMHO.

rest/folder-versions-get.rst
acolomb: Sorry, Emacs likes to mix tabs and spaces by default.  Need to tell it that JSON and reST are not good candidates for that.

imsodin:
> The one in `message FileInfoTruncated`, which ends up in each file entry. The other attributes seem obvious enough to me, but the average API consumer might not have any clue what a version vector is and what they would do with the field.

That has nothing to do with the endpoint though, and other stuff like localflags I'd consider even more obscure - this endpoint just gives you the raw file-info by design. There's https://docs.syncthing.net/users/syncing.html to explain some of it.

rest/cluster-pending-folders-delete.rst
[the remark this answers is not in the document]
imsodin: >  Dunno why GitHub doesn't highlight it here, but there is a change in the link actually (dropping /rest/). I

That's indeed weird and exactly what I missed - thanks for pointing it out. I do agree on doing formatting changes when you anyway change the content, so exactly what you do - all good.

imsodin (approved):

acolomb:
You're right, it's found on several endpoints where it's not really explained.  Will just remove the todo comment and accept that some understanding of the internals is needed to fully grok the response.  Probably not interesting to most API users anyway at this point.

acolomb:
I filled in all the introduced `todo` blanks to the best of my knowledge.  Ready for a final glancing over the last few commits @imsodin.  My aim here was for completeness instead of perfection, so feel free to merge if you don't find any big issues.

imsodin (approved):

imsodin:
Thank you!

PR #728, opened 2022-02-07
users/autostart: Also refer to the repo for .desktop files

imsodin:


acolomb:
Isn't this already included in the paragraph below?

imsodin:
Sorry about this, rebased.

tomasz1986:
The link formatting seems broken (see the deploy preview), but as @acolomb mentioned above, the exact same link is already listed right below this paragraph. Is it necessary to list it twice like that?

imsodin:
Fixed the link.

And yes, some hint is necessary, see https://forum.syncthing.net/t/autostart-on-manjaro/17963

acolomb (approved):

Issue #718, opened 2022-01-22
Should display banner when browsing outdated version of the documentation


calmh:
When browsing the documentation and:
1. you're on a specific version URL (`/v1.2.3/*`), and
2. that version is not the latest version
we should show a banner to the effect of you "You're browsing the documentation for Syncthing v1.2.3 which is not the latest version and may be outdated, click here for the current/latest version".

acolomb:
Right now we don't actually have the version string / tag available in the built docs, do we?  Should probably use a similar mechanism as in the main repo to extract it from `git describe`...

acolomb:
Ok so let's skip this until the merger is done, okay?  Then I'll also integrate a version picker directly into the banner.

PR #729, opened 2022-02-13
Include Git version string in footer and sidebar

acolomb:
Try to fetch the current commit description from "git describe" to dynamically populate the "version" and "release" configuration values.  Add a reference to the template page footer stating the release information, just below the "Edit on GitHub" link.

Additionally, include a new snippet into the navigation sidebar showing the short version number.  The html_sidebars setting needs to explicitly enumerate the default snippets, in order to insert the new one.

Ref. #718.

Feedback on styling / placement is very welcome:

Screenshot 2022-02-13 at 21-50-13 Introduction — Syncthing v1 19 0-4-g6d0ea81-dirty documentation_edited


acolomb:
Seems like the Netlify preview deployment is picking up an unexpected Git version... Should be `v1.19.0-3-g0c55e0a` while it actually displays `v1.18.3-34-g0c55e0a`.  No idea where that may come from.

tomasz1986:
How does it interact with https://github.com/syncthing/docs/pull/721? If there is already going to be a version picker at the top with the current version listed there, is it even necessary to mention the version number in other places?

acolomb:
The idea is that this one here will be hard-coded into the HTML files, while the version picker box will just look at the first URL path component.  Since that doesn't really work offline, I think this could still be a worthwhile addition.

tomasz1986:
> Since that doesn't really work offline, I think this could still be a worthwhile addition.

That makes sense.

> as having it in three places may indeed be too much.

The sidebar one does feel a little bit superfluous 😉. The bottom one looks fine, but how about adding some spacing between the "edit this page…" and the version string?

acolomb:
> The bottom one looks fine, but how about adding some spacing between the "edit this page…" and the version string?

I had that initially, but thought it looked awkward.  And the GitHub link is closely related to the commit reference, so I think it's quite alright to keep them close together.

acolomb:
One more thing to note: The page title includes the `release` info as well, which right now is the elaborate string from `git describe`.  We could switch the `version` and `release` variables to shorten that, although it contradicts the documented semantics of these variables in `conf.py`.

calmh:
Let's try it.

calmh:
Netlify uses a cached repo without pulling tags, so it looks outdated in the live build. This could be fixed in the build script, though it's a bit annoying having to do so...

https://answers.netlify.com/t/git-tags-during-build/16756/17

acolomb:
Understood, let's try your solution in #739.

PR #730, opened 2022-02-14
Update Apache reverse proxy config to include X-Forwarded-Proto

zertrin:
Needed to properly handle secure cookies when using HTTPS
(see https://github.com/syncthing/syncthing/issues/7399#issuecomment-967770293)

### DIFF
--- a/users/reverseproxy.rst
+++ b/users/reverseproxy.rst
@@ -24,16 +24,17 @@ GUI hosted on ``localhost:8384``.
 Apache
 ~~~~~~
 
-First of all, execute the following command to enable the Apache HTTP Proxy
-module: ``a2enmod proxy_http``.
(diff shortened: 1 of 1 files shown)


users/reverseproxy.rst
```diff
@@ -24,16 +24,17 @@ GUI hosted on ``localhost:8384``.
 Apache
 ~~~~~~
 
-First of all, execute the following command to enable the Apache HTTP Proxy
-module: ``a2enmod proxy_http``.
+First of all, execute the following commands to enable the Apache HTTP Proxy
```
calmh, line 27:
```suggestion
First of all, execute the following command to enable the Apache HTTP Proxy
```

calmh (approved):

Issue #712, opened 2022-01-14
Document default ignores


imsodin:
Introduced in https://github.com/syncthing/syncthing/pull/7530

acolomb:
This was wrongfully closed, #727 only documents the default ignores endpoint, but not the configuration side.

PR #737, opened 2022-03-05
users/config: Describe default-ignore approach

salim-b:
I've only now realized how easy it is to sync (potentially) device-independent dot files (and other config files) between computers and would have loved to learn that the first time I read this documentation section. So I added a paragraph describing it.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -1404,3 +1404,19 @@ If you'd like to sync your home folder in non-send only mode, you may add the
 folder that stores the configuration files to the :ref:`ignore list <ignoring-files>`.
 If you'd also like to backup your configuration files, add another folder in
 send only mode for just the configuration folder.
+
+If you dare to actively sync configuration files in use (implies you really know
+what you're doing), you can add a dedicated folder with the path set to the common
+parent path of the configuration files (usually the home directory). At the same
+time you configure a default-ignore approach for this folder where only explicitly
+stated files and subfolders are synced and everything else is ignored.
+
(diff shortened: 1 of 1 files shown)


acolomb (changes requested):
I think this should be moved to the `users/ignoring` section, although there is already an example there.  The paragraph above here is talking explicitly about syncing Syncthing's *own* configuration, which could potentially have bad effects.  A link to the other section could still be added.

Note that this should not be confused with the "default ignores" feature introduced some months ago.   There's #712 still in need of attention for that part of the configuration.  What you mean is more of an "ignore all, except some" approach, for which the term "default" ignores should not be used.

salim-b:
> I think this should be moved to the `users/ignoring` section, although there is already an example there. The paragraph above here is talking explicitly about syncing Syncthing's *own* configuration, which could potentially have bad effects. A link to the other section could still be added.

Funny, I didn't read the **Syncing Configuration Files** section that way because it essentially applies to user configuration files of almost any software, not only syncthing's own. I'm not sure whether `users/ignoring` is the ideal place to put my addition then. Would a dedicated FAQ entry be more appropriate @acolomb ?

> Note that this should not be confused with the "default ignores" feature introduced some months ago. (...)

Ok, thanks for the hint. I agree that I need to change my terminology.

acolomb:
I'm curious, how did you get to the configuration format page then?  What were you searching for?

I agree there is no proper "best practices" section in here, where stuff like this would belong.  There was some effort toward that in the past, but unfortunately stalled at some point.

Finding the info in the "Ignoring" section is not easy, I agree.  But it's the most on-topic we have right now, and the config file format page is definitely not the right place for such recommendations.

st-review:
:robot: *beep boop*

PR #739, opened 2022-03-07
Fetch tags as part of Netlify build

calmh:


acolomb:
In fact I saw the correct tag version `v1.19.1-2-g4a77332` even just before you merged this.  But I guess it doesn't hurt.

PR #740, opened 2022-03-07
Deployment fixes

acolomb:
Avoid `-dirty` string in displayed git version, which could inevitably occur when back-porting any future changes to the docs-pre-rendered repo.

Adjust build tools version requirements to not end up with older versions than what the other repo was built with.


calmh:
The pre-rendered stuff all happens in _build so doesn't affect `--dirty`, but this is fine nonetheless.

PR #741, opened 2022-03-08
events: Mention events not in the default event type mask (fixes #404)

### LINKED ISSUE #404: Event API issues
LocalChangeDetected and RemoteChangeDetected events were not documented correctly. It should be in /rest/events/disk, not /rest/events

acolomb:
Add a note to LocalChangeDetected and RemoteChangeDetected events
about needing to explicitly select them on the /rest/events endpoint.

Simplify cross-references along the way.

### DIFF
--- a/events/localchangedetected.rst
+++ b/events/localchangedetected.rst
@@ -6,8 +6,8 @@ previous scan.  This does *not* include events that are discovered and copied fr
 other devices (:doc:`remotechangedetected`), only files that were changed on the
 local filesystem.
 
-.. note:: This is not included in the default mask for the
-   :doc:`/rest/events-get` endpoint,but needs to be selected explicitly.
+.. note:: This event is not included in :doc:`/rest/events-get` endpoint without
+   a mask specified, but needs to be selected explicitly.
 
 .. code-block:: json
 
--- a/events/localchangedetected.rst
+++ b/events/localchangedetected.rst
@@ -1,12 +1,13 @@
-.. _local-change-detected:
-
 LocalChangeDetected
 -------------------
 
 Generated upon scan whenever the local disk has discovered an updated file from the
 previous scan.  This does *not* include events that are discovered and copied from
-other devices (:ref:`remote-change-detected`), only files that were changed on the
+other devices (:doc:`remotechangedetected`), only files that were changed on the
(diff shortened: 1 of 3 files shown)


events/localchangedetected.rst
```diff
@@ -1,13 +1,14 @@
...
 -------------------
 
 Generated upon scan whenever the local disk has discovered an updated file from the
 previous scan.  This does *not* include events that are discovered and copied from
-other devices (:ref:`remote-change-detected`), only files that were changed on the
+other devices (:doc:`remotechangedetected`), only files that were changed on the
 local filesystem.
 
+.. note:: This is not included in the default mask for the
+   :doc:`/rest/events-get` endpoint,but needs to be selected explicitly.
```
imsodin, line 10:
```suggestion
.. note:: This event is not included in :doc:`/rest/events-get` endpoint without a mask specified, but needs to be selected explicitly.
```

I restructured the sentence, because there's no such thing as an explicit "default mask", you can just not give a mask at all - maybe that's a source of confusion? I am not sure about that part of the change. 
Repeating the "event" and space after comma is unrelated, and also applies to the other change below

imsodin (commented):
The main place where I'd expect a note about what is returned without specifying a mask, is on the endpoint/mask parameter description itself: https://docs.syncthing.net/rest/events-get.html#get-rest-events-disk

acolomb:
From the linked issue, one point is still not addressed:  `ItemStarted` and `ItemFinished` are still contained in the `DefaultEventMask`.  That would need "fixing" in the main repo, but is a potentially breaking change.  Opening a PR just to add a FIXME comment about that seemed overkill ;-)

> The main place where I'd expect a note about what is returned without specifying a mask, is on the endpoint/mask parameter description itself

It used to be there, but was removed during some restructuring, as discussed in the linked issue.  I wouldn't mind if it was mentioned there, especially since the inverse filter on the other (kind of deprecated?) endpoint is on the same page.  But for troubleshooting issues like #404, I think my approach is more likely to get noticed.

events/localchangedetected.rst
acolomb: Right, it was probably getting too late when I made the change.

PR #742, opened 2022-03-08
users/autostart: Mention "loginctl enable-linger" for user services (fixes #319)

### LINKED ISSUE #319: Docs say systemd user units only start when logged in, but this isn't entirely correct
[This documentation section](https://docs.syncthing.net/users/autostart.html#using-systemd) claims that, if Syncthing is to be started as a systemd user service (that is, a `.service` file under `~/.config/systemd/user`), it will only start when the user is logged in.

This is indeed the default behavior, but there is another option here. If the administrator does `loginctl enable-linger {username}`, then that user's services are started when the machine boots, much like system services do.

The documentation should mention this option.

acolomb:


users/autostart.rst
```diff
@@ -404,6 +404,13 @@ How to set up a user service
    Otherwise the user service will not start, because by default, systemd checks for user
    services before your home directory has been decrypted.
 
+Automatic start-up of systemd user instances at boot (before login) is possible
+through systemd's "lingering" function, if a system service cannot be used
+instead.  Refer to the `loginctl enable-linger`__ command to allow this for a
+particular user.
+
+__ https://www.freedesktop.org/software/systemd/man/loginctl.html#enable-linger%20USER%E2%80%A6
```
imsodin, line 412: New link syntax - how does that work if there's more than one? Maybe lets keep using the `text <url>`__ one just to not add even more variability to the syntax?

acolomb: It's called an anyonymous indirect link IIRC and works only once, locally.  Having this long URL inline looked ugly and I think it disturbs the reading flow when editing the document.  We can switch to a regular indirect link (not anonymous) instead.

imsodin: I'd appreciate that.

acolomb: Done, sorry for the delay.

imsodin (approved):

PR #749, opened 2022-04-01
conf: Hide the version from title unless it exactly matches a tag (fixes #748)

### LINKED ISSUE #748: Do not add version number to the website title when browsing the latest version
Since the recent version picker and version string additions, version numbers seem to be always added to the website title. I don't think this is really desired when linking to the generic Docs URL, i.e. without any specific version added to it.

To make it clear what I mean here, below is the title as of right now.

```
<title>Welcome to Syncthing’s documentation! &#8212; Syncthing v1.19.1-13-g0dbd5f4 documentation</title>
```

I believe this will create complexity with search engines, e.g. Google, which will probably add these strings to the title there too, resulting in unnecessary noise. Also, the same thing has already been happening on the forums, where the software automatically grabs the current title and displays it as follows.

image

This is problematic, because a) the actual URL behind the link is just [https://docs.syncthing.net/users/ignoring.html](https://docs.syncthing.net/users/ignoring.html) without any version string, and also, as far as I understand, b) the link on the forum is going to stay like that forever.

In sum, I'd propose to add the version string to the title only when we're actually browsing an older version of the Docs, and skip it when browsing the generic Docs URL or the `latest` version.

acolomb:
Find out during build whether the source is currently described by a
Git tag.  Show that (shorter than the "release" info), or no version
at all otherwise.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -64,8 +64,10 @@ copyright = u'2014-2019, The Syncthing Authors'
 # The full version, including alpha/beta/rc tags.
 try:
     release = os.popen('git describe --tags --long --always').read().strip()
+    _git_tag = os.popen('git describe --tags --exact-match').read().strip()
 except Exception:
     release = 'v1'
+    _git_tag = ''
 # The short X.Y version.
 version = release.partition('-')[0]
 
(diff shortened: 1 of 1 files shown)


calmh (approved):
Looks good

PR #750, opened 2022-04-05
build, dev/building: Update go version to 1.17

tomasz1986:


tomasz1986:
I'm actually not sure about updating the Go version listed in the `netlify.toml` file. Is it also supposed to be updated like that?

PR #751, opened 2022-04-10
Default ignores configuration (fixes #712)

### LINKED ISSUE #712: Document default ignores

acolomb:
Describe the format of the `<defaults><ignores>` XML structure.  Explain when it gets used in the section about ignoring files.

Extend CLI subcommand description to mention the `syncthing cli operations default-ignores` method, albeit without naming the command.  The possibilites in there are meant to be discovered by the help functions, so we only give a rough hint here in the overview.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -1347,10 +1347,10 @@ be present in the ``defaults`` element:
     .. versionadded:: 1.19.0
 
     Template for the :ref:`ignore patterns <ignoring-files>` applied to new
-    folders.  These are automatically copied to the ``.stignore`` file when a
-    folder is automatically accepted from a remote device.  The GUI uses them to
-    pre-fill the respective field when adding a new folder as well.  In XML,
-    each pattern line is represented as by a ``<line>`` element.
+    folders.  These are copied to the ``.stignore`` file when a folder is
+    automatically accepted from a remote device.  The GUI uses them to pre-fill
+    the respective field when adding a new folder as well.  In XML, each pattern
+    line is represented as by a ``<line>`` element.
 
 
 .. _listen-addresses:
--- a/users/config.rst
+++ b/users/config.rst
@@ -1309,6 +1309,13 @@ Defaults Element
             <untrusted>false</untrusted>
             <remoteGUIPort>0</remoteGUIPort>
         </device>
+	<ignores>
+            <line>!foo2</line>
+            <line>// comment</line>
+            <line>(?d).DS_Store</line>
+            <line>*2</line>
+            <line>qu*</line>
+        </ignores>
     </defaults>
 
 The ``defaults`` element describes a template for newly added device and folder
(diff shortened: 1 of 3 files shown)


users/config.rst
```diff
@@ -1309,6 +1309,13 @@ Defaults Element
             <untrusted>false</untrusted>
             <remoteGUIPort>0</remoteGUIPort>
         </device>
+	<ignores>
+            <line>!foo2</line>
+            <line>// comment</line>
+            <line>(?d).DS_Store</line>
+            <line>*2</line>
+            <line>qu*</line>
+        </ignores>
```
imsodin, line 1318: An indentation step too much?

users/config.rst
```diff
@@ -1335,6 +1342,16 @@ be present in the ``defaults`` element:
...
 
+.. option:: defaults.ignores
+
+    .. versionadded:: 1.19.0
+
+    Template for the :ref:`ignore patterns <ignoring-files>` applied to new
+    folders.  These are automatically copied to the ``.stignore`` file when a
+    folder is automatically accepted from a remote device.  The GUI uses them to
+    pre-fill the respective field when adding a new folder as well.  In XML,
+    each pattern line is represented as by a ``<line>`` element.
```
imsodin, line 1353: Maybe those two scenarios (gui and auto-accept) are all there are, however I'd suggest just extending your first sentence to have a generic explanation which applies to everything (directly covers auto-accept, and the GUI behaviour is also just a direct consequence of this). 

```suggestion
    Template for the :ref:`ignore patterns <ignoring-files>`, which are applied to newly created folders (unless a `.stignore` file is already in place).
```

imsodin (commented):
Thanks for documenting this!

One formatting nit and one suggestion.

users/config.rst
acolomb: I don't think so, the rest of the code example is also indented with four spaces per level.  I did sneak another tab in there though, which is now fixed.

users/config.rst
acolomb: I do think the difference matters.  The auto-accept part is done in the backend, while suggesting the default ignore patterns happens in the GUI by actively reading the defaults and submitting them after folder creation.  Since this is also the reference documentation for whoever uses the config API, they need to know that the second part will not happen unless they implement it.

users/config.rst
```diff
@@ -1335,6 +1342,16 @@ be present in the ``defaults`` element:
     Even sharing with other remote devices can be done in the template by
     including the appropriate :opt:`folder.device` element underneath.
 
+.. option:: defaults.ignores
+
+    .. versionadded:: 1.19.0
+
+    Template for the :ref:`ignore patterns <ignoring-files>` applied to new
+    folders.  These are automatically copied to the ``.stignore`` file when a
+    folder is automatically accepted from a remote device.  The GUI uses them to
```
imsodin, line 1351: Optional nit:

```suggestion
    folders.  These are copied to the ``.stignore`` file when a
    folder is automatically accepted from a remote device.  The GUI uses them to
```

Seems repetitive. Same below.

imsodin (approved):

users/config.rst
acolomb: Thanks, fixed.

PR #753, opened 2022-04-19
users/config: Add alias for the defaults.ignores.lines element

acolomb:
Ref https://github.com/syncthing/syncthing/pull/8265

When systematically generating the links, this will be the name coming from the config's JSON serialization format.


imsodin (approved):
Reiterating to check my understanding: This alias is needed such that the help link in that PR lead to this section.

acolomb:
Yes, because we didn't bother to write an own subsection for the `line` elements.

PR #756, opened 2022-05-20
users/ignoring: Add "synced" prefix to folder

DennisGaida:
I was totally confused on where I was supposed to put the `.stignore` file and treated it like a `.gitignore` file that can be placed in any folder-level-hierarchy and lower hierarchies overwrite higher hierarchies. I was not aware that I need to put the `.stignore` file into the **synced folder's root**.

Three things would have helped me understand better:
* Backlinks: Actually linking the "getting started" page that would help me understand concepts. There are rarely links in between docs articles
* Using a custom term for a "Syncthing synced folder" (just like we have a "Dropbox folder" or a "OneDrive folder") - I know it can be multiple folders that are synced, it is still a special thing. The documentation and everything else always just talks about "folders" (and sometimes "directories" interchangeably, maybe also something to be improved?) - a Syncthing folder is not just any folder in the system, it is treated differently hence makes it a "Syncthing folder" or "synced folder" or whatever you want to call it
* Have the `.stignore` file included in the directory structure so I would see where it goes

### DIFF
--- a/users/ignoring.rst
+++ b/users/ignoring.rst
@@ -17,10 +17,10 @@ Description
 
 If some files should not be synchronized to (or from) other devices, a file called
 ``.stignore`` can be created containing file patterns to ignore. The
-``.stignore`` file must be placed in the root of the folder. The
+``.stignore`` file must be placed in the root of the synced folder (see :ref:`Folder concepts <getting-started>`). The
 ``.stignore`` file itself will never be synced to other devices, although it can
 ``#include`` files that *are* synchronized between devices. All patterns are
-relative to the folder root.
+relative to the synced folder root.
 The contents of the ``.stignore`` file must be UTF-8 encoded.
 
 .. note::
@@ -70,14 +70,14 @@ The ``.stignore`` file contains a list of file or path patterns. The
    you can type ``?banana?`` to match both ``[banana]`` and
    ``{banana}``, and so on.
 
--  A pattern beginning with ``/`` matches in the root of the folder only.
+-  A pattern beginning with ``/`` matches in the root of the synced folder only.
    ``/foo`` matches ``foo`` but not ``subdir/foo``.
 
 -  A pattern beginning with ``#include`` results in loading patterns
    from the named file. It is an error for a file to not exist or be
    included more than once. Note that while this can be used to include
    patterns from a file in a subdirectory, the patterns themselves are
-   still relative to the folder *root*. Example:
+   still relative to the synced folder *root*. Example:
    ``#include more-patterns.txt``.
 
 -  A pattern beginning with a ``!`` prefix negates the pattern: matching files
@@ -115,9 +115,10 @@ The ``.stignore`` file contains a list of file or path patterns. The
 Example
 -------
 
-Given a directory layout::
+Given a synced folder directory layout::
 
     .DS_Store
+    .stignore
     foo
     foofoo
     bar/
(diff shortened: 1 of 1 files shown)


calmh:
I agree the terminology can be clearer, but we generally mean "Syncthing folder" by "folder", and use "directory" for meaning just any directory on disk. Hence the "at the root of the folder" in the thing you're editing.

tomasz1986:
Not a fan of "synced folder" personally, as I can very well edit `.stignore` before even sharingthe folder, i.e. in the "unshared" state. There's nothing "synced" about it at that point yet. "Syncthing folder" is indeed very clear and specific, but changing all instances of "folder" to "Syncthing folder" would be very repetitive…

How about adding a short glossary to the Docs welcome screen that would very briefly explain the main concepts (e.g. folder, device, directory, etc.).

users/ignoring.rst
```diff
@@ -115,9 +115,10 @@ The ``.stignore`` file contains a list of file or path patterns. The
 Example
 -------
 
-Given a directory layout::
+Given a synced folder directory layout::
```
imsodin, line 118: This isn't correct, what's displayed is a "directory layout", it's the filesystem objects not any Syncthing concept. If you want to be explicit about the folder root part, that could be the following (though that feels overly verbous):

```suggestion
Given a directory layout starting at the synced folder's root::
```

imsodin (commented):
I like the change.  
A glossary doesn't help - the misunderstanding is not happening for people that know Syncthing or research in depth, it's for the average user. And they won't read the glossary and/or get the significance of directory vs folder in Syncthing context. So giving a clue with "synced" to make it more clear is a good thing imo. It doesn't have to be everywhere, once the context is established that should help already.

And being able to edit a `.stignore` outside of a synced folder or before syncing is seems like a very far fetched issue: It's still intended to end up in a synced folder.

st-review:
:robot: *beep boop*

AudriusButkevicius:
I think there is an outstanding point that needs to be addressed before this can be merged.

Do you plan on addressing it?

users/ignoring.rst
```diff
@@ -17,10 +17,10 @@ Description
 
 If some files should not be synchronized to (or from) other devices, a file called
 ``.stignore`` can be created containing file patterns to ignore. The
-``.stignore`` file must be placed in the root of the folder. The
+``.stignore`` file must be placed in the root of the synced folder (see :ref:`Folder concepts <getting-started>`). The
```
imsodin, line 20: Following that link to the getting started guide I don't find what "Folder concepts" refers to.

imsodin (commented):
The bot is helpful mainly because it draws attention to forgotten PR like here (I did forget it too). The cleanup aspect is also nice, but secondary to me.

Unfortunately I found one more thing I don't understand, see comment below.

users/ignoring.rst
imsodin: Am I missing something - the only bit about folders there I find is a description of the UI:

> On the left is the list of “folders”, or directories to synchronize. You can see the Default Folder was created for you, and it’s currently marked “Unshared” since it’s not yet shared with any other device. On the right is the list of devices. Currently there is only one device: the computer you are running this on.

I don't see how this is about folder concepts or helpful in the context of understanding putting a .stignore into the synced folder root.

imsodin: Yup, removing sounds good. And a concepts page would be nice too, respectively maybe add it to (Understanding Synchronisation](https://docs.syncthing.net/users/syncing.html), that's close to concepts I'd say.

imsodin (approved):

PR #758, opened 2022-06-08
Update running on port 443 as unprivileged user

JsBergbau:
Removed the filter `-i eth0' since predictable network interface names like ens2 are nowadays common. This makes it easier for most of the users setting up a working relay server on Port 443, since copy and paste now works. More advanced users with multiple NICs should be familiar with iptables more and can add the NIC filter on their own.

### DIFF
--- a/users/strelaysrv.rst
+++ b/users/strelaysrv.rst
@@ -168,7 +168,7 @@ there are a couple of approaches available to you.
 One option is to run the relay on port 22067, and use an ``iptables`` rule
 to forward traffic from port 443 to port 22067, for example::
 
-    iptables -t nat -A PREROUTING -i eth0 -p tcp --dport 443 -j REDIRECT --to-port 22067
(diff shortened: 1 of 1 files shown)


JsBergbau:
Can you describe your concerns more?
Without interface specification traffic of all interfaces and port 443 is redirected to synchting port. Having multiple WAN interfaces needs policy routing / multiple routing tables to have a working connections, otherwise packets entering from eth0 may leave the system on eth1 which will break the connection because of wrong source address at the receiver. More details see here https://tldp.org/HOWTO/Adv-Routing-HOWTO/lartc.rpdb.multiple-links.html

So from a technical point of view I have no concerns ommitting `-i eth0`. I'm using the proposed iptables setting and I have to admit that I had copied and pasted the proposed setting. This didn't work and even as advanced iptable user it took me some time to figure out what was going on.

calmh:
No, this is probably fine. My concern was that it would match locally generated outgoing packets, but the prerouting chain isn't used there.

PR #759, opened 2022-07-05
advanced: Sync ownership

calmh:


advanced/folder-sync-ownership.rst
```diff
@@ -0,0 +1,64 @@
...
+The two are not mixed -- that is, ownership information is not synchronised
+between POSIX and Windows systems, only POSIX-to-POSIX and
+Windows-to-Windows. Nonetheless Syncthing attempts to leave unknown
+ownership data intact: POSIX systems will not overwrite Windows ownership
+data but pass it on unmodified, and vice versa.
+
+POSIX implementation
+--------------------
+
+Syncthing records both the numberical UID and GID for a file and the
```
acolomb, line 19:
```suggestion
Syncthing records both the numerical UID and GID for a file and the
```

advanced/folder-sync-ownership.rst
```diff
@@ -0,0 +1,64 @@
...
+Elevated permissions
+~~~~~~~~~~~~~~~~~~~~
+
+Syncthing, when running as a normal user account, doesn't have permission to
+alter file ownership. There are several reasonable ways of running Syncthing
+with elevated permissions to enable ownership sync:
+
+- As root, in a Docker container, with the synced data mounted as a
+  volume.
+- As a normal user, with extra capabilities granted the executable.
```
acolomb, line 34:
```suggestion
- As a normal user, with extra capabilities granted to the executable.
```

acolomb (changes requested):
Small typos.  Excited to see how the rest of it turns out.  Unfortunately I can't help with Windows stuff much, as I tend to avoid that OS when possible.

acolomb:
The "Elevated Permissions" headline is duplicated, please remove at the end.

Would you mind integrating the option in `users/config.rst` as well, in the examples and as an `.. option::` definition?  That's needed to make the GUI links work.  Just linking to the new page should be sufficient.

calmh:
The duplication was intended, in that it is different in POSIX (which I'll rename to Unix) and Windows... But since there's currently nothing I can say about how this works in Windows I'll just skip that part.

PR #760, opened 2022-08-12
advanced: Document scanOwnership

calmh:
syncthing/syncthing#8497

### DIFF
--- a/advanced/folder-scan-ownership.rst
+++ b/advanced/folder-scan-ownership.rst
@@ -1,0 +1,12 @@
+.. _folder-scan-ownership:
+
+scanOwnership
+=============
+
+.. versionadded:: 1.21.0
+
+When enabled, Syncthing will record ownership for local files when they are
+scanned, and send this information to peer devices. Peer devices configured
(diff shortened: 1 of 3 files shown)


acolomb (changes requested):
If you want to reference another document, there is a separate `:doc:...` syntax for it. So you don't need to clutter the global namespace of reference labels, but just give the path of the document. That's also easier to find when looking at the unrendered document.

advanced/folder-scan-ownership.rst
```diff
@@ -0,0 +1,12 @@
+.. _folder-scan-ownership:
+
+scanOwnership
+=============
+
+.. versionadded:: 1.21.0
+
+When enabled, Syncthing will record ownership for local files when they are
+scanned, and send this information to peer devices. Peer devices configured
```
tomasz1986, line 9:
```suggestion
scanned and send this information to peer devices. Peer devices configured
```

tomasz1986:
Not sure if suitable for this PR, but what do you think about adding a note/warning in the Windows subsection that syncing ownership has a significant performance impact/drawback when used in the OS?

acolomb (approved):

PR #762, opened 2022-09-02
Update IP in reverse proxy page

stfnx:
Using `127.0.0.1` instead of `localhost` fixed long 502 errors on nginx for me.
On systems where both IPv4 and IPv6 is enabled, this seems cause issues.
See [here](https://stackoverflow.com/a/72188777) for an explanation.

### DIFF
--- a/users/reverseproxy.rst
+++ b/users/reverseproxy.rst
@@ -19,7 +19,7 @@ Server Configuration
 
 If you have access to your web server's configuration use the following
 examples to pass the location ``/syncthing`` on your web server to Syncthing's
-GUI hosted on ``localhost:8384``.
+GUI hosted on ``127.0.0.1:8384``.
 
 Apache
 ~~~~~~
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
It shouldn't matter, that's something wrong woth your setup, if localhost somehow does not resolve to 127.0.0.1.

stfnx:
`localhost` does resolve to `127.0.0.1` as it should, but on systems with both IPv4 and IPv6 it also resolves to `::1`. And for some reason this results in a 502 error that is way longer than it should be.

While restarting Syncthing it is not available, so I get a 502, which is normal behaviour. But even when the restarting process finished (you can tell by watching logs), the 502 error is still there for quiet some time. When using 127.0.0.1 instead of localhost, the 502 goes away instantly when Syncthing has restartet and is available again.

AudriusButkevicius:
That still sounds like some sort of misconfiguration.

PR #766, opened 2022-09-14
advanced: scanOwnership became sendOwnership

calmh:


calmh:
Indeed, and with the current docs versioning stuff it's even preserved for instances actually running 1.21.0 out there.

PR #770, opened 2022-10-07
dev/rest: added documentation for noauth

er-pa:
Added documentation to support recently made changes to the REST API (Syncthing PR: #8585).

- Added a short explanatory remark where the requirement of having a set API KEY is being explained to clarify the exception-case which 'noauth' is;
- Then I added a page with a short general explanation the purpose of /rest/noauth;
- And I included the actual endpoint "health", explaining what it returns.

I noticed that there are two ways to implement this, but since a slight general explanation was (imo) preferred I chose a path similar to  /rest/config and /rest/debug.

Also I noticed an inconsistency in the identation-level in code blocks regarding json, so I wasn't entirely sure which one would be preferred.

### DIFF
--- a/dev/rest.rst
+++ b/dev/rest.rst
@@ -116,7 +116,7 @@ Debug Endpoints
 Noauth Endpoints
 ----------------
 
-Calls that do not require authentication
+Calls that do not require authentication.
 
 .. toctree::
    :maxdepth: 1
--- a/dev/rest.rst
+++ b/dev/rest.rst
@@ -120,6 +120,5 @@ Calls that do not require authentication
 
 .. toctree::
    :maxdepth: 1
-   :glob:
 
    /rest/noauth/... <../rest/noauth.rst>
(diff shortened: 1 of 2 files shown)


acolomb (changes requested):
> I noticed that there are two ways to implement this, but since a slight general explanation was (imo) preferred I chose a path similar to /rest/config and /rest/debug.

I'd prefer the other way (single file per endpoint) in this case.  The `/rest/config/...` endpoints are all strongly related and I understand the debugging stuff as being a little less "official", e.g. subject to change and not to be used as a stable interface.  `/rest/noauth/...` on the other hand is a property for arbitrary endpoints that we may add later, which will probably have nothing to do with the health check.  So collecting them on a single page just because of that common property seems wrong.  You do have the generic text in the authentication section `dev/rest.rst` and a short description under the headline.  Please merge the introductory text into those two places.

> Also I noticed an inconsistency in the identation-level in code blocks regarding json, so I wasn't entirely sure which one would be preferred.

It's not perfectly consistent, right. With your two-spaces indentation, I think it matches the majority of JSON code blocks though.

acolomb (approved):
Thanks for the quick reaction :slightly_smiling_face:

PR #771, opened 2022-10-08
users/syncing: Add note about Syncthing namespace

tomasz1986:
Add a note about how files starting with any of the two prefixes `.syncthing.` and `~syncthing~` are ignored and not synchronised.

### DIFF
--- a/users/syncing.rst
+++ b/users/syncing.rst
@@ -149,3 +149,11 @@ The temporary files are named ``.syncthing.original-filename.ext.tmp`` or,
 on Windows, ``~syncthing~original-filename.ext.tmp`` where
 ``original-filename.ext`` is the destination filename. The temporary file is
 normally hidden. If the temporary file name would be too long due to the addition of the prefix and extra extension, the temporary files are named ``.syncthing.<hash>.tmp`` or, on Windows, ``~syncthing~<hash>.tmp`` where ``<hash>`` is a SHA-256 hash of the original filename.
+
+.. note::
+
+    Note that the two prefixes ``.syncthing.`` and ``~syncthing~`` are
+    considered Syncthing namespace, meaning that any files whose names
+    start with them will automatically be ignored and excluded from
+    synchronisation by Syncthing. Please avoid using these prefixes in
+    your filenames if you want those files to be synchronized.
--- a/users/syncing.rst
+++ b/users/syncing.rst
@@ -156,4 +156,4 @@ normally hidden. If the temporary file name would be too long due to the additio
     considered Syncthing namespace, meaning that any files whose names
     start with them will automatically be ignored and excluded from
     synchronisation by Syncthing. Please avoid using these prefixes in
-    your filenames if you want those files to be synchronized.
+    your filenames.
--- a/users/faq-parts/general.rst
+++ b/users/faq-parts/general.rst
@@ -47,6 +47,8 @@ The following are *not* synchronized;
 -  Windows ACLs (not preserved)
 -  Devices, FIFOs, and other specials (ignored)
 -  Sparse file sparseness (will become sparse, when supported by the OS & filesystem)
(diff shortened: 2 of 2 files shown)


tomasz1986:
For reference: https://github.com/syncthing/syncthing/issues/8595

users/syncing.rst
```diff
@@ -149,3 +149,11 @@ The temporary files are named ``.syncthing.original-filename.ext.tmp`` or,
...
 ``original-filename.ext`` is the destination filename. The temporary file is
 normally hidden. If the temporary file name would be too long due to the addition of the prefix and extra extension, the temporary files are named ``.syncthing.<hash>.tmp`` or, on Windows, ``~syncthing~<hash>.tmp`` where ``<hash>`` is a SHA-256 hash of the original filename.
+
+.. note::
+
+    Note that the two prefixes ``.syncthing.`` and ``~syncthing~`` are
+    considered Syncthing namespace, meaning that any files whose names
+    start with them will automatically be ignored and excluded from
+    synchronisation by Syncthing. Please avoid using these prefixes in
+    your filenames if you want those files to be synchronized.
```
calmh, line 159:
```suggestion
    your filenames.
```

calmh (commented):
Slightly stronger recommendation.

tomasz1986:
I'd just like to add a short explanation why I even named the script `.syncthing.ps1`. The reason is that I sync several so-called "global ignore" files (e.g. `.stignoreglobal`, etc.), so that I can edit them in one place and use with `#include` on multiple devices. Recently, I've also also been experimenting with a custom versions folder that I wanted to synchronise across multiple devices as well.

Because of the above, the folder root became quite cluttered, so I decided to create a folder called `.syncthing` and move all my custom ignore files and versions folder there. This `.syncthing` folder is synchronised with no issues. However, even with this approach, I still need to update the ignore files in each Syncthing folder separately, and I've got many of them, so I use a script to do it in a single run.

It was logical for me to name the script `.syncthing.ps1` after the `.syncthing` folder 🙂. Only then I found out that it was a reserved filename.

acolomb:
Maybe add a quick mention here as well:

https://docs.syncthing.net/users/faq.html#what-things-are-synced

and / or

https://docs.syncthing.net/users/ignoring.html

calmh:
Just adding a note here,

> This .syncthing folder is synchronised with no issues.

You're on dangerous ground as this has been proposed as the replacement .stfolder several times. If and when that ever happens your thing will break and/or be blown away. Don't camp there. :D

tomasz1986:
Yeah, I'm aware of that, but I want to the folder to be identifiable as something related to Syncthing (and not my name or anything like that), and I couldn't really come up of anything better or more suitable 🙁. This should probably go on the forum, but I'm open to any propositions on how to name it, so that a) it's obvious that it's Syncthing-related, and b) it doesn't conflict with any current or future Syncthing namespaces.

tomasz1986:
> Maybe add a quick mention here as well:
> 
> [docs.syncthing.net/users/faq.html#what-things-are-synced](https://docs.syncthing.net/users/faq.html#what-things-are-synced)
> 
> and / or
> 
> [docs.syncthing.net/users/ignoring.html](https://docs.syncthing.net/users/ignoring.html)

I've updated the FAQ. Is it necessary to have this in the latter as well? I don't think it really helps the user in any meaningful way, as they can't do anything with those files anyway (i.e. they're not added to the folder at all, and the user can't force Syncthing to sync them with ignore patterns).

acolomb:
Yeah the idea was just that it's clear there are some ignored names even when the ignore patterns are empty. Feel free to ignore it if you don't see a good place on that page.

st-review:
:robot: beep boop

tomasz1986:
Was there anything else that I was supposed to change in the PR?

calmh (approved):

PR #772, opened 2022-10-10
Improve clarity: exclusively from rather than of

chmac:
I read this in the docs:

> Syncthing can be configured to use specific relay servers **(exclusively of the public pool)** by adding the required servers to the Sync Protocol Listen Address field, under Actions and Settings.

I'm not sure what "exclusively of the public pool" means in this context. Does it mean that in addition to the public pool relay servers any additional servers can be specified? Or does it mean that only servers from the public pool can be used? As in, this option supports servers exclusively sourced from the public pool.

I chose to interpret the meaning as the latter. I thus created a PR which changes the word "of" to "from". I think this greatly improves the clarity of the sentence. However, I stress again, **I don't know which option is actually correct**. So I hope somebody who knows the correct answer can review this PR to ensure that it doesn't end up making the docs less accurate.

### DIFF
--- a/users/relaying.rst
+++ b/users/relaying.rst
@@ -19,7 +19,7 @@ communicating via the relay.
 Configuring clients
 -------------------
 
-Syncthing can be configured to use specific relay servers (exclusively of the public pool) by adding the required servers to the Sync Protocol Listen Address field, under Actions and Settings. The format is as follows::
+Syncthing can be configured to use specific relay servers (exclusively from the public pool) by adding the required servers to the Sync Protocol Listen Address field, under Actions and Settings. The format is as follows::
 
   relay://<host name|IP>[:port]/?id=<relay device ID>
 
--- a/users/strelaysrv.rst
+++ b/users/strelaysrv.rst
@@ -146,7 +146,7 @@ procedure for your operating system.
 Client configuration
 ~~~~~~~~~~~~~~~~~~~~
 
-Syncthing can be configured to use specific relay servers (exclusively of the public pool) by adding the required servers to the Sync Protocol Listen Address field, under Actions and Settings. The format is as follows::
+Syncthing can be configured to use specific relay servers (exclusively from the public pool) by adding the required servers to the Sync Protocol Listen Address field, under Actions and Settings. The format is as follows::
 
   relay://<host name|IP>[:port]/?id=<relay device ID>
 


chmac:
Hmm, on second thought, I think I got the wrong meaning of the 2. I'm still not sure, but I now begin to suspect that I don't need to specify an additional relay, that syncthing will automatically use relays. So I suspect this PR should be closed in favour of the PR I'll create next which makes the text clearer in the opposite direction...

imsodin:
I think that can just be removed entirely: You can add a relay as described in addition to pool relays (enabled by default) or instead. The relay you add may be in the pool or not, that doesn't matter (though both having the pool and a relay statically that's also in the pool is redundant).

chmac:
@imsodin Okay, thanks for clarifying. I think what you wrote here is much clearer than what I read in the docs so far.

Do I enable or disable all pool relays with the "enable relays" option in the GUI?

Do you know if both of my clients need to be connected to the same relay? Or will being connected to different relays work if global discovery is also enabled?

Are you open to a new PR that adds this info in a bid to make it clearer?

imsodin:
The enable relay just means whether syncthing establish relay connections at all, regardless of which. I.e. it's independent.

No need for connecting to the same relay - you always need to discover the other device, even if on the same relay.

> Are you open to a new PR that adds this info in a bid to make it clearer?

Sure, though no need for a new PR, just edit this one. If you are using GH directly to edit, go to the files changed tab, clicke the three dots on the top right of the file you want to edit, and choose "edit file".

chmac:
@imsodin Sorry about the delay, got caught up in some family stuff. I've reworded it to say that these servers are added in addition to public pool servers. I'm not sure how the public pool servers are disabled, and given the time since I opened the PR, I thought this was the safest change. I believe this to be correct in accordance with your comments above.

Thanks for working on syncthing btw, it's an awesome tool and I greatly appreciate it.

imsodin:
I still think the easiest would be to just drop it. The current wording doesn't seem correct to me, as to me it contradicts that one can also add a relay instead of the default pool. So yeah, dropping or maybe `in addition to or to replace the public pool` - but that's a mouthful and really what one would expect anyway, isn't it?

chmac:
@imsodin What I couldn't wrap my head around is how to replace the pool. Is that in a separate option?

imsodin:
By keeping/adding `dynamic+https://relays.syncthing.net/endpoint` in the listen address or not:  
https://docs.syncthing.net/users/config.html#listen-addresses

st-review:
:robot: beep boop

PR #775, opened 2022-11-27
rest: Add isLocal flag to GET /rest/system/connections.

acolomb:
Ref https://github.com/syncthing/syncthing/pull/8694.


calmh (approved):

acolomb:
@calmh Are you going to make another v1.22.2-rc.4 with this change? If not, we should wait before merging this to the docs.

calmh:
Not for 1.22.2, no, so yes better to hold off a few days

acolomb:
Oh darn, I just noticed that v1.22.2 has not actually been released yet.  Just saw the tag on the main repo already and assumed it was finished.  We've somehow got a delay in our last release cycle, no?

I'll just revert the commit and push it back when ready.

PR #776, opened 2022-12-02
Update usage.rst

leunga1000:


acolomb (changes requested):
This is factually wrong. When Syncthing is stopped, as recommended, no `syncthing cli` command can really work. And it requires access to the API key, which the previously recommended command doesn't (that's fine however, assuming both variants need direct access to the config file).

users/faq-parts/usage.rst
```diff
@@ -269,13 +269,11 @@ How do I reset the GUI password?
...
-:option:`--gui-password` (and possibly :option:`--gui-user`) options to the
-``syncthing generate`` subcommand.  This should be done while Syncthing is not
-running.
+:option:`syncthing cli config gui password` (and possibly :option:`syncthing cli config gui user`) subcommands.  
+
+1. ``syncthing cli config gui user set newUserName``
+2. ``syncthing cli config gui password set myNewPassword``
 
-1. Stop Syncthing: ``syncthing cli operations shutdown``
-2. ``syncthing generate --gui-password=myNewPassword --gui-user=newUserName``
```
acolomb, line 277: The "step 2" referenced below is now something different.  In fact, the whole following paragraph doesn't make much sense if you only document the CLI commands.

acolomb (changes requested):
I'm not fond of dropping the previous approach completely.  Listing the config CLI commands is fine, but it assumes that Syncthing is still running and the API can be accessed.  Better add this variant as an alternative, but keep the existing methods as they were.

Please make sure to blend in with the surrounding text regarding both style and formatting (esp. line wrapping).

leunga1000:
> I'm not fond of dropping the previous approach completely. Listing the config CLI commands is fine, but it assumes that Syncthing is still running and the API can be accessed. Better add this variant as an alternative, but keep the existing methods as they were.
> 
> Please make sure to blend in with the surrounding text regarding both style and formatting (esp. line wrapping).

I think there's something fundamentally I'm missing here, `syncthing generate` returns an error for me.

acolomb:
> I think there's something fundamentally I'm missing here, `syncthing generate` returns an error for me.

Well what is the error?  The documented procedure usually does work, so if you're just having trouble using it, let's diagnose why.

PR #778, opened 2022-12-24
Elaborate that double slashes do not introduce comments everywhere

ooker777:
see https://forum.syncthing.net/t/would-using-double-slashes-in-the-middle-of-a-line-have-the-effect-of-commenting/19227

### DIFF
--- a/users/ignoring.rst
+++ b/users/ignoring.rst
@@ -94,7 +94,7 @@ The ``.stignore`` file contains a list of file or path patterns. The
    they are preventing directory deletion. This prefix should be used by any OS
    generated files which you are happy to be removed.
 
--  A line beginning with ``//`` is a comment and has no effect.
+-  A line beginning with ``//`` is a comment and has no effect. Putting it in the middle of the line (e.g. `file // comment`) will have it literally ignore a file named `file // comment`, which is a valid path in Unix
 
 .. note::
 
--- a/users/ignoring.rst
+++ b/users/ignoring.rst
@@ -94,7 +94,9 @@ The ``.stignore`` file contains a list of file or path patterns. The
    they are preventing directory deletion. This prefix should be used by any OS
    generated files which you are happy to be removed.
 
--  A line beginning with ``//`` is a comment and has no effect. The same double slashes in any other place are interpreted literally, e.g. trying to do ``file // comment`` will make Syncthing look for a file called ``file // comment``.
+-  A line beginning with ``//`` is a comment and has no effect. The same double 
+   slashes in any other place are interpreted literally, e.g. trying to do
+   ``file // comment`` will make Syncthing look for a file called ``file // comment``.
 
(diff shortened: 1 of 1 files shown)


calmh:
I mean, it already says "a line beginning with", I think that's quite clear?

tomasz1986:
Just for the record, `/` is the one character that is **not** valid in probably any OS (see https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names), so even if accepted, I'd suggest to at least remove the ", which is a valid path in Unix" part.

rasa:
> Just for the record, `/` is the one character that is **not** valid in probably any OS

There are [several](https://en.m.wikipedia.org/wiki/Comparison_of_file_systems#Limits) OSs that allow `/` in filenames, HFS+ being one of them. As a result, in iOS 16.2's Files app, you can create a filename with a slash, but not with a colon. See https://discussions.apple.com/thread/7540284 .

ooker777:
@calmh 
> it already says "a line beginning with", I think that's quite clear?
Well not to me I guess. IIRC, my discovery path was like this:
- I assumed commenting in the middle of the line works because most other programming languages support it
- It didn't work, and I checked the doc just to see if I was using the correct character. It was correct, and my mind just skipped the "a line beginning with". I saw it there, but didn't consider it carefully
- Tried several times with no luck. Coming back to the doc to read it again. This time I suspected that I had been assuming incorrectly. Hence the question in the forum.

So I think there are people who goes the same path as me, and I want to make it clearer. 

@tomasz1986 
>  I'd suggest to at least remove the ", which is a valid path in Unix" part.
Hmm, but paths with double slashes are still accepted. They are just interpreted as one slash: https://unix.stackexchange.com/a/11965/60602

I think if we want to keep it short, we can just say "not in the middle", then give the link. But I'm thinking that if people has taken time to read the doc, then there is no need for short. It's just better be right in the gui

calmh:
> Hmm, but paths with double slashes are still accepted. They are just interpreted as one slash

Not in our patterns though, we'll just try to match a path against double slashes and it'll never match.

ooker777:
so is this PR good? Or do you suggest any edit on it?

tomasz1986:
Maybe something like this?

```
A line beginning with ``//`` is a comment and has no effect. The same double
slashes in any other place are interpreted literally, e.g. trying to do ``file // comment``
will make Syncthing look for a file called ``file // comment``.
```

ooker777:
sure. I update it in here. Not sure if it works or not https://github.com/syncthing/docs/pull/778/files

acolomb:
Please do adhere to the 80-character line wrapping limit used in the surrounding text. It doesn't matter for the HTML output, but the ReST files should be easily human-readable in raw form.

acolomb (approved):

PR #782, opened 2023-01-12
Add table of historical versions

calmh:
This adds a table of every released Syncthing version, together with the corresponding date and Go version. Apart from some historical interest, this can be used to find the latest version built with a given Go toolchain.

Also a script and GitHub Action to auto-maintain this list in the future.


tomasz1986:
Very nice 🙂. There is already "Versions & Releases" (https://docs.syncthing.net/users/releases.html) in "Usage". What do you think about just adding the table there instead of adding a completely new item to the "Usage" list (which is already a little lengthy)?

PR #792, opened 2023-03-08
Clarify handling of symbolic & hard links

codesections:
This PR clarifies Syncthing's treatment symbolic and hard links.  This has been a source of (minor) confusion for at least a few years (see, e.g., this [2018 forum question](https://forum.syncthing.net/t/handling-of-symbolic-and-or-hard-links-on-linux/11933/8).  In the linked discussion, someone suggested submitting a PR to clarify the existing docs; this is that PR.

Thanks to all the Syncthing contributors for an amazing project!

### DIFF
--- a/users/faq-parts/general.rst
+++ b/users/faq-parts/general.rst
@@ -32,6 +32,8 @@ The following may be synchronized or not, depending:
 -  File permissions (when supported by file system; on Windows only the
    read only bit is synchronized)
 -  Symbolic links (synced, except on Windows, but never followed)
+   - Because a symbolic link is never followed, the synced link will
+     be broken unless the link's target is also synced
 -  File or directory owners and groups (when enabled)
 -  Extended attributes (when enabled)
 -  POSIX or NFS ACLs (as part of extended attributes)
@@ -40,6 +42,10 @@ The following are *not* synchronized;
 
 -  Directory modification times (not preserved)
 -  Hard links (followed, not preserved)
+   - Following the link means that the linked file *is* synced, just as if it
+     were a regular file at that path.  What isn't synced is the status as a 
+     link; for example, syncing two hardlinks to the same file will result in 
+     two independent copies of that file (without any hard links). 
 -  Windows junctions (synced as ordinary directories; require enabling in
    :stconf:opt:`the configuration <folder.junctionsAsDirs>` on a per-folder
    basis)
--- a/users/faq-parts/general.rst
+++ b/users/faq-parts/general.rst
@@ -32,8 +32,6 @@ The following may be synchronized or not, depending:
 -  File permissions (when supported by file system; on Windows only the
    read only bit is synchronized)
 -  Symbolic links (synced, except on Windows, but never followed)
-   - Because a symbolic link is never followed, the synced link will
-     be broken unless the link's target is also synced
 -  File or directory owners and groups (when enabled)
 -  Extended attributes (when enabled)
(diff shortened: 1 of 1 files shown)


users/faq-parts/general.rst
```diff
@@ -32,6 +32,8 @@ The following may be synchronized or not, depending:
 -  File permissions (when supported by file system; on Windows only the
    read only bit is synchronized)
 -  Symbolic links (synced, except on Windows, but never followed)
+   - Because a symbolic link is never followed, the synced link will
```
calmh, line 35: This is a bit imprecise. The target doesn't need to be synced, it just need to exist, like for any other symlink.

users/faq-parts/general.rst
```diff
@@ -40,6 +42,10 @@ The following are *not* synchronized;
 
 -  Directory modification times (not preserved)
 -  Hard links (followed, not preserved)
+   - Following the link means that the linked file *is* synced, just as if it
```
calmh, line 45: This seems a bit wordy for the middle of an otherwise concise bullet list. Is there really confusion about what not preserving hard links means, that doesn't stem from not understanding hard links to begin with?

calmh (commented):
Maybe a lengthier discussion on links fits better after the list itself?

users/faq-parts/general.rst
calmh: Maybe something like `Hard links (files synced as usual, hard links not preserved)` would cover it?

codesections:
Thanks for the feedback, @calmh.  I agree that a large percentage of the confusion stems from "not understanding hard links to begin with" (which I'm willing to bet is pretty common even in fairly tech-savy users – hard links aren't used all that often).  And I also agree that explaining that in the middle of a bulleted list makes the list a bit too wordy.

To address those points, I broke the explanation of links into a separate FAQ entry and added an example; hopefully that makes things more approachable to users who aren't as clear on [sym|hard]links without cluttering the bulleted list.

FYI, I don't expect to be able to devote much more time to this PR, so anyone should feel free to make any edits/changes directly.  Thanks again!

users/faq-parts/general.rst
```diff
@@ -48,6 +48,37 @@ The following are *not* synchronized;
...
+directory will replicate the contents of ``../some-file.txt`` to
+`./Sync/hardlink` on every device.  But it will *not* sync the fact that
+``./Sync/hardlink`` was a link on the original device – on other devices,
+``./Sync/hardlink`` will be completely independent from ``../some-file.txt`` (if
+that file exists).
+
+All of this results in a general rule of thumb: if Syncthing is syncing both a
+link and its target, it's better to use a symbolic link.  However, if the link
+points to a file that is *not* in a synced directory and you want to sync the
+contents of that file, you should use a hard link instead.
```
acolomb, line 80: I don't think it's appropriate to give advice on what type of links a user should create, from the perspective of the file synchronization software.  Symlinks and hardlinks are fundamentally different concepts despite the similar names.  Symlinks work on a path level, while hardlinks work on the inodes (basically).

We should just state that Syncthing does not respect the fact that two or more files have a hardlink relationship, it is not preserved when syncing to other devices.

The hard-link example here talks about the situation where two links (file names) point to the same file, one inside and one outside the synced folder.  However the explanation is just as relevant for the case when both are inside, since this obviously will double the space requirement on receiving devices.

st-review:
:robot: beep boop

PR #798, opened 2023-04-29
users: Troubleshooting for "folder marker missing" error messages

GermanCoding:
Primary motivations are many forum threads over the years such as this one: https://forum.syncthing.net/t/the-problem-with-forum-support/20154

Also an old discussion: https://github.com/syncthing/syncthing/issues/5207

Once/If this PR is merged, I would also like to link it from within the error message in syncthing, so that it can be found immediatly.

### DIFF
--- a/users/faq-parts/troubleshooting.rst
+++ b/users/faq-parts/troubleshooting.rst
@@ -157,3 +157,38 @@ are run by volunteers all over the world. They usually listen on ports 443 or
 address you are concerned about with `the current list of active relays
 <https://relays.syncthing.net>`__. Relays do not and can not see the data
 transmitted via them.
+
+I am seeing the error message "folder marker missing". What do I do?
+--------------------------------------------------------------------
+
+Syncthing uses a specific marker usually called ``.stfolder`` to determine whether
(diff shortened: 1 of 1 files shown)


calmh:
Great. This might even deserve its own page, and we can iterate, but this a very good start.

GermanCoding:
So how do we want to go about this? I would like to link to the troubleshooting guide from within the error message. This currently says:

`folder marker missing (this indicates potential data loss, search docs/forum to get information about how to proceed)`

This message could be made more clear by changing it to include some sort of link to the docs guide. However, if we want to move this to a separate page we should do so first. Otherwise maintaining link because otherwise maintaining link availability annoying.

I can look into merging the current folder FAQ (which looks a bit split currently) into its own page, if that's what we want.

calmh:
This is from memory as I'm not at a real computer, but I think we're have the capability to handle arbitrary redirects on the doc site. So we could point towards something generic like /users/folder-marker-missing and handle that with a redirect to the FAQ entry or in the future to a separate page. (Essentially guessing the URL to the future page today.)

PR #801, opened 2023-05-29
Update autostart.rst

logical-view:
[How to set up a system service - Linux - Starting Syncthing Automatically](https://docs.syncthing.net/users/autostart.html#how-to-set-up-a-system-service)


imsodin (changes requested):
Why this change?

The `@.service` notation allows using any value for user/instance. As long as you don't hardcode anything specific to the user in the init file, the user name shouldn't be in the filename. And if you want to do that, I think that should be a separate drop-in file. Anyway that's digressing into systemd config specifics that's not really relevant (or needed) here.

logical-view:
> Why this change?

When using Debian package (syncthing/stable _armhf_),

```
Failed to start syncthing@.service: Unit name syncthing@.service is missing the instance name.
See system logs and 'systemctl status syncthing@.service' for details.
```

Rename the service file, `sudo mv /usr/lib/systemd/system/syncthing@.service /usr/lib/systemd/system/syncthing@pi.service` ... and success, it works okay.

acolomb (changes requested):
As far as I understood, the instance name is parsed dynamically from the unit specified in systemctl commands.  It doesn't have to be in the .service file name, as that is specifically installed as a **template**.

In my experience, `systemctl enable --now syncthing@pi.service` works out of the box if the DEB package is installed (the one from apt.syncthing.net, dunno about the Raspbian / Debian version if any).  So this change is really no good advice and should not be applied.

PR #803, opened 2023-06-01
advanced: Number of connections

calmh:
For syncthing/syncthing#8918. (We should add a link in the GUI as well, since this setting is not only available through the advanced interface.)

### DIFF
--- a/advanced/device-numconnections.rst
+++ b/advanced/device-numconnections.rst
@@ -1,0 +1,64 @@
+numConnections
+==============
+
+.. versionadded:: 1.25.0
+
+``numConnections`` is a device setting that affects connection handling. A
+zero value means to use the Syncthing default. As of version 1.25.0 the
+default is to use one connection, like earlier versions of Syncthing. This
+may change in the future.
+
+Multiple connections will be maintained to the device if you set this to a
+value greater than one. Multiple connections can yield improved performance
+by load-balancing traffic over multiple physical links or in other
+scenarios.
+
+A simple form of negotiation is used to decide how many connections to use
+between a device pair. It goes like this:
+
+- If either side is configured to use a single connection, then a single
+  connection is used. Since the default is to use a single connection this
+  means that to use more than one connection both sides must be configured
+  to do so.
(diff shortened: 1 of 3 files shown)


advanced/device-multipleconnections.rst
```diff
@@ -0,0 +1,55 @@
...
+of the number of connections. The limit is not per connection.
+
+Connection Types
+----------------
+
+Both TCP and QUIC connections are supported for multiple connections.
+Syncthing will, however, only keep connections with the best priority; by
+default, TCP has better priority than QUIC, so establishing a TCP connection
+will cause existing QUIC connections to be closed. Connection priorities can
+be configured.
```
AudriusButkevicius, line 53: If its the same priority, can you have mixed transports? If no, why not. If yes, how is that represented in the UI?

calmh: You can indeed, and it isn't really represented in the GUI at all. Currently it shows the info about the primary (oldest) connection, and a count of additional connections, but no details on the additionals. It's available in the REST API though, so it could be represented if we wanted to.

advanced/device-multipleconnections.rst
```diff
@@ -0,0 +1,55 @@
+.. _multiple-connections:
+
+multipleConnections
+===================
+
+.. versionadded:: TBD
+
+``multipleConnections`` is an advanced device setting that affects
+connection handling. The default is zero, equivalent to *one*: only one
```
bt90, line 9: Is that still the case?

calmh: No, this will need updating obviously.

tomasz1986:
This may a material for a different PR, but is there something to add to https://docs.syncthing.net/users/tuning regarding the connection number? I'm asking because the commit message specifically mentions possible tweaks for low-end hardware already (see https://github.com/syncthing/syncthing/commit/c6334e61aa56575db6b274d45d55f7ada4dcd0fe).

calmh:
I can add some words

PR #806, opened 2023-06-12
Match description field names to definition

mseravalli:
In the description, the field names should match the names defined in the proto definition for better discoverability and less mental load.

When trying to reading and understand the protocol it should be possible to look for the description of a field by searching the name of such a field in the document. However, at the moment, if a reader would like to find out more about the field `encryption_password_token` they should scroll down until they find `enc pw token` and a little extra effort is needed to ensure that this the same as `encryption_password_token`. By using the same name a reader can search and find the field in the document without extra scrolling.

### DIFF
--- a/specs/bep-v1.rst
+++ b/specs/bep-v1.rst
@@ -288,16 +288,16 @@ the folder.
 The **label** field contains the folder label, the human readable name of
 the folder.
 
-The **read only** field is set for folders that the device will accept no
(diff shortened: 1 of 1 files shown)


acolomb:
I think the change is fine, only cosmetic.  Since this is touching a versioned protocol specification though, I'll let the other maintainers weigh in on the decision.

PR #812, opened 2023-07-31
add Caddy Proxy to Discovery Server Description

MrOggy85:
## add Caddy Proxy to Discovery Server Description

Update stdiscosrv.rst to include Caddyfile for reverse proxy with Caddy to the Discovery Server

### DIFF
--- a/users/stdiscosrv.rst
+++ b/users/stdiscosrv.rst
@@ -340,6 +340,28 @@ configuration:
 
     RemoteIPHeader X-Forwarded-For
 
+Caddy
+""""""
+The following lines must be added to the Caddyfile:
+
+.. code-block:: caddyfile
+
+    discovery.example.com {
+      reverse_proxy 192.0.2.1:8443 {
+        header_up X-Forwarded-For {http.request.remote.host}
+        header_up X-Client-Port {http.request.remote.port}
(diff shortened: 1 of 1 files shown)


users/stdiscosrv.rst
```diff
@@ -340,6 +340,28 @@ configuration:
...
     RemoteIPHeader X-Forwarded-For
 
+Caddy
+""""""
+The following lines must be added to the Caddyfile:
+
+.. code-block:: caddyfile
+
+    discovery.example.com {
+      reverse_proxy 192.0.2.1:8443 {
```
calmh, line 350:
```suggestion
      reverse_proxy 127.0.0.1:8443 {
```

Maybe? I didn't check now but I think most of the other samples refer to a local instance, and people might not recognize the documentation prefix

MrOggy85: Thank you for your feedback and I agree with you in general. However for this specific document I was using the IP address used in the previous section for nginx (see line `274`). So in order to be consistent for this document I suggest to use `192.0.2.1`. What do you think? @calmh

calmh: Agreed

calmh (approved):

PR #817, opened 2023-08-20
Explain more clearly how to use rest/db/file

timabell:
This adds an example curl call and explains the parameters more clearly.

It's not immediately obvious how to use this endpoint from the existing docs. Hopefully this will help others figure it out. Reading the forums I don't think I was the only person confused by this page.

### DIFF
--- a/rest/db-file-get.rst
+++ b/rest/db-file-get.rst
@@ -1,10 +1,27 @@
 GET /rest/db/file
 =================
 
-Returns most data available about a given file, including version and
-availability. Takes ``folder`` and ``file`` parameters. ``local`` and
-``global`` refer to the current file on disk and the globally newest file,
-respectively.
+Returns data about a given file, including version and
+availability.
+
+Request
+-------
+
+The call requires parameters for `folder` and `file` in the querystring:
+
+- `folder` is the folder code which you can find in the synthing UI, e.g. `5camp-slpa8`
+
+- `file` is the full path to the file you are interested in. The path and filename must be correctly url-encoded.
+
+
+Example Request
+^^^^^^^^^^^^^^^
+
+  curl --silent --get --header "X-API-KEY: YOUR_API_TOKEN" "http://localhost:8080/rest/db/file?folder=YOUR_FOLDER_ID" --data-urlencode "file=path/to/file.pdf"
+
+Response
+--------
+
 
 .. code-block::
 
@@ -18,6 +35,8 @@ respectively.
       "global": { /* a file entry */ },
       "local": { /* a file entry */ }
     }
+
+``local`` and ``global`` refer to the current file on disk and the globally newest file, respectively.
 
 A file entry looks like this::
 


rest/db-file-get.rst
```diff
@@ -1,10 +1,27 @@
...
-availability. Takes ``folder`` and ``file`` parameters. ``local`` and
-``global`` refer to the current file on disk and the globally newest file,
-respectively.
+Returns data about a given file, including version and
+availability.
+
+Request
+-------
+
+The call requires parameters for `folder` and `file` in the querystring:
```
tomasz1986, line 10:
```suggestion
The call requires parameters for `folder` and `file` in the query string:
```

rest/db-file-get.rst
```diff
@@ -1,10 +1,27 @@
...
+
+- `folder` is the folder code which you can find in the synthing UI, e.g. `5camp-slpa8`
```
tomasz1986, line 12:
```suggestion
- `folder` is the folder ID which you can find in the Synthing Web GUI, e.g. `5camp-slpa8`
```

imsodin, line 12: It's called ID in the UI and other docs, so I'd use that here too.

```suggestion
- `folder` is the folder ID which you can find in the synthing UI, e.g. `5camp-slpa8`
```

rest/db-file-get.rst
```diff
@@ -1,10 +1,27 @@
...
+
+- `file` is the full path to the file you are interested in. The path and filename must be correctly url-encoded.
```
tomasz1986, line 14:
```suggestion
- `file` is the full path starting from the folder root to the file you are interested in. The path and filename must be correctly URL-encoded.
```

rest/db-file-get.rst
```diff
@@ -1,10 +1,27 @@
...
+
+
+Example Request
+^^^^^^^^^^^^^^^
+
+  curl --silent --get --header "X-API-KEY: YOUR_API_TOKEN" "http://localhost:8080/rest/db/file?folder=YOUR_FOLDER_ID" --data-urlencode "file=path/to/file.pdf"
```
tomasz1986, line 20:
```suggestion
  curl --silent --get --header "X-API-KEY: YOUR_API_TOKEN" "http://localhost:8384/rest/db/file?folder=YOUR_FOLDER_ID" --data-urlencode "file=path/to/file.pdf"
```

rest/db-file-get.rst
```diff
@@ -19,6 +36,8 @@ respectively.
       "local": { /* a file entry */ }
     }
 
+``local`` and ``global`` refer to the current file on disk and the globally newest file, respectively.
```
tomasz1986, line 39:
```suggestion
``local`` and ``global`` refer to the current file on the local device and the globally newest file, respectively.
```

rest/db-file-get.rst
imsodin, line 14: To me "full path" indicates an absolute file path, while the path here is relative to the folder root path. And seems confusing to mention "path and filename", as if they were somehow separate, I'd just say `The path must ...`.

rest/db-file-get.rst
```diff
@@ -1,10 +1,27 @@
...
-respectively.
+Returns data about a given file, including version and
+availability.
+
+Request
+-------
+
+The call requires parameters for `folder` and `file` in the query string:
+
+- `folder` is the folder ID which you can find in the Synthing Web GUI, e.g. `5camp-slpa8`
```
tomasz1986, line 12:
```suggestion
- `folder` is the folder ID which you can find in the Syncthing Web GUI, e.g. `5camp-slpa8`
```

Missed the typo.

rest/db-file-get.rst
timabell: applied, thanks!

rest/db-file-get.rst
timabell: Fix applied, thanks!

rest/db-file-get.rst
timabell: Good point, alternate wording added in 89ee8d1ab11b5bb1fe9fcf5b3daac1211d9af55e

timabell:
Thanks for all the suggested improvements to my patch, I've applied them all now

rest/db-file-get.rst
```diff
@@ -1,12 +1,31 @@
...
+- `file` is the relative path of the file starting from the folder root to the
+  file you are interested in.  The path and filename must be correctly
+  URL-encoded.
+
+Example Request
+^^^^^^^^^^^^^^^
+
+.. code-block:: bash
+
+    curl --silent --get --header "X-API-KEY: YOUR_API_TOKEN" "http://localhost:8384/rest/db/file?folder=YOUR_FOLDER_ID" --data-urlencode "file=path/to/file.pdf"
```
acolomb, line 23: There are already numerous examples involving `curl` in the docs.  I'd find it convenient to adopt the same CLI options used there, e.g. `-H` instead of `--header` and omit the `--get`.

rest/db-file-get.rst
```diff
@@ -19,7 +38,10 @@ respectively.
       "local": { /* a file entry */ }
     }
 
-A file entry looks like this::
+``local`` and ``global`` refer to the current file on the local device and the
+globally newest file, respectively.  A file entry looks like this:
```
acolomb, line 42: There is a factual error here.  The `mtime` attribute is not part of the respective file entry, but top-level next to `availability`.  When that is fixed, the outer object wrapping can go away in this code sample.

acolomb (changes requested):
I took the liberty to fix some formatting.  Could you please fix the factual discrepancies while you're touching this part?

acolomb (approved):

PR #818, opened 2023-08-25
Make it clear that there is no option to restrict only to LAN

orena1:
I assume many ppl will want to make sure Synchthing is restricted only to LAN. I assume the first attempt would be to set a specific IP in the config, but it is not possible, and will never be:
https://forum.syncthing.net/t/limit-access-to-web-gui/12700/7

I think adding this comment will save some time for other users.

### DIFF
--- a/users/guilisten.rst
+++ b/users/guilisten.rst
@@ -16,6 +16,9 @@ also set a username and a strong password for authentication and check the
 option to use HTTPS. You are otherwise, potentially, opening up your
 Syncthing installation for the world.
 
+Those are the only two options available, for restricting GUI to only LAN
+another alternative is required (Firewall).
+
 Port Numbers
 ------------
 


acolomb:
Just to clarify, what alternative do you see to restrict access using a firewall?  Filter based on source address?  Is that really safe from spoofing and what happens with NATed connections?

I'd say the two documented options are quite clear, either it's open from "the outside" (of the machine it's running on) or it's not.  Protecting from traffic originating outside of the LAN is a task best fit for a router / firewall.  I don't see how a single application could even sensibly filter this.  What would you specify, a source IP range?  That gives a false sense of security IMHO, because it assumes the router is to be trusted.  Imagine e.g. someone putting a second, fast router on the network that hijacks some NAT connections and makes them appear to come from a local IP.  Just an unlikely example to illustrate that a single application program can do very little for network security, that's best left to a proper firewall.

orena1:
Thanks @acolomb , this is great information's. It might be clear to you, but I think there are a few more ppl like me that do not have this knowledge. What do you think about writing or adding some reference to this info in the documentation?

acolomb:
Just opened #829 with an alternative and more elaborate version of this remark.

PR #832, opened 2023-10-14
.

ghost:


tomasz1986:
This one seems weird. In `config.xml`, the option is indeed called `allowedNetwork`. However, in the Advanced Configuration (and many other places in the code), it's the plural form `allowedNetworks`. Why is it so?

er-pa:
The error lies more in the description of https://docs.syncthing.net/users/config.html#config-option-device.allowednetwork than in the plural form elsewhere, imo.

Under the hood the multiple networks become multiple single `<allowedNetwork>` entries, but from a user's perspective it's simply just a list of elements and not limited to one single network.

calmh:
This discrepancy is there in all multiple-valued config items, like `listenAddresses` which is called that in the GUI and in the JSON (where it's a list of strings) but in the XML is repeated `listedAddress` tags, and maybe could use some clarification.

acolomb:
The advanced settings GUI is deduced from the JSON representation.  In XML, it's a list of `<allowedNetwork>` elements, but in JSON it gets serialized to an array named `allowedNetworks`.  The latter is where the option name comes from.  The link to the config docs page does work, because both plural and singular forms are accepted as aliases.

Now we need to be clear on whether the heading on this page refers to the GUI label in the advanced settings modal (`Allowed Networks`), or to the XML description.

acolomb:
Actually I think it would be better to write this page from the perspective of the GUI. The configuration page is the one which describes the intricate details of the XML file, and links to the other pages when there is more to write about the topic, like in this case. But those users who discover this page directly will probably not be sifting through XML, but looking for an explanation what the thing in this "advanced settings" dialog is doing. Even if they just clicked on the question mark icon and came here through the link from the configuration page.

acolomb:
Opened #890 with the mentioned alternative approach.

PR #833, opened 2023-10-22
users: New default config/data locations

calmh:
See syncthing/syncthing#9180

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -19,7 +19,7 @@ Synopsis
 Description
 -----------
 
-.. versionchanged:: 1.26.0
+.. versionchanged:: 1.27.0
 
     The default location of the configuration and database directory on
     Unix-like systems was changed to ``$XDG_STATE_HOME/syncthing`` or
--- a/users/config.rst
+++ b/users/config.rst
@@ -19,13 +19,26 @@ Synopsis
 Description
 -----------
 
+.. versionadded:: 1.26.0
+
+    The default location of the configuration and database directory on
+    Unix-like systems was changed to ``$XDG_STATE_HOME/syncthing`` or
+    ``$HOME/.local/state/syncthing``. Previously the default config location
+    was ``$XDG_CONFIG_HOME/syncthing`` or ``$HOME/.config/syncthing``. The
+    database directory was previously ``$HOME/.config/syncthing`` or, if the
+    environment variable was set, ``$XDG_DATA_HOME/syncthing``. Existing
+    installations may still use these directories instead of the newer
+    defaults.
+
 .. versionadded:: 1.5.0
 
     Database and config can now be set separately. Previously the database was
     always located in the same directory as the config.
 
 Syncthing uses a single directory to store configuration and crypto keys.
-Syncthing also has a database, which is often stored in this directory too.
+Syncthing also keeps an index database with file metadata which is by
+default stored in the same directory, though this can be overridden.
+
 The location defaults to ``$XDG_STATE_HOME/syncthing`` or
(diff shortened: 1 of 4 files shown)


users/config.rst
```diff
@@ -25,11 +26,12 @@ Description
 
 Syncthing uses a single directory to store configuration and crypto keys.
 Syncthing also has a database, which is often stored in this directory too.
-The config location defaults to ``$HOME/.config/syncthing``
-(Unix-like), ``$HOME/Library/Application Support/Syncthing`` (Mac),
-or ``%LOCALAPPDATA%\Syncthing`` (Windows). It can be changed at runtime
-using the ``--config`` flag. In this directory the following files are
-located:
+The config location defaults to ``$XDG_STATE_HOME/syncthing`` or
```
acolomb, line 29:
```suggestion
The location defaults to ``$XDG_STATE_HOME/syncthing`` or
```

I think talking especially about a "config location" here can be confusing after the previous statement.  Please reflow at will if accepting the suggestion. :wink:

imsodin (approved):

users/config.rst
```diff
@@ -48,22 +50,27 @@ located:
 
 The database is stored either in the same directory as the config (usually the
 default), but may also be located in one of the following directories (Unix-like
-platforms only):
+platforms only) for legacy reasons:
```
acolomb, line 53: We should mention the legacy *config* location as well, i.e. `%XDG_CONFIG_HOME/syncthing` and `$HOME/.config/syncthing`, as they are is still tried.

users/config.rst
```diff
@@ -48,22 +50,27 @@ located:
...
-The location of the database can be changed using the ``--data`` flag. The
-``--home`` flag sets both config and database locations at the same time.
-The database contains the following files:
+The database directory contains the following files, among others:
 
 :file:`index-{*}.db`
     A directory holding the database with metadata and hashes of the files
     currently on disk and available from peers.
 
+:file:`syncthing.log`
```
acolomb, line 66: `csrftokens.txt` also lands here if I read the code correctly.  Move from the paragraph above please.

users/faq-parts/troubleshooting.rst
```diff
@@ -128,7 +128,7 @@ The web GUI contains a ``Recent Changes`` button under the device list which
 displays changes since the last (re)start of Syncthing. With the ``--audit``
 option you can enable a persistent, detailed log of changes and most
 activities, which contains a ``JSON`` formatted  sequence of events in the
```
acolomb, line 130: Nitpick: Double space.  Which is fine after a full-stop (end of sentence), but not within.

```suggestion
activities, which contains a ``JSON`` formatted sequence of events in the
```

acolomb (changes requested):
Nicely done, but some corrections necessary.

acolomb:
Oh and we should maybe have a `..versionadded::` tag here somewhere, to make it explicit when the defaults changed.

calmh:
I rewrote stuff to be, in my mind, a bit clearer. A new `versionadded` describes the change. Then we don't need to deal with the change/legacy further down in the doc. I deliberately left out mentioning csrftokens.txt (but missed removing it from the old config section) because it's a tiny implementation wart that should go away (live in the database, if anywhere) -- it's covered by the new "among others" on the database dir. :)

calmh:
n.b. the `versionchanged` is next+1 since I think this is too dangerous a change to slide into an `rc.2`. We should delay merge accordingly (of the docs change), until after 1.26 release.

PR #835, opened 2023-10-29
users: Clarify .stignore is only read from the folder root

YakovL:
This is a follow up to https://forum.syncthing.net/t/is-it-correct-that-stignore-inside-a-subfolder-is-not-applied/20997

Also amended whitespacing in the .md file: as it was written, it was somewhat difficult to find the statement to edit (and also will cause multiple edits when applying the max-length best practice; splitting lines statement-wise seems to be more consistent).

### DIFF
--- a/users/ignoring.rst
+++ b/users/ignoring.rst
@@ -16,11 +16,12 @@ Description
 -----------
 
 If some files should not be synchronized to (or from) other devices, a file called
-``.stignore`` can be created containing file patterns to ignore. The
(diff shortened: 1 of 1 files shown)


acolomb (commented):
Looks good, but please don't change the line wrapping like that. Just follow the pattern of the surrounding text and wrap at a limit of 80 characters (probably, it's not always consistent throughout the docs).

acolomb (approved):

PR #838, opened 2023-11-30
Fixup merge error duplicating lines, wrapping & whitespace.

acolomb:


acolomb:
Oh dang, sometimes it sucks using GitHub to edit stuff.  Sorry for the noise.  Would be good to have direct push access for cases like this.  Seems like @calmh does push commits bypassing the PR flow sometimes?

calmh:
Yep, sure do, when the circumstances warrant it. I've added the same privileges to you, I think, you can try it here.

PR #844, opened 2023-12-13
users: Add link to repository for systemd config files

Linger206:
The 'mentioned above' description could be misleading and not user-friendly.


acolomb:
I think the intention here is to have less repetition of the repo URL.  If we now include it here, we might just as well turn the path into a direct link, e.g.: https://github.com/syncthing/syncthing/raw/main/etc/linux-systemd/user/syncthing.service

What do you think?

acolomb (approved):

acolomb:
Thank you very much!

PR #846, opened 2023-12-25
rest/events: Warn about future event IDs

KeyWeeUsr:
I'm not sure whether somebody else has fallen into this as it might be too obvious yet maybe an extra note/warning about being careful could help, especially when one would want to chunk by let's say multiples of 100 and ended up with a blocking call wondering why.

### DIFF
--- a/rest/events-get.rst
+++ b/rest/events-get.rst
@@ -22,6 +22,11 @@ waits for new events to happen before returning. By default it times out after
 60 seconds returning an empty array. The time out duration can be customized
 with the optional parameter ``timeout=<seconds>``.
 
+.. warning::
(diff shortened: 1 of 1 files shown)


calmh (commented):
Hm. I don't think this warrants a warning as it mostly says again what the previous two paragraphs explain.

KeyWeeUsr:
@calmh I think you might be right. Let me know if you'd like to reword the previous two paragraphs or close the PR.

calmh:
Maybe add a sentence to the preceding paragraph along the lines that asking for an event ID from the future will block until that even ID is reached, or the timeout occurs?

KeyWeeUsr:
@calmh Is it better now?

calmh (approved):

PR #854, opened 2024-01-15
users/ignoring: Update notes on negated ignore patterns and scanning/watching

calmh:
Ref syncthing/syncthing#9340

### DIFF
--- a/users/ignoring.rst
+++ b/users/ignoring.rst
@@ -61,14 +61,14 @@ The ``.stignore`` file contains a list of file or path patterns. The
    special meaning. For example, ``\{banana\}`` matches ``{banana}`` exactly
    and does not denote a set of alternatives as above.
 
-.. note::
+   .. note::
 
-   Escaped characters are not supported on Windows, where ``\`` is the
-   path separator. If you still need to match files that have square or
-   curly brackets in their names, one possible workaround is to replace
-   them with ``?``, which will then match any character. For example,
-   you can type ``?banana?`` to match both ``[banana]`` and
-   ``{banana}``, and so on.
+      Escaped characters are not supported on Windows, where ``\`` is the
+      path separator. If you still need to match files that have square or
+      curly brackets in their names, one possible workaround is to replace
+      them with ``?``, which will then match any character. For example,
+      you can type ``?banana?`` to match both ``[banana]`` and
+      ``{banana}``, and so on.
 
 -  A pattern beginning with ``/`` matches in the root of the synced folder only.
    ``/foo`` matches ``foo`` but not ``subdir/foo``.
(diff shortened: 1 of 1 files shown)


imsodin (approved):
This is quite nice and clear to me, but that's not very significant as I knew how it works before reading it already - we'll see how users take it :)

calmh:
Frankly the whole ignore article should probably be rewritten to be much more user friendly but here we are

users/ignoring.rst
```diff
@@ -98,26 +124,15 @@ The ``.stignore`` file contains a list of file or path patterns. The
    they are preventing directory deletion. This prefix should be used by any OS
    generated files which you are happy to be removed.
 
+   .. note::
```
acolomb, line 127: I'd leave this one outside of the bullet list, as it applies to the last two points equally.

users/ignoring.rst
```diff
@@ -98,26 +124,15 @@ The ``.stignore`` file contains a list of file or path patterns. The
...
+
+      Prefixes can be specified in any order (e.g. ``(?d)(?i)``), but cannot
+      be combined in a single pair of parentheses like :strike:`(?di)`.
+
 -  A line beginning with ``//`` is a comment and has no effect. The same double
```
acolomb, line 132: This should be shuffled way up higher IMO, probably close to the `/rooted_ignore_pattern` bullet point.  Then the closing note about prefix order stands correctly *after* the list, but still directly related to the prefixes.

users/ignoring.rst
```diff
@@ -88,6 +88,32 @@ The ``.stignore`` file contains a list of file or path patterns. The
    are *included* (that is, *not* ignored). This can be used to override
    more general patterns that follow.
 
+   .. note::
+
+      Negated patterns that can match items below the folder root will cause
+      Syncthing to traverse otherwise ignored directories. If the
```
acolomb, line 94: This first sentence requires serious mental capacity to parse, even for me as an advanced user.  Better to split it into smaller statements:

"Syncthing will sometimes look at the files even inside ignored directories.  Some file that matches any negated pattern (thus, should be *included*) could very well reside somewhere below them.  This happens only for directories matching an ignore pattern *after the first negated pattern*."

Then the example, with speaking directory names that directly show what happens:

```
/1-not_looked_at
/2-also_completely_disregarded
!3-filename_to_sync_anyway
!4-foo
/5-could_have_foo_below_so_scan_it
*
```

Not that this is the ideal solution, but hopefully easier to grasp.  So I hope you don't mind me just sketching the concept without following through.

users/ignoring.rst
```diff
@@ -88,6 +88,32 @@ The ``.stignore`` file contains a list of file or path patterns. The
...
+      example::
+
+         /foo
+         /bar
+         !baz
+         *
+
+      The directories ``foo`` and ``bar`` will be entirely ignored. However any
+      other directories present must be scanned entirely to find any items
+      named `baz`, despite the fact that they will be ignored due to the
```
acolomb, line 107: Also a difficult sentence, should be split up at the comma and refactored.  Speaking names in the example will help to reference easily.  Include a number to avoid repeating long names here.

users/ignoring.rst
```diff
@@ -88,6 +88,32 @@ The ``.stignore`` file contains a list of file or path patterns. The
...
+      ``*``. As a special case, top-level rooted patterns (e.g. ``!/foo``) do
+      not cause this behaviour::
```
acolomb, line 109: So far "this behavior" (U.S. English seems to be used mostly in the docs) is not very well understood, which makes it hard to see what exactly top-level rooted patterns really do.  They're obviously negated, but don't cause something which essentially meant not doing something.  So do they do it or don't they? :exploding_head:

And "top-level rooted patterns" is a bit too terse for parsing correctly IMO, even with the example.  The example is also negated btw. so does the remark apply to all top-level rooted negated patterns or what?

users/ignoring.rst
```diff
@@ -88,6 +88,32 @@ The ``.stignore`` file contains a list of file or path patterns. The
...
+
+         !/baz
+         *
+
+      In this case, only the directory ``baz`` will be scanned, since
+      everything else is ignored by the ``*`` pattern.
```
acolomb, line 115: Despite having seen (but not thoroughly checked) the related PR, this special case suprised me.  I.e. could be better explained.

users/ignoring.rst
```diff
@@ -88,6 +88,32 @@ The ``.stignore`` file contains a list of file or path patterns. The
...
+      watched. Directories ignored before the first negated pattern can
+      however be safely skipped, since the first matching pattern wins. For
+      example::
+
+         /foo
+         /bar
+         !baz
+         *
+
+      The directories ``foo`` and ``bar`` will be entirely ignored. However any
```
acolomb, line 105: A note about the version where this optimized behavior was introduced would help.  Yes, we do have the docs versioning, but an unsuspecting user might not come up with the idea that the handling was ever different.

acolomb (commented):
@calmh Sorry for being late.  I do think that the text definitely needs improvments to make it easier to grasp.  Some suggestions follow, but I didn't try to rewrite it completely.  Also because I'm not completely sure I did understand it myself.

acolomb:
@calmh Are you going to follow up on this?  See my previous review comments, specifically.

PR #857, opened 2024-02-16
Update reverseproxy.rst - added HAProxy example

Schm1tz1:
I added HAProxy example that I am running in my setup (TLS termination at Proxy with own certificate).
The most important part is to set the **Host** headers correctly (unless one wants to use _insecureSkipHostcheck_).

### DIFF
--- a/users/reverseproxy.rst
+++ b/users/reverseproxy.rst
@@ -82,6 +82,21 @@ Caddy v2
         }
     }
 
+HAProxy
+~~~~~~~
+
+.. code-block:: none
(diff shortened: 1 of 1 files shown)


acolomb:
I'm sorry nobody has reacted to your PR yet. Just guessing not many people are actually using such a setup with Syncthing, and therefore your addition couldn't be verified. I for one can't judge if it will be a worthwhile change or maybe misleading to other users.

How confident are you that this example will actually help others and work for them?

PR #861, opened 2024-03-26
Update autostart.rst

MrAnimaniac:
Added workaround for not working "--no-console" flag

### DIFF
--- a/users/autostart.rst
+++ b/users/autostart.rst
@@ -176,6 +176,10 @@ user account in Windows. No console or browser window will pop-up, but
 you can still access the interface by opening http://localhost:8384 in
 a Web browser.
 
+Using newer versions of Windows, the flag "--no-console" is reported to be not working
+correctly. An easy workaround is to use ``stctl.exe --start`` instead of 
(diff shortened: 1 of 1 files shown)


calmh:
What's `stctl.exe`?

er-pa:
> What's `stctl.exe`?

https://github.com/Bill-Stewart/SyncthingWindowsSetup?tab=readme-ov-file#helper-tools

Wrapper-specific tooling, it seems.

calmh:
Cool, so it's part of the WindowsSetup package, we should reference that.

users/autostart.rst
```diff
@@ -176,6 +176,10 @@ user account in Windows. No console or browser window will pop-up, but
 you can still access the interface by opening http://localhost:8384 in
 a Web browser.
 
+Using newer versions of Windows, the flag "--no-console" is reported to be not working
```
tomasz1986, line 179:
```suggestion
When using Windows 10 or 11 with Windows Terminal set as the default shell, the flag "--no-console" is reported to be not working
```

PR #862, opened 2024-04-09
users/config: be more precise about trafficClass value

kaueraal:
Although described, it was not obvious to me what exact value needs to be specified. To improve this, it was explicitly stated that it's about the respective [IPv4](https://en.wikipedia.org/wiki/Type_of_service) and [IPv6](https://en.wikipedia.org/wiki/IPv6_packet#Fixed_header) header fields.

Additionally, users most likely want to specify a [DSCP](https://en.wikipedia.org/wiki/Differentiated_services) value. To help in this case the required bit shift was mentioned.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -1221,7 +1221,10 @@ The ``options`` element contains all other global configuration options.
 
 .. option:: options.trafficClass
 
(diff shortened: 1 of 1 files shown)


calmh (approved):
👌

PR #863, opened 2024-05-10
Add example command to changing the listen address

sjehuda:


acolomb (commented):
This page is about the overall concept, not directly related to the CLI.  So at least the new heading should mention what the context is.  And I'm a bit hesitant to "duplicate" this information here, since the CLI API is auto-generated from the configuration structure and if the command ever changes, this place might get overlooked to adjust.

sjehuda:
I suggest to add an example command to every page, as it woud save time.

sjehuda:
https://forum.syncthing.net/t/accessing-gui-from-an-external-ip/22113/2

users/guilisten.rst
```diff
@@ -60,3 +60,12 @@ To use a port number lower than 1024, you will need to:
...
 We do not recommend using a port number lower than 1024.
+
+Example
+-------
+
+Changing the listen address to ``0.0.0.0:8384``.
+
+```
+syncthing cli config gui raw-address set 0.0.0.0:8384
+```
```
marbens-arch, line 71: Triple backticks don't work with reStructuredText.
```suggestion
``syncthing cli config gui raw-address set 0.0.0.0:8384``
```

users/guilisten.rst
```diff
@@ -60,3 +60,12 @@ To use a port number lower than 1024, you will need to:
   port number.
 
 We do not recommend using a port number lower than 1024.
+
+Example
+-------
+
+Changing the listen address to ``0.0.0.0:8384``.
```
marbens-arch, line 67: It's not exactly clear what the context is from the header name, like @acolomb said.
```suggestion
Using the CLI to change the GUI Listen Address
-------

Here is an example of changing the listen address to ``0.0.0.0:8384`` using the CLI:
```

marbens-arch (changes requested):

PR #864, opened 2024-05-18
chore(dev): update source code layout in intro

rasa:


calmh (approved):

dev/intro.rst
```diff
@@ -150,15 +141,18 @@ lib/
...
-   Legacy location of the protocol package.
-
 script/
    Various utility scripts for auto generating stuff and so on.
 
+  genassets/
+      Generates asset files that are compiled into ``syncthing`` as part of the build process (build utility).
+
+  transifexdl/
+      Downloads translations from Transifex (build utility).
```
calmh, line 151: There seems to be some inconsistent indentation here, not sure how significant

rasa: Fixed, thx.

imsodin:
Thanks. I merged as it's clearly an improvement, however it's also already/still out-of-date with reality. Kinda highlights we likely either should automate this or make it less detailed (only describe the intent/overall structure, not dir-per-dir, in the hopes of getting out of date less quickly that way :) ).

PR #870, opened 2024-07-21
Add WebAuthn documentation

emlun:
Documentation for https://github.com/syncthing/syncthing/pull/9175.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -144,6 +144,9 @@ The following shows an example of a default configuration file (IDs will differ)
             <address>127.0.0.1:8384</address>
             <apikey>k1dnz1Dd0rzTBjjFFh7CXPnrF12C49B1</apikey>
             <theme>default</theme>
+            <webauthnUserId>4_KyxKWr6x2KvB3GGHYLkjmn1M6xTip5ITZQUgaUzJW5e023M0j4NBOkgR-4aQarM7RRCv7TGkmOD53kQBPhLQ==</webauthnUserId>
+            <webauthnRpId>localhost</webauthnRpId>
+            <webauthnOrigin>https://localhost:8384</webauthnOrigin>
         </gui>
         <ldap></ldap>
         <options>
@@ -842,6 +845,8 @@ set on the ``gui`` element:
 
     If not ``true``, the GUI and API will not be started.
 
+.. _gui-tls:
+
 .. option:: gui.tls
     :aliases: gui.useTLS
 
@@ -941,6 +946,43 @@ The following child elements may be present:
 
     When this setting is disabled, the GUI will not send 401 responses so users
     won't see browser popups prompting for username and password.
+
+
+.. option:: gui.webauthnUserId
+
+    .. versionadded:: TODO 1.28.0
+
+    The base64url-encoded `user handle <https://www.w3.org/TR/webauthn/#user-handle>`_
+    to use when registering WebAuthn credentials (passkeys).
+    This is automatically set and should usually never need to be changed.
+    Authenticators may use this to overwrite existing credentials
+    with the same combination of user handle and RP ID when creating a new credential.
+
+.. option:: gui.webauthnRpId
+
+    .. versionadded:: TODO 1.28.0
+
+    The `RP ID <https://www.w3.org/TR/webauthn/#rp-id>`_
+    to use for WebAuthn (passkey) registration and authentication.
+    If not set, this defaults to ``localhost``.
+
+    The RP ID is a domain name and must be the same as or a parent domain of the
+    domain where the Syncthing GUI is hosted. The RP ID cannot be a raw IP address.
+
+    When you register a new WebAuthn credential (passkey), it gets tied to this RP ID.
+    If you change the RP ID, any existing keys tied to the previous RP ID will stop working.
+
+.. option:: gui.webauthnOrigin
+
+    .. versionadded:: TODO 1.28.0
+
+    The scheme, host and port of the address where WebAuthn logins will take place.
+    If not set, this defaults to ``https://localhost:8384``.
+    WebAuthn registration and login will only work if the GUI is hosted at exactly this host address
+    (excluding the path, query string and hash fragment).
+
+    In general, this should be set to ``https://<webauthnRpId>:<address port>``,
+    omitting the port if it is ``443``.
 
 
 LDAP Element
--- a/users/webauthn.rst
+++ b/users/webauthn.rst
@@ -1,0 +1,39 @@
+WebAuthn (Passkey) Authentication
+=================================
+
+Syncthing can be configured to allow GUI authentication using `WebAuthn <https://webauthn.guide>`_ (passkeys)
+as an alternative to a password.
+
+To enable WebAuthn, the GUI must use HTTPS (see config :ref:`gui.tls <gui-tls>`)
+and must be served at exactly the address ``https://localhost:8384``.
+
+The GUI address can be customized via the advanced GUI settings
+:stconf:opt:`gui.webauthnRpId` and :stconf:opt:`gui.webauthnOrigin`.
+
+If you access the GUI at some other address than ``https://localhost:<port>``,
+you'll need to set the ``webauthnRpId`` setting to the domain name or a parent domain name of that address
+and ``webauthnOrigin`` to the full address including scheme and port (except the default port), but not path.
+For example, if you serve the GUI at the address ``https://syncthing.mydomain.org:8443/syncthing/gui``,
+set ``webauthnRpId`` to one of ``mydomain.org`` or ``syncthing.mydomain.org``
+and set ``webauthnOrigin`` to ``https://syncthing.mydomain.org:8443``.
+
+WebAuthn authentication will be enabled if you have at least one `credential` registered.
+A credential is a public-private key pair stored either on an external security key,
+or a `platform credential` stored on your computer or phone.
+Some platforms might sync platform credentials between devices signed into the same cloud account.
+
+Use the settings GUI to register a new credential.
+
+.. note::
+   We use the term "passkey" more inclusively here than usual.
+   For technical reasons, the term "passkey" usually means
+   a credential that consumes storage space on the authenticator device.
+   Some external security keys have limited storage capacity
+   and therefore also have a limited capacity for storing passkeys.
+
+   However, because Syncthing has only a single user account per installation,
+   we can enable the same use cases as passkeys
+   but with credentials that do not need to consume storage space.
+   We therefore sometimes refer to WebAuthn credentials in Syncthing as "passkeys",
+   because they enable most of the same UI flows as passkeys,
+   even though they do not consume storage space on external security keys like passkeys usually do.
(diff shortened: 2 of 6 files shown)


users/config.rst
```diff
@@ -943,6 +948,43 @@ The following child elements may be present:
     won't see browser popups prompting for username and password.
 
 
+.. option:: gui.webauthnUserId
+
+    .. versionadded:: TODO 1.28.0
+
+    The base64url-encoded `user handle <https://www.w3.org/TR/webauthn/#user-handle>`_
+    to use when registering WebAuthn credentials (passkeys).
+    This is automatically set and should usually never need to be changed.
```
rasa, line 957:
```suggestion
    This is automatically set and should usually not need to be changed.
```

users/config.rst
```diff
@@ -943,6 +948,43 @@ The following child elements may be present:
...
+
+.. option:: gui.webauthnRpId
+
+    .. versionadded:: TODO 1.28.0
+
+    The `RP ID <https://www.w3.org/TR/webauthn/#rp-id>`_
+    to use for WebAuthn (passkey) registration and authentication.
+    If not set, this defaults to ``localhost``.
+
+    The RP ID is a domain name and must be the same as or a parent domain of the
```
rasa, line 969:
```suggestion
    The RP ID is a domain name and must be the same as, or a parent domain of, the
```

users/webauthn.rst
```diff
@@ -0,0 +1,39 @@
+WebAuthn (Passkey) Authentication
+=================================
+
+Syncthing can be configured to allow GUI authentication using `WebAuthn <https://webauthn.guide>`_ (passkeys)
+as an alternative to a password.
+
+To enable WebAuthn, the GUI must use HTTPS (see config :ref:`gui.tls <gui-tls>`)
+and must be served at exactly the address ``https://localhost:8384``.
```
rasa, line 8:
```suggestion
and must be served at exactly the address ``https://localhost:8384``, unless configured otherwise, as described below.
```

users/config.rst
emlun: Done, thanks!

users/config.rst
emlun: Done, thanks!

users/webauthn.rst
emlun: Done, thanks!

emlun:
Added a section on the "2FA" setting (renamed from "Require UV") covering most of what I laid out in https://github.com/syncthing/syncthing/pull/9175#issuecomment-2253145692.

users/webauthn.rst
```diff
@@ -0,0 +1,39 @@
...
+   a credential that consumes storage space on the authenticator device.
+   Some external security keys have limited storage capacity
+   and therefore also have a limited capacity for storing passkeys.
+
+   However, because Syncthing has only a single user account per installation,
+   we can enable the same use cases as passkeys
+   but with credentials that do not need to consume storage space.
+   We therefore sometimes refer to WebAuthn credentials in Syncthing as "passkeys",
+   because they enable most of the same UI flows as passkeys,
+   even though they do not consume storage space on external security keys like passkeys usually do.
```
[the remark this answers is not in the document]
emlun: Like this :slightly_smiling_face: https://www.yubico.com/blog/yubicos-u2f-key-wrapping/

Essentially, one of the outputs of the registration procedure, alongside the newly generated public key, is a _credential ID_, sometimes also known as a "key handle". When initiating an authentication ceremony, the caller can specify zero or more credential IDs that are eligible for the authentication. The trick is that this means the private key can actually be stored encrypted\* inside that credential ID (this is probably where the term "key handle" comes from) instead of in persistent memory inside the secure element, so it doesn't consume storage space on the authenticator (for example a USB security key).

\* Or equivalently: the credential ID can encode parameters to combine with a static base secret for deterministic key (re)generation, as illustrated in the above link.

Of course the drawback of this is that the credential cannot be used unless the caller is able to specify the credential ID. For sites with more than one user, that means you first need to identify the user - so that prevents the "username-less" authentication flow you get if you specify zero credential IDs, which means that any credential registered on that site is eligible, and the user is identified after the fact when the server receives the authentication signature and its associated credential ID. This kind of credential is what's commonly called a "passkey", and those do consume storage on the authenticator.

But in Syncthing we don't actually need to identify the user, because there is only one, so we can always send that user's credential IDs. So we can still enable the same login flow as with a passkey, but our "passkeys" don't need to consume storage space on the authenticator unless the authenticator wants them to.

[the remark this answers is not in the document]
emlun: Ah, but I also see now that the grammar above doesn't quite add up:

>A credential is a public-private key pair stored either on an external security key, or a `platform credential` stored on your computer or phone.

I probably half-wrote and rewrote this a few times and failed to thoroughly re-read it.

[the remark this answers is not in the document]
emlun: How about this? Is this easier to follow?

```
WebAuthn authentication will be enabled if you have at least one `credential`, also called a `passkey`, registered.
A credential is a public-private key pair that is stored on an `authenticator`,
which could be an external security key, a smartphone, or built into your computer.
Some platforms might sync platform credentials between devices signed into the same cloud account.

Use the settings GUI to register a new credential.

.. note::
   We use the term "passkey" more inclusively here than usual.
   A "passkey" is a credential that enables "username-less login",
   which identifies the user automatically without needing them to enter a username first.
   For technical reasons, this is incompatible with a cryptographic trick commonly used by external security keys
   to support an unlimited number of credentials without consuming storage space.
   Therefore, a "passkey" generally must consume storage space on the authenticator.

   However, because a Syncthing instance has only a single user account,
   we can enable "username-less login" without preventing the unlimited storage trick.
   We therefore sometimes refer to WebAuthn credentials in Syncthing as "passkeys",
   even though they do not consume storage space on external security keys like passkeys generally do.
```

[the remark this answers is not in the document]
imsodin: Thanks for the link, neat. 

And yes I can follow now, however it's always easier to do that when you already grok things. It's definitely still quite technical, but that's unavoidable.

Oh and on a complete tangent on the topic of "username-less" auth, just out of curiosity - feel free to ignore :)  
Why can't we get effectively username-less auth on multi-user servers by adding another roundtrip to the ceremony: The hardware key has a public identifier (not username, one global ID for the device) and on registration sends that along with the key handle. And on login the sever requests that key identifier, finds the fitting key handles, sends them back and done. Sure it's another roundtrip, but it still avoids the user input.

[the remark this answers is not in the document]
emlun: > The hardware key has a public identifier (not username, one global ID for the device)

One of the [explicit design goals](https://www.w3.org/TR/2021/REC-webauthn-2-20210408/#sctn-non-correlatable-credentials) of WebAuthn is to not do just that :slightly_smiling_face: because that would be a global correlation handle by which you could trivially track people across the entire web, with even greater confidence than a tracking cookie. A per-device, per-service identifier would also not work since it would allow the service to link multiple accounts to the same person. So no, within these privacy-first design constraints I don't think it's possible for the authenticator to always know the appropriate user ID and key to use, because that would limit you to at most one key per service per device.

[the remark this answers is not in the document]
imsodin: Oh, I should have caught that - thanks again for the explanation. I did get through quite a bit of this PR, but far from done. And calling it a day now :)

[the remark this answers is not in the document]
emlun: Text updated in 97f8023aa6edc38713f2ac190ada59c98dfb39e9 as discussed above.

users/webauthn.rst
```diff
@@ -10,25 +10,24 @@ and must be served at exactly the address ``https://localhost:8384``,
 unless configured otherwise as described in :ref:`webauthn-custom-gui-address`.
 
 WebAuthn authentication will be enabled if you have at least one `credential`, also called a `passkey`, registered.
-A credential is a public-private key pair stored either on an external security key,
-or a `platform credential` stored on your computer or phone.
+A credential is a public-private key pair that is stored on an `authenticator`,
+which could be an external security key, a smartphone, or built into your computer.
```
rasa, line 14:
```suggestion
which could be a smart device (such as a phone, tablet, or watch), a dedicated hardware security key (such as a YubiKey), or built right into your computer.
```
"security key" might be too vague to the uninitiated, and aren't they all external?

users/webauthn.rst
```diff
@@ -10,25 +10,24 @@ and must be served at exactly the address ``https://localhost:8384``,
...
+   For technical reasons, this is incompatible with a cryptographic trick commonly used by external security keys
+   to support an unlimited number of credentials without consuming storage space.
+   Therefore, a "passkey" generally must consume storage space on the authenticator.
+
+   However, because a Syncthing instance has only a single user account,
+   we can enable "username-less login" without preventing the unlimited storage trick.
    We therefore sometimes refer to WebAuthn credentials in Syncthing as "passkeys",
-   because they enable most of the same UI flows as passkeys,
-   even though they do not consume storage space on external security keys like passkeys usually do.
+   even though they do not consume storage space on external security keys like passkeys generally do.
```
rasa, line 30:
```suggestion
   even though they do not consume storage space on hardware security keys like passkeys generally do.
```

users/webauthn.rst
```diff
@@ -10,25 +10,24 @@ and must be served at exactly the address ``https://localhost:8384``,
...
-   a credential that consumes storage space on the authenticator device.
-   Some external security keys have limited storage capacity
-   and therefore also have a limited capacity for storing passkeys.
-
-   However, because Syncthing has only a single user account per installation,
-   we can enable the same use cases as passkeys
-   but with credentials that do not need to consume storage space.
+   A "passkey" is a credential that enables "username-less login",
+   which identifies the user automatically without needing them to enter a username first.
+   For technical reasons, this is incompatible with a cryptographic trick commonly used by external security keys
```
rasa, line 23:
```suggestion
   For technical reasons, this is incompatible with a cryptographic trick commonly used by hardware security keys
```

users/webauthn.rst
```diff
@@ -85,3 +85,8 @@ and ``webauthnOrigin`` to the full address including scheme and port (except the
 For example, if you serve the GUI at the address ``https://syncthing.mydomain.org:8443/syncthing/gui``,
 set ``webauthnRpId`` to one of ``mydomain.org`` or ``syncthing.mydomain.org``
 and set ``webauthnOrigin`` to ``https://syncthing.mydomain.org:8443``.
+
+For hostnames other than ``localhost`` you will also need an HTTPS certificate your browser considers valid.
+For guidance on how to create or obtain one, see for example
+`OpenSSL Cookbook <https://www.feistyduck.com/library/openssl-cookbook/online/>`_
+or `Let's Encrypt <https://letsencrypt.org/getting-started/>`_.
```
rasa, line 92: How about adding:

To create locally trusted HTTPS certificates on the command-line see 
`mkcert <https://github.com/FiloSottile/mkcert>`_.

acolomb:
For the record, I'd like some time to review this as well before someone merges it.  Just waiting for the main PR to settle down a bit.

PR #873, opened 2024-08-08
Update and rename contrib.rst to Add Syncthing GTK Fork

Rikiub:
- https://github.com/kozec/syncthing-gtk is unmaintained.
- A fork of this project was made in https://github.com/syncthing-gtk/syncthing-gtk
- Also it's the version available in Flatpak https://flathub.org/apps/me.kozec.syncthingtk

### DIFF
--- a/users/Add Syncthing GTK Fork
+++ b/users/Add Syncthing GTK Fork
@@ -20,6 +20,10 @@ Cross-platform
 Android
 ~~~~~~~
 
+- `Syncthing GTK <https://github.com/syncthing-gtk/syncthing-gtk>`_
+
+  GUI and notification area icon for Syncthing.
+
 - `syncthing-android <https://github.com/syncthing/syncthing-android>`_
 
   A wrapper app for the Syncthing binary.


acolomb:
Your renaming of the file will not work.

Please file a proper patch against the existing file.  Ideally, you would only add a remark to the section where Syncthing-GTK is already mentioned.  Pointing to the new *and* old URL might be better for people to recognize the software.

Rikiub:
> Ideally, you would only add a remark to the section where Syncthing-GTK is already mentioned

Don't make sense because it's already in [Older, Possibly Unmaintained](https://github.com/syncthing/docs/blob/09a11d0befdd223c22d9771a18dd677d02b14c54/users/Add%20Syncthing%20GTK%20Fork#L280) section. The purpose of this PR is [revive the project](https://github.com/syncthing/docs/blob/09a11d0befdd223c22d9771a18dd677d02b14c54/users/Add%20Syncthing%20GTK%20Fork#L274-L277), so make sense add it again to "Linux" section.

I don't understand, need I add something like _"This is a fork of the old project <link>"_ on the description?
I would like keep things simple.

imsodin:
The new one doesn't look very maintained, last activity is a year ago. Seems like it never actually took off. So maybe just add the link and leave it in the unmaintained section.

Rikiub:
It makes sense.
I move and replace it to "Unmaintained" section.

Hope the project get up.

acolomb:
You really need to stop renaming files.  Whatever you do while working on the PR, it leads to the file being renamed after matching your commit message.  Don't do that, please.

users/Add from where was forked
```diff
@@ -273,7 +273,7 @@ Older, Possibly Unmaintained
    accordingly.
 
 - `a-sync <https://github.com/davide-imbriaco/a-sync>`_
-- `Syncthing-GTK <https://github.com/kozec/syncthing-gtk>`_
+- `Syncthing GTK <https://github.com/syncthing-gtk/syncthing-gtk>`_ (Fork from `Kozec <https://github.com/kozec/syncthing-gtk>`_)
```
acolomb, line 276:
```suggestion
- `Syncthing-GTK <https://github.com/syncthing-gtk/syncthing-gtk>`_ (Fork from `Kozec <https://github.com/kozec/syncthing-gtk>`_)
```

The project is actually named with a hyphen.

acolomb (changes requested):

Rikiub:
> You really need to stop renaming files. Whatever you do while working on the PR, it leads to the file being renamed matching your commit message. Don't do that, please.

Sorry, I not notice it in the first commit.
Now should be ok.

acolomb (approved):

PR #876, opened 2024-09-01
wording changes to remove ambiguity

ersanchez:
Changed the "This" at the beginning of the sentence to more specifically refer to one of the three files described in the preceding sentence (local, others, global/"best").


acolomb (commented):
Sorry, this is not really an improvement and makes it harder to understand for me.  More "version" repetitions do not help.  And right after the *best* version was defined, you reference it again by a different term (that was previously an example).  If anything, the just defined *best* version should be named again, or the term dropped completely in favor of the *global* version.

PR #878, opened 2024-09-20
Added another scenario for a service-based configuration on Windows

wsw70:
Another case where a Windows service is the only solution is when the IP one wants to bind to may not be available yet. This is the case with VPNs, in my case - Tailscale. A dependency on another service is how this problem is solved.


calmh (commented):
Yeah I think it's a mistake to begin with to say "the only scenario(s) ..." etc. Adding more alternatives just invites further additions "this scenario also makes sense!". Let's just simplify to the first sentence and something like "Running as a service is useful in scenarios with specific requirements not met by running as a regular user account" or something like that.

PR #882, opened 2024-11-12
chore(dev): warn from-source builders that syncthing auto-upgrades

orblivion:
Some people may prefer to build from source for their own use. They should be aware that the default behavior is to overwrite the binary with a new downloaded binary when an upgrade is available.


orblivion:
Speaking for myself, I was taken by (unpleasant) surprise here when I realized it was happening. I think there should at least be a warning.

acolomb (approved):
Thanks for the addition.  I snuck in some changes for consistency, always using `--double-dash` option style.

PR #886, opened 2024-12-01
docs(contrib): update Syncthing GUI wrappers for Linux

salim-b:


acolomb:
Thanks for your contribution.  I must say I'm not really fond of introducing another level of headings here.  Could you please simplify your patch to remove those again?  Reordering is fine, to keep all GNOME Shell extensions together.  Just not another two headings.

salim-b:
Ist it ok with you to use a nested list à la


### Linux

- GNOME Shell extensions
  - `Syncthing Icon <https://extensions.gnome.org/extension/989/syncthing-icon/>`_

    A Syncthing status icon in the top bar.
  - `Syncthing Indicator <https://extensions.gnome.org/extension/1070/syncthing-indicator/>`_

    A Syncthing indicator for starting, monitoring and controlling the Syncthing daemon using systemd.
  - ...
- Other
  - `SyncThingy <https://github.com/zocker-160/SyncThingy>`_

    Simple tray indicator written in C++ targeted at Flatpak users.

?

acolomb:
I guess with the current number of entries it's not worth the separation.  That's personal opinion, partly based on a scientific writing rule I remember to only introduce a sub-level if there are at least three headings in it.

salim-b:
> I guess with the current number of entries it's not worth the separation.

I truly think it is since otherwise it doesn't immediately become clear for readers that the first entries are gnome shell extensions – only the URLs provide that hint.

acolomb:
Actually the description of each included the fact, until you removed it.

salim-b:
> Actually the description of each included the fact, until you removed it.

yup, because I think it's more elegant / less redundant that way

acolomb:
How about moving the distinction up one level?

    Linux - GNOME Shell Extensions

    Linux - Generic

I'm having a hard time coming up with a nice heading for the second category.  One is pretty targeted (Steam Deck), the other two are for completely different environments, but more generic than only GNOME.

So if it's hard to make a meaningful separation, let's just not do it.  "GNOME Shell" / "Not GNOME Shell" is not meaningful.  It puts GNOME in a special position, which it doesn't have in our Syncthing world.

salim-b:
I'm fine with moving the distinction up one level.

But I disagree with

> "GNOME Shell" / "Not GNOME Shell" is not meaningful. It puts GNOME in a special position, which it doesn't have in our Syncthing world.

It's not about Gnome having a special position in Syncthing world but about it being one of *many* Linux DEs, so it's meaningful to all Linux users.

acolomb:
Sorry I don't get your argument.  Yes, it's one of *many* DEs, I agree.  My consequence from that is that it should not be the only one we specifically mention in Syncthing docs.

Please, revert to the simple flat list under one "Linux" heading and we can apply the rest of your (sensible and welcome) fixes.

salim-b:
> My consequence from that is that it should not be the only one we specifically mention in Syncthing docs with its own heading / section / grouping.

Sure it *shouldn't* be, but it currently still is. Nothing hinders others to add more KDE-, Cosmic- or whatever specific software.

Anyways, I applied your suggestion :)

acolomb:
Sorry, you misunderstood.  My suggestion was not something I am happy with, but just an idea to spark the discussion about the separation.

salim-b:
Well, I think the idea is fine and a good compromise until more DE-specific software is added.

acolomb:
I don't and frankly I didn't expect this to turn into a lengthy discussion.  Conditions for merging have been stated.  We can also merge as is and then revert the headings in another PR.  But what's the point to that?

salim-b:
I'm just trying to serve Syncthing (Linux) users when grouping gnome shell extensions together. This is an improvement over the current status quo. Nothing hinders you from further improving on this once you "come with a nice heading for the second category", as you wrote. :)

users/contrib.rst
```diff
@@ -47,33 +47,37 @@ macOS
...
   It hosts and wraps Syncthing, making it behave more like a native macOS application and less like a command-line utility with a web browser interface.
 
-Linux
-~~~~~
-
-- `SyncThingy <https://github.com/zocker-160/SyncThingy>`_
-
-  Simple tray incidator written in C++ targeted at Flatpak users.
+Linux – GNOME Shell extensions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
acolomb, line 51:
```suggestion
Linux
~~~~~
```

users/contrib.rst
```diff
@@ -47,33 +47,37 @@ macOS
...
 
 - `Syncthing Icon <https://extensions.gnome.org/extension/989/syncthing-icon/>`_
 
-  A GNOME Shell extension displaying a Syncthing status icon in the top bar.
+  A Syncthing status icon in the top bar.
```
acolomb, line 55:
```suggestion
  GNOME Shell extension: A Syncthing status icon in the top bar.
```

users/contrib.rst
```diff
@@ -47,33 +47,37 @@ macOS
...
 
 - `Syncthing Indicator <https://extensions.gnome.org/extension/1070/syncthing-indicator/>`_
 
-  A GNOME Shell indicator for starting, monitoring and controlling the Syncthing daemon using systemd.
+  A Syncthing indicator for starting, monitoring and controlling the Syncthing daemon using systemd.
```
acolomb, line 59:
```suggestion
  GNOME Shell extension: A Syncthing indicator for starting, monitoring and controlling the Syncthing daemon using systemd.
```

users/contrib.rst
```diff
@@ -47,33 +47,37 @@ macOS
...
+
+- `Syncthing Toggle <https://extensions.gnome.org/extension/7180/syncthing-toggle/>`_
+
+  A GNOME quick setting for turning on and off the Syncthing systemd service and opening the Web GUI.
```
acolomb, line 63:
```suggestion
  GNOME Shell extension: A quick setting for turning on and off the Syncthing systemd service and opening the Web GUI.
```

users/contrib.rst
```diff
@@ -47,33 +47,37 @@ macOS
...
+
+Linux – Generic
+~~~~~~~~~~~~~~~
```
acolomb, line 66:
```suggestion
```

acolomb (changes requested):

acolomb (approved):

PR #888, opened 2024-12-13
Update contrib.rst

Gaudi111:
Added QNAP updated links.

### DIFF
--- a/users/contrib.rst
+++ b/users/contrib.rst
@@ -206,8 +206,9 @@ Synology NAS (DSM)
 QNAP NAS (QTS)
 ~~~~~~~~~~~~~~
 
-`Syncthing QPKG <https://qnapclub.eu/en/qpkg/692>`__ (Qnap
-Package) available for ALL models x86, x86\_64, Arm (all including new models).
+`Syncthing QPKG <https://www.myqnap.org/product/syncthing/>`__ (Qnap Package).
+`Syncthing QPKG as root (This version "run as root (admin)" for people who don’t know how to configure their space and for very old devices) <https://www.myqnap.org/product/syncthing-run-as-root/>`__ (Qnap Package).
+`available for ALL models x86, x86\_64, Arm (all including new models)
 
 RockStor
 ~~~~~~~~


imsodin:
Are you the creator of that package or a user, resp. in either case, are you able to provide some info around the following questions/concerns? This contrib page isn't vetted or anything, but if possible I'd still like to understand a little bit about what we are linking to and I have no idea how qnap packages work (and couldn't find answers quickly on the linked page), thus the questions. Any info, also partial/different is appreciated, but obviously if you can't answer and you just wanted to add the link, that's perfectly fine too:  
Why is running as root a separate package, and not an option (or something, e.g. install choice) in the existing one? Presumably it's just simpler - or maybe there is no way to have options in qnap packages.  
How are qnap packages created/defined - is there some info/manifest/repo for that somewhere? And related, is it the same person/group creating both syncthing qnap packages?

Gaudi111:
Hi Imsodin, let me address your questions.

> Are you the creator of that package or a user, resp. in either case, are you able to provide some info around the following questions/concerns? 
I am not the creator. I am a QNAP user and have some activity in the [forum.qnap](https://forum.qnap.com/) community.
The link you provided in your WiKi is outdated, as the repository listed has been long superceeded byt he one I have posted.
I have been in contact with the package creatorto address some issues and requests, and he has been very helpful.
You can check his Discord channel:
https://discord.com/channels/1037035252792709130/1037036960650375168

> This contrib page isn't vetted or anything, but if possible I'd still like to understand a little bit about what we are linking to and I have no idea how qnap packages work (and couldn't find answers quickly on the linked page), thus the questions. 
Seems perfectly reasonable, so I am responding to the best of my knowledge.

> Any info, also partial/different is appreciated, but obviously if you can't answer and you just wanted to add the link, that's perfectly fine too: Why is running as root a separate package, and not an option (or something, e.g. install choice) in the existing one? 
I am unaware of the reason. It seems that the package on old units require some fiddling with permissions, that many users are unable to adjust, hence the root package. For further information I suggest you come in contact with Stephane, the developer.

> Presumably it's just simpler - or maybe there is no way to have options in qnap packages. How are qnap packages created/defined - is there some info/manifest/repo for that somewhere? And related, is it the same person/group creating both syncthing qnap packages?
I am unaware, you may contact the developer. He is very supportive, and hopefully he should be able to fully answer your concerns/questions.


Best regards

PR #890, opened 2025-01-06
chore(config): clarify repeated XML tags vs. comma-separated in GUI

acolomb:
Add similar text to all repeatable configuration options, mentioning
the difference between comma-separated lists in the GUI vs. repeating
the tag (without plural "s") in the XML file.  Loosely based on the
existing sentence for `<listenAddress>`.


calmh (approved):
LGTM

PR #894, opened 2025-02-05
Untrusted devices will not delete data

WhyNotHugo:
Explicitly mention that untrusted devices can delete files from a synchronised folder. While this can be inferred from the current description, it's probably best to be explicitly for this kind of caveat.

### DIFF
--- a/users/untrusted.rst
+++ b/users/untrusted.rst
@@ -58,6 +58,10 @@ becomes available.
         T2 -> U1 [label="Encrypted by T2"]
     }
 
+In the above scenarios, if the untrusted device U1 deletes a local copy of a
+file, this is not replicated directly to trusted hosts. Untrusted hosts might
+receive the notice, however.
+
 Similarly, it's fine to add "normal mode" synchronization between untrusted devices.
 
 .. graphviz::


calmh:
I don't think they can, really. Current implementation wise I suspect it will result in the untrusted side just being out of sync and wanting to download it again, and protocol wise it doesn't actually matter what the untrusted side says about the delete bit, I think...

calmh:
It's an effect of the whole encrypted file metadata being carried inside the untrusted metadata. That original metadata can't be affected by an untrusted device, so it can't set the delete bit inside it.

An untrusted device can indeed delete a file locally and _another_ _untrusted_ device would replicate that delete because it doesn't know better. A trusted device that can decrypt the "inner" information will ignore the "outer" delete bit, though.

calmh:
No, but to other untrusted devices.

users/untrusted.rst
```diff
@@ -58,6 +58,10 @@ becomes available.
         T2 -> U1 [label="Encrypted by T2"]
     }
 
+In the above scenarios, if the untrusted device U1 deletes a local copy of a
+file, this is not replicated into any of the others hosts. An untrusted host
+cannot remotely trigger a deletion of data present in other hosts.
```
bean5, line 63:
```suggestion
In the above scenarios, if the untrusted device U1 deletes a local copy of a
file, this is not replicated directly to trusted hosts. Untrusted hosts might
receive the notice, however.
```

rasa (approved):

users/untrusted.rst
```diff
@@ -58,6 +58,10 @@ becomes available.
         T2 -> U1 [label="Encrypted by T2"]
     }
 
+In the above scenarios, if the untrusted device U1 deletes a local copy of a
+file, this is not replicated directly to trusted hosts. Untrusted hosts might
+receive the notice, however.
```
calmh, line 63: I mean, they won't, because Syncthing won't send it. This whole thing is a hypothetical on a potential Evil Syncthing that does the wrong thing but speaks the same protocol. 🤷

bean5: I understand better now. Just the sentence is a good add.

calmh:
I don't really feel this clarifies the situation, to be honest.

PR #895, opened 2025-02-09
chore(contrib): move syncthing-android and SyncTrayzor to old list

### LINKED ISSUE #892: Remove or mark "syncthing-android" as deprecated at "Community Contributions"
With https://github.com/syncthing/syncthing-android now in archived mode and no longer being maintained,
it seems probably it should be removed from the listing for "[Community Contributions](https://docs.syncthing.net/users/contrib.html#android)",
or at least a note added to the description on that page that it is no longer being maintained.

Probably the description for "Syncthing-Fork" on that page should be rephrased then too,
as it being listed as "An alternative wrapper app" when it is the only alternative remaining for this essential use may be confusing,
as well as removing the part about "with extended functionality" since that was likely intended to compare  to the non-fork listing
and so now would be compared to nothing.

marbens-arch:
SyncTrayzor hasn't received any commits in 4 years, and syncthing-android is officially discontinued, so add them to the old list.

Fixes #892.

### DIFF
--- a/users/contrib.rst
+++ b/users/contrib.rst
@@ -20,24 +20,10 @@ Cross-platform
 Android
 ~~~~~~~
 
-- `syncthing-android <https://github.com/syncthing/syncthing-android>`_
-
-  A wrapper app for the Syncthing binary.
-
 - `Syncthing-Fork <https://github.com/catfriend1/syncthing-android>`_
 
   An alternative wrapper app for the Syncthing binary with extended
   functionality.
-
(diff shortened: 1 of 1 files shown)


acolomb (changes requested):
Not sure what to do with the remove `_contrib-windows` reference.  It is used as a target in `users/autostart.rst`.  You should clean that up somehow to avoid a broken link / build warning.

acolomb (approved):

PR #901, opened 2025-03-20
chore: updates for SQLite database

calmh:
wip


calmh:
Well the version incompatibility hasn't happened yet, but if it did it would be another major version 😅

PR #902, opened 2025-03-30
chore(config): remove discontinued STUN servers from default

marbens-arch:
We're removing discontinued STUN servers in syncthing/syncthing#10012, so we update the documentation accordingly.


calmh:
Let's just say "... gets expanded to a set of public STUN servers." and not have to duplicate the entire list here.

calmh (approved):

PR #904, opened 2025-04-03
chore(config): mention STUN SRV record

bt90:
Mention the SRV record in the docs and clarify preference.

### DIFF
--- a/users/config.rst
+++ b/users/config.rst
@@ -1255,8 +1255,7 @@ The ``options`` element contains all other global configuration options.
 
     Server to use for STUN, given as ip:port. The keyword ``default`` gets
     expanded to a set of public STUN servers, with preference given to those
-    hosted by the Syncthing project. The latter are dynamically resolved via
-    the SRV record ``_stun._udp.syncthing.net``
+    hosted by the Syncthing project.
 
     To configure multiple servers, you can either: repeat ``<stunServer>`` tags
     in the configuration file or enter several servers separated by commas in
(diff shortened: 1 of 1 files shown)


users/config.rst
```diff
@@ -1252,8 +1252,10 @@ The ``options`` element contains all other global configuration options.
 .. option:: options.stunServer
     :aliases: options.stunServers
 
-    Server to be used for STUN, given as ip:port. The keyword ``default`` gets
-    expanded to a set of public STUN servers.
+    Server to use for STUN, given as ip:port. The keyword ``default`` gets
+    expanded to a set of public STUN servers, with preference given to those
+    hosted by the Syncthing project. The latter are dynamically resolved via
+    the SRV record ``_stun._udp.syncthing.net``
```
calmh, line 1258:
```suggestion
    Server to use for STUN, given as ip:port. The keyword ``default`` gets
    expanded to a set of public STUN servers, with preference given to those
    hosted by the Syncthing project.
```

let's not even ...

bt90: The AI scrapers will find it anyway...

bt90: I can try to obfuscate the SRV entry if we *really* want to go the extra mile

marbens-arch (approved):

users/config.rst
calmh: No need to obfuscate, but someone maybe you mentioned Hyrum a while back, and I dont' see a need to at least _specify_ this is how we're going to be doing things. Let's just do something and let it that be that, be it srv records or a backend service returning a random stun address etc.

bt90:
@calmh also part of v1.29.6

PR #908, opened 2025-04-14
fix(specs): state that both A and B are in session mode (fixes #802)

### LINKED ISSUE #802: Incorrect image description for Session mode example
Source: https://docs.syncthing.net/specs/relay-v1.html#id1

The Example Exchange caption references permanent and temporary protocol mode instead of session mode. Additionally, I think a deliminator between Client A and Client B descriptions would increase clarity. I propose a change to:

Client A - Session mode sender; Client B - Session mode receiver

Must add the deliminator to other descriptions for consistency.

marbens-arch:
Permanent protocol and temporary protocol submodes don't exist in session mode.


calmh (commented):
I guess, if you say so 😅

calmh (approved):

PR #909, opened 2025-04-14
fix(relay): document RelayFull message (fixes #880)

### LINKED ISSUE #880: Relay Protocol misses documentation for RelayFull packet
### What happened?

At [the relay protocol v1 documentation](https://docs.syncthing.net/specs/relay-v1.html), there are only 7 kinds of packets (Type = 0..6).

However, there is a kind of RelayFull packet defined in the code:
https://github.com/syncthing/syncthing/blob/47f48faed7331b7ba4ad3d6775d5cffacf8931b5/lib/relay/protocol/packets.go#L23

marbens-arch:
The RelayFull message indicates that the relay has reached its limits, such as connection limits.


specs/relay-v1.rst
```diff
@@ -125,11 +126,12 @@ Session mode
 The first and only message the client sends in the session mode is the
 JoinSessionRequest message which contains the session key identifying which
 session you are trying to join. The relay responds with one of the following
-Response messages:
+messages:
 
 1. ResponseNotFound - Session key is invalid
 2. ResponseAlreadyConnected - Session is full (both sides already connected)
 3. ResponseSuccess - You have successfully joined the session
+4. RelayFull - Relay limits are too strict for you to be able to join the session. The relay immediately terminates the connection after sending this.
```
acolomb, line 134: Please wrap the line to limit its length.  I can't give you an exact rule for maximum number of columns per line, but try to deduce it from the surrounding text.  80 to 90 characters seems usual in the docs.

acolomb (changes requested):

specs/relay-v1.rst
marbens-arch: Done

acolomb (approved):
Haven't checked whether this actually matches the protocol definition.  I trust you did that correctly :-)

PR #911, opened 2025-04-16
chore(advanced): clarify multiple connections over relays

tomasz1986:
chore(advanced): clarify multiple connections over relays

Clarify that it is possible to establish two connections when each
device uses a different relay.

Ref: https://forum.syncthing.net/t/multiple-connections-established-over-a-relay-but-docs-claim-its-impossible/20975


tomasz1986:
Filing this PR, because I've been seeing these `1 + 1` connections over relays basically all the time.

calmh (approved):

PR #914, opened 2025-04-21
feat(faq): add information on setting network to private in Windows

tomasz1986:
feat(faq): add information on setting network to private in Windows

The issue of Windows devices being unable to connect directly with each
other comes up often on the forum. One of them main reasons for the
problem is Windows setting the default network profile to "Public"
instead of "Private", which prevents direct connections.

Instead of repeating the same instructions on how to change the network
profile to "Private" over and over, add a new entry to the FAQ, which we
can then use as a reference when needed.

### DIFF
--- a/users/faq-parts/troubleshooting.rst
+++ b/users/faq-parts/troubleshooting.rst
@@ -196,7 +196,7 @@ will be merged with files from remote devices.
 Also see the :ref:`marker FAQ <marker-faq>` for more information about the folder marker.
 
 Why do my Windows computers always connect through a relay?
---------------------------------------------------------------
+-----------------------------------------------------------
 
 When connecting to a new network, Windows by default sets its profile to
 "Public". While being more secure, this setting also prevents applications like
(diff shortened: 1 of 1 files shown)


users/faq-parts/troubleshooting.rst
```diff
@@ -193,4 +193,14 @@ at the same location. This causes syncthing to attempt an automatic re-creation
...
 
-Also see the :ref:`marker FAQ <marker-faq>` for more information about the folder marker.
\ No newline at end of file
+Also see the :ref:`marker FAQ <marker-faq>` for more information about the folder marker.
+
+Why do my Windows computers always connect through a relay?
+-----------------------------------------------------------
+
+When connecting to a new network, Windows by default sets its profile to
+"Public". While being more secure, this setting also prevents applications like
```
marbens-arch, line 202: It is still possible to get direct connections through WAN addresses, even when the network is set to "Public".

```suggestion
"Public". While being more secure, this setting also often prevents applications like
```

tomasz1986: I've changed it to "commonly prevents".

marbens-arch (approved):

acolomb (approved):

Issue #624, opened 2021-03-03
Add a Tuning section to the documentation


bt90:
There are some things to consider which might positively affect syncthings performance. We should probably collect and curate this information in the documentation.

e.g safe options like putting the database on SSD storage or even unsafe ones with a proper warning like turning off fsync for initial sync.

IMHO it's better to have it in the documentation with proper warnings in place. People will always look for ways to speed things up and maybe we can steer those users away from questionable blog posts created years ago and instead point to the documentation.

imsodin:
I moved this over to the docs, as that's what it targets.

There's a lot of things that make sense to document (and might already be documented), like progress updates. However I think unsafe tweaks should not be documented in a nice overview-like section as proposed: No amount of warning will stop people from still trying. That's why I'd not document fsync or other "not-to-be-named  options there  :) 

In general I believe in most cases knowing the knobs is a secondary problem: Most people never arrive at understanding what the limitation in their setup is. I'd consider a help article explaining how to diagnose typical problems more helpful.

tomasz1986:
I would say that this is a good idea, but we should clearly distinguish between tuning purely for performance, and also tuning for reducing the resource usage, e.g. when using Syncthing on low-end hardware, etc.

calmh:
There are a few knobs we (or I) quite often recommend adjusting upwards or downwards for either very-small or very-large setups. I think we could have at least a couple of halfway general sections "to reduce resource usage at the cost of performance / features, lower and disable these things" and "to improve performance in large setups, increase these things and/or rate limit these other things".

bt90:
I think the following sections might make sense:

* General
  * e.g keep database on SSD
* Resource usage dependent (knobs to turn up/down depending on the hardware)
  * e.g number of connections
* Large folders
  * things to consider if share size is in the TB scale

marbens-arch:
Resolved in #789. This should be closed now.

PR #915, opened 2025-04-21
feat(tuning): add section for LAN-only (fixes #834)

### LINKED ISSUE #834: Tuning article for avoiding any Internet/non-local chatter
There should be one, with the necessary knobs mentioned and tradeoffs discussed.

marbens-arch:
This adds a section to the tuning article for restricting network traffic to your local network only.


users/tuning.rst
```diff
@@ -214,3 +214,35 @@ Other things:
...
+
+You are paranoid about security and/or privacy, and want to keep Syncthing's
+communication constrained to your local network only.
+
+For restricting sync traffic to only LAN:
+
+- :opt:`relaysEnabled`
+    Set to false to disable relaying sync traffic through servers other than
+    your sync partners. When relays are used, those that you connect to can
+    see your encrypted data, device ID, and public IP address.
```
acolomb, line 229: This sounds a bit misleading to me.  "See your encrypted data" could give a false impression that the relay can actually see the data *although* it is encrypted.  The truth is, it can only see some gibberish byte stream *because* the data stays encrypted while passing through the relay.

```suggestion
    see your device ID and public IP address, but not the synchronized data,
    as that is encrypted.
```

Feel free to improve the wording, it's just a quick draft.

acolomb:
I think you're missing some pieces where Internet facing communications take place.  We already have somewhat of an overview here:

https://docs.syncthing.net/users/security.html

And it's probably good to align with this FAQ entry, or link from it to your new section:

https://docs.syncthing.net/users/faq.html#why-does-syncthing-connect-to-this-unknown-suspicious-address

marbens-arch:
Wasn't thinking about usage reporting or crash reporting when I was writing this. But those should be included too.

acolomb:
... and version upgrade checks as well?

acolomb (approved):

acolomb:
I removed the reference to #624, as I don't see how this covers the whole scope requested there.

PR #919, opened 2025-04-27
feat(users/ignoring): add escaping on Windows

rasa:
Ref syncthing/syncthing#10058


acolomb (approved):

marbens-arch:
Should the base be main or v2?

acolomb:
I don't think we have a v2 branch here?  The main PR is against v2, so I created a milestone to collect changes here which should not be merged before v2 becomes the "current" release.

marbens-arch:
> I don't think we have a v2 branch here? The main PR is against v2, so I created a milestone to collect changes here which should not be merged before v2 becomes the "current" release.

https://github.com/syncthing/docs/tree/jb/v2

acolomb:
Which is the basis for #901, but unrelated to this change here.  That's why I added both to the v2 milestone for easily filtering out when the time has come.

rasa:
@acolomb Should we add a
```
.. versionadded:: 2.0.0
```
say, in the footer of this note?

rasa:
> And indeed add a `.versionadded` for 2.0.0 into the appropriate place 🙏

[Done](https://github.com/syncthing/docs/pull/919/commits/29bf4e537ddcc3aed3280730a563aac43490548c).

calmh (approved):

PR #925, opened 2025-05-06
Update folder-caseSensitiveFS.rst

camoz:
Add Android to list of systems that are case-insensitive.

(The last part of the sentence that was removed was kind of redundant.)

### DIFF
--- a/advanced/folder-caseSensitiveFS.rst
+++ b/advanced/folder-caseSensitiveFS.rst
@@ -11,8 +11,9 @@ handling.
 With ``caseSensitiveFS`` set to  ``false`` (the default setting)
 Syncthing's case sensitivity safety checks are enabled. 
 Syncthing will then attempt to detect and prevent case-only file
-name collisions that can occur on case insensitive systems such as Windows
-and macOS, or other systems with case insensitive file systems.
+name collisions that can occur on case insensitive operating systems such as
+Windows, macOS, iOS, and Android, or other systems with case insensitive file
+systems.
 
 When set to ``true`` the extra safety checks for case insensitive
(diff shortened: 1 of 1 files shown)


calmh:
It's not really redundant as you may well run with case insensitive FAT on Linux etc.

camoz:
Yeah, that's why I wrote "kind of". The sentence talks about case(in)sensitive systems and case(in)sensitive filesystems, which makes it more complicated. I only removed the second part of the sentence, because with it, it became unnecessarily long and a bit weird to read, IMO. Feel free to suggest anything better or just dismiss this PR. I just thought it would be nice to have Android included there, since it's a popular platform for syncthing.

advanced/folder-caseSensitiveFS.rst
```diff
@@ -11,8 +11,8 @@ handling.
 With ``caseSensitiveFS`` set to  ``false`` (the default setting)
 Syncthing's case sensitivity safety checks are enabled. 
 Syncthing will then attempt to detect and prevent case-only file
-name collisions that can occur on case insensitive systems such as Windows
-and macOS, or other systems with case insensitive file systems.
+name collisions that can occur on case insensitive systems such as Windows,
+macOS, and Android.
```
tomasz1986, line 15: How about something like this?

```
name collisions that can occur on case insensitive operating systems such as
Windows, macOS, iOS, and Android, or other systems with case insensitive file
systems.
```

I've changed the first "systems" to "operating systems" and also added iOS to the mix. I think these are the big four.

camoz: I updated the PR to use your proposal.

PR #926, opened 2025-05-07
docs: add more context to Windows execution

rafaeloledo:
- ~~Creating tasks without the GUI~~
- Editing screenshots that mention `--no-console` (see #955 )
- ~~Workarounds on running without console pop-up (relates to #861 )~~


rafaeloledo:
@tomasz1986 
Thank you for the suggestions. Can you please see the new changes? 🤗

users/autostart.rst
```diff
@@ -271,6 +252,26 @@ by a sysadmin who knows enough to understand the security implications.
 #. Start the service via ``sc start syncthing`` in the Command Prompt.
 #. Connect to the Syncthing UI, enable HTTPS, and set a secure username and password.
 
+For Hard Users
```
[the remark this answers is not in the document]
rafaeloledo: No, i mean for users that take steps to optimize the workflow that, sometimes, can be complex configurations in the system. I'm not english native. Changing it to a more semantic word.

users/autostart.rst
```diff
@@ -252,6 +252,26 @@ by a sysadmin who knows enough to understand the security implications.
...
+^^^^^^^^^^^^^^^^^^
+
+It's possible to run Windows Task Scheduler CLI API with a command such as 
+``schtasks /create /sc ONLOGON /tn Syncthing /tr "<program-path> [--no-console --no-browser]"``.
+The operation requires elevated privileges. Preventing the pop-up console that hides after
+some delay is possible wrapping the executable with a ``PowerShell`` script.
+
+``script.ps1``::
+
+   Start-Process "<path to syncthing>.exe" -WindowStyle Hidden
```
rasa, line 265: Your schtasks example includes `--no-console --no-browser` but your 
Powershell example doesn't. Perhaps it should, so the two examples mirror each other?

Also, isn't `-File` required? So: `powershell -File <path to script>.ps1` ?

rasa (changes requested):

users/autostart.rst
rafaeloledo: >Windows Terminal and Default Console:
In modern Windows versions, particularly with Windows Terminal set as the default console application, the --no-console flag might not completely hide the console window. It might still appear and then minimize, or the flag might be ignored entirely.

https://forum.syncthing.net/t/win-10-and-11-updates-breaks-no-console/19142

I was trying the `--no-console` flag that only exists in the `Windows` version of the `syncthing` and faced some issues related.

Switching to the mentioned `legacy mode` may be discouraged, since `Windows Terminal` is being a default on all installations.

Wrapping withj a `ps1` script works even on `wt.exe` pop-ups.

`-File/-FilePath` is not mandatory.

```

Comment (bt90) [reply users/autostart.rst:265]: Might also try `conhost.exe --headless ...`

**Review by rafaeloledo — COMMENTED** [review]
Made some changes according to the suggestions. Please check #936. `conhost.exe` example may be unnecessary and duplication of that subject. @tomasz1986 will powershell example be sufficient? Both seems to be working, but i can't even find online docs for the additional command and both seems to work.

https://learn.microsoft.com/en-us/windows/console/definitions#console-host

User may be overhelmed similar to `vbsript` usage. Following a standardized approach with powershell may be the case and only mention other technologies when it not work IMHO. :hugs:

**Comment by tomasz1986** [comment]
Just my opinion, but after looking at the PR, and also the other PR from https://github.com/syncthing/docs/pull/936, my feeling is that we may be getting a bit too in depth here. This goal of this section is to provide the user with a few methods that would let them start Syncthing automatically on startup/boot, not to teach them how to use the command line 😉.

At the moment, the existing instructions can be summarised as follows:

1. The "run at system startup" using Task Scheduler method still works and doesn't require any changes.
2. The "run at user logon" using Task Scheduler method is outdated and doesn't work in Windows 10/11 with Windows Terminal installed and set as the default console. It needs to be updated (e.g. to utilise `conhost` or `powershell` to do the window hiding job).
3. The "run at user logon using the Startup folder method faces the same issue as the previous one. It also requires an update to actually hide the console window.
4. The "run as a service" method doesn't require any changes.

I'm not sure adding yet another entry just to explain how to use PowerShell with Task Scheduler is really the way to go. How about adding `powershell` (and possibly `conhost`) to the existing two methods that need to be updated anyway instead? The other https://github.com/syncthing/docs/pull/936 PR does it, but only as a note, while we actually need to update the current instructions themselves (including the screenshots).

File: users/autostart.rst (hunk 2 above)
Comment (rasa) [reply users/autostart.rst]: > `-File/-FilePath` is not mandatory.

I wasn't referring to `Start-Process` but to the powershell command itself:

`powershell -File "<path to script>.ps1"`

Comment (rafaeloledo) [reply users/autostart.rst]: > It still works without `-File`, but I'm not sure if it implicitly defaults to `-File script.ps1` or rather `-Command script.ps1` 🙃.

Unfortunately @rasa example is what happening in my system too.

image

So, the `-File` flag is mandatory in this case.

**Comment by rafaeloledo** [comment]
Waiting for the merge of #955 to make more changes here and not break anything.

File: users/autostart.rst (hunk 2 above)
Comment (rafaeloledo) [reply users/autostart.rst]: > > `-File/-FilePath` is not mandatory.
> 
> I wasn't referring to `Start-Process` but to the powershell command itself:
> 
> `powershell -File "<path to script>.ps1"`

image

I can only find docs offline about `powershell.exe`. But, as you can see, even if it is treated as optional, you'll need to use the `dot operator` to resolve file paths. Using `-File` directly can me more user friendly considering that the operator is less known by non programmers.

**Comment by rafaeloledo** [comment]
image

I've managed to make the service work as this:

`syncthingw.ps1`
```ps1
Start-Process -FilePath "C:\dev\scoop\apps\syncthing\current\syncthing.exe" -ArgumentList "--no-browser" -WindowStyle Hidden
```

`install-synthng-service.bat`or directly on terminal (as admin)
```cmd
schtasks /create /sc ONLOGON /tn Syncthing /tr "powershell -File C:\\dotfiles\\win32\\syncthingw.ps1"
```

The thing is @tomasz1986 and @rasa: running with `powershell.exe` and `Start-Process` doesn't persist a runtime process of powershell. Under the hood, it's using `conhost.exe`. So, it's not more lighweight. Poweshell is the modern solution that wraps the old execution of `conhost` if needed with the `.NET std lib`.

tomasz1986:
Yeah, so the question remains, is there any benefit of recommending PowerShell instead of `conhost.exe --headless`?

I think the `schtasks` part isn't really necessary, as we already describe how to create tasks via the GUI, so there is no need to complicate things further (and also clutter the Docs with Windows-specific stuff).

rafaeloledo:
I found one little breaking change using conhost together with `schtasks` that will break things. Current only working with `powershell`

The problem is:

```ps1
schtasks /create /sc ONLOGON /tn Syncthing /tr "powershell -File C:\\dotfiles\\win32\\syncthingw.ps1"
schtasks /create /sc ONLOGON /tn Syncthing /tr "conhost --headless C:\\scoop\\apps\\syncthing\\current\\syncthing.exe --no-browser"
```

Both tasks parses `powershell` and `conhost` as the program/script.

image

image

If you want to use `conhost` WITH args, you'll also need to wrap the entire command withing a `bat` or `cmd`.

`syncthingw.bat`
```cmd
conhost --headless "C:\dev\scoop\apps\syncthing\current\syncthing.exe" --no-browser
```

So, when i register the `conhost` script i must give like this:


Which will result in a different registering through CLI

schtasks /create /sc ONLOGON /tn Syncthing /tr "C:\\dotfiles\\win32\\syncthingw.bat"

Using inside `schtasks` and using `conhost --headless` will pop-up the terminal even with the `--headless` argument.

If i click on the terminal that is popped, it closes. A visual annoying thing.

This visual glitch is not happening using `powershell.exe` interpreter.

rafaeloledo:
https://github.com/user-attachments/assets/1c4bae24-eb78-41ef-9651-92b775d1bc71

Shablone:
From my experience, using PowerShell has the potential to create more headaches. I've run into restrictive execution policies, which can be really tricky.

I think the conhost solution is reasonable and pragmatic since it works without additional hurdles for now. In the long run, instead of having another workaround, it might be ideal if Syncthing itself had a proper native background mode with tray functionality or. That would make these console workarounds unnecessary.

rafaeloledo:
Following @Shablone point, the service can only run with `ExecutionPolicy` changes as shown below

https://github.com/user-attachments/assets/d3055c95-3789-400e-95ba-28e8985a481a

The user must run 

```ps1
Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
```

Or other permissive policy similar to `scoop` to be able to use `powershell` in my example.

rafaeloledo:
> I'm sorry for changing the subject: We could also address the root cause directly.
> 
> The console flashes because Syncthing is built as a console application. We can create a separate build (e.g., `windows-daemon`) compiled as a Windows GUI application. This tells the OS not to allocate a console, eliminating the flash without any external tools.
> 
> I've tested this and it works. The change is minimal: conditionally add the `-H=windowsgui` linker flag for that specific build [in `build.go` around line 903](https://github.com/syncthing/syncthing/blob/da7a75a8239222b61085a6d70a13d357abbc65d1/build.go#L903).
> 
> This would be a separate build to avoid breaking console functionality for the main binary. It's a clean, native solution for users who only run Syncthing as a service.
> 
> For windows we actually could ship both console and no-console by providing two files e.g.: syncthing.exe and syncthing-d.exe

Good idea

PR #933, opened 2025-06-01
Update config.rst to warn about https://github.com/syncthing/syncthing/issues/10109

leobard:
Documenting the underlying problem of https://github.com/syncthing/syncthing/issues/10109 right where I would have expected it


calmh:
I don't think the config reference is a good place for this. Also, it concerns several other options.

PR #935, opened 2025-06-13
chore(config): mention that only the default marker is created

imsodin:


users/config.rst
```diff
@@ -531,8 +531,9 @@ The following child elements may exist:
 
 .. option:: folder.markerName
 
-    Name of a directory or file in the folder root to be used as
-    :ref:`marker-faq`. Default is ``.stfolder``.
+    Name of a directory or file in the folder root to be used as a marker - see
+    :ref:`marker FAQ <marker-faq>` for it's purpose.
```
marbens-arch, line 535:
```suggestion
    :ref:`marker FAQ <marker-faq>` for its purpose.
```

users/config.rst
```diff
@@ -531,8 +531,9 @@ The following child elements may exist:
 
 .. option:: folder.markerName
 
-    Name of a directory or file in the folder root to be used as
-    :ref:`marker-faq`. Default is ``.stfolder``.
+    Name of a directory or file in the folder root to be used as a marker - see
+    :ref:`marker FAQ <marker-faq>` for its purpose.
+    A marker directory is only created by syncthing for the default ``.stfolder``, not otherwise.
```
calmh, line 536:
```suggestion
    A marker directory is only created by Syncthing for the default ``.stfolder``, not otherwise.
```

calmh (approved):

PR #937, opened 2025-06-20
feat(contrib): mention Synctrain

pixelspark:


tomasz1986:
I think once should be enough 😉. Maybe just add a new "macOS and iOS" category?

pixelspark:
Agreed, maybe like this? (Feel free to comment on the wording as well, I think it would be good to make clear that Synctrain does not expose all Syncthing features but has a friendly/native UI, whereas syncthing-macos exposes all of Syncthing features and uses the normal web UI).

users/contrib.rst
```diff
@@ -33,14 +33,19 @@ We currently don't have any actively updated Windows-specific GUI wrappers here.
...
+~~~~~~~~~~~
 
 - `syncthing-macos <https://github.com/syncthing/syncthing-macos>`_
 
   syncthing-macos is a native macOS Syncthing tray application bundle.
   It hosts and wraps Syncthing, making it behave more like a native macOS application and less like a command-line utility with a web browser interface.
 
+- `Synctrain <https://github.com/pixelspark/sushitrain>`_
+
+  Synctrain is an open source, native app for iOS, iPadOS and macOS that combines Syncthing with a file-centered UI.
```
tomasz1986, line 46:
```suggestion
  Synctrain is an open source, native app for iOS, iPadOS and macOS that combines Syncthing with a file-centered UI with focus on on-demand file synchronization.
```

calmh (approved):

PR #941, opened 2025-07-07
chore(autostart): use double dashes in Windows create shortcut screenshot (fixes #940)

### LINKED ISSUE #940: fix flags in windows screenshots
These two flags in [./users/autostart.rst](https://github.com/syncthing/docs/blob/d55a43269b6c2c1f1846874a68e39a535d5277c0/users/autostart.rst#windows) only have one dash `-` in the screenshot.

Image

kaitwillows:
fixes #940 

not sure if it matters all that much, but mine is the windows 11 ui.

also sorry idk how anything works do i just merge onto main?


tomasz1986 (approved):
There are more screenshots like that, so I think it would be nice to update all of them using the same version of the OS for consistency, but this one looks almost the same as the original one. The others will still need to be updated later when Syncthing v2 is officially released.

kaitwillows:
> Just my opinion, but I personally think the screenshots are useful, especially considering that this section is targeted at new users with limited tech skills. Otherwise, we will be left with a wall of text, and no matter how well written, text-only instructions are daunting for the non-tech-savvy.

no i agree, this guide made my life a little easier.

i was able to google "ooh how do i make a program start at startup," but i didnt know what binary to include, and what flags. following this guide was nice because i was confident i wasn't doing anything wrong :).

PR #952, opened 2025-08-14
Update device-numconnections.rst for st 2.0 as it allows 3 connections by default

realeatham:


tomasz1986:
Thanks for the update! Please just change the line breaks. The original text was written with up to 76 characters per line, so after the update, it should be something like:

```
``numConnections`` is a device setting that affects connection handling. A
zero value means to use the Syncthing default. As of version 2.0 the
default is to use three connections: one for index metadata and two for
data exchange. This may change in the future.
```

realeatham:
> Frankly, let's just remove the sentence entirely. I know I wrote this from the beginning, but really I'm convinced it's a bad idea to codify default values in the documentation as they tend to change...

Ok, i've done that.

calmh (approved):

PR #955, opened 2025-08-31
fix(autostart): remove --no-console from Windows instructions

tomasz1986:
fix(autostart): remove --no-console from Windows instructions

With the advent of Windows Terminal and it being the default shell both
in newer releases of Windows 10, and in Windows 11, the `--no-console`
command line switch no longer works. For this reason, remove it from the
commands listed in the autostart instructions. When it comes to the Task
Scheduler part, it has never actually been needed anyway, and in the
Startup folder instructions, it can be worked around by using the native
`conhost.exe --headless` method.

### DIFF
--- a/users/autostart.rst
+++ b/users/autostart.rst
@@ -90,7 +90,7 @@ https://docs.microsoft.com/windows/win32/taskschd.
    #. Click "New...".
    #. Enter the path to ``syncthing.exe`` in "Program/script:" (for
       example ``C:\syncthing\syncthing.exe``).
-   #. Enter ``--no-console --no-browser`` in "Add arguments (optional):"
+   #. Enter ``--no-browser`` in "Add arguments (optional):"
    #. Click "OK".
 
    |Windows Task Scheduler Actions Screenshot|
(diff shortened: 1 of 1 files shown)


tomasz1986:
Please note that the screenshots will also need to updated, but I'm leaving that for later (as I would like to spin up a Windows 11 VM in order to take them in a clean environment, but I won't be able to do it in a short time).

tomasz1986:
Thanks for the screenshots, I think in this specific PR including the ones that have `--no-console` removed is appropriate. How would you prefer to be recognised as the author of them? If I download and push them to the PR myself, the whole PR will still end up being authored by myself only.

calmh:
You can just add a co-authored-by when merging.

rafaeloledo:
> You can just add a co-authored-by when merging.

This is sufficient.

rafaeloledo:
The images that not get merged are out of scope of this PR, right? Sorry for that. :hugs:

239:
> With the advent of Windows Terminal and it being the default shell both
> in newer releases of Windows 10, and in Windows 11, the `--no-console`
> command line switch no longer works.

It still works with `Run: Minimized`

Minimized

PR #956, opened 2025-08-31
chore(autostart): add note about allowing access in Windows Firewall

tomasz1986:
chore(autostart): add note about allowing access in Windows Firewall

Currently, we always recommend to "Run whether user is logged on or not"
when using the Task Scheduler to start Syncthing automatically. The
problem is that this makes Syncthing run completely hidden in background
which prevents Windows Firewall from prompting the user to allow
Syncthing to access the network. Therefore, add a note explaining that
the user should run Syncthing interactively at least once, which will
allow the Windows Firewall pop-up to be displayed on the screen.

### DIFF
--- a/users/autostart.rst
+++ b/users/autostart.rst
@@ -105,9 +105,9 @@ https://docs.microsoft.com/windows/win32/taskschd.
 
 .. note::
 
-   You should start Syncthing interactively (i.e. without relying on Task
+   You should start Syncthing interactively (i.e. without relying on the Task
    Scheduler) at least once in order to trigger a Windows Firewall pop-up,
-   which will ask you you to allow Syncthing to access your network. This is
+   which will ask you to allow Syncthing to access your network. This is
    required, e.g. if you want Syncthing to establish direct connections inside
    the given network, and also if you intend to access the web GUI remotely.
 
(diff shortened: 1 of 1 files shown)


users/autostart.rst
```diff
@@ -103,6 +103,14 @@ https://docs.microsoft.com/windows/win32/taskschd.
 
    |Windows Task Scheduler Settings Screenshot|
 
+.. note::
+
+   You should start Syncthing interactively (i.e. without relying on Task
+   Scheduler) at least once in order to trigger a Windows Firewall pop-up,
+   which will ask you you to allow Syncthing to access your network. This is
+   required, e.g. if you want Syncthing to establish direct connections inside
+   the given network, and also if you intend to access the web GUI remotely.
```
imsodin, line 112:
```suggestion
   You should start Syncthing interactively (i.e. without relying on the Task
   Scheduler) at least once in order to trigger a Windows Firewall pop-up,
   which will ask you to allow Syncthing to access your network. This is
   required, e.g. if you want Syncthing to establish direct connections inside
   the given network, and also if you intend to access the web GUI remotely.
```

imsodin (commented):
Just two nits, otherwise lgtm

Also if I am already nitpicking: Maybe chore instead of feat? :P

imsodin (approved):

PR #957, opened 2025-08-31
fix(syncthing): add note about --no-console requiring Command Prompt

tomasz1986:
fix(syncthing): add note about --no-console requiring Command Prompt

Clarify that the `--no-console` option only works when the Command
Prompt is set to be the default Windows shell. On the other hand, the
option does nothing when the default shell is set to Windows Terminal.

### DIFF
--- a/users/syncthing.rst
+++ b/users/syncthing.rst
@@ -174,7 +174,8 @@ Serve options
 
 .. cmdoption:: --no-console
 
-    Hide the console window. (On Windows only)
+    Hide the console window. (On Windows, and only if the OS default shell is
+    set to Command Prompt)
 
 .. cmdoption:: --no-port-probing
 


calmh:
Unsure, unclear if the scope adds much value anyway. We might just skip them entirely for the docs repo?

tomasz1986:
> Unsure, unclear if the scope adds much value anyway. We might just skip them entirely for the docs repo?

Do you mean like in the old days, just without capitalisation? 😀

`users/syncthing: add note about --no-console requiring Command Prompt`

tomasz1986:
Ah, I see. I'm not really sure about that then, as most of the commits do seem to have their scope clearly stated. I personally find it useful when skimming through the commit list.

PR #958, opened 2025-09-01
fix(rest): update `/rest/system/connections` docs to match status quo

marbens-arch:
ref syncthing/syncthing#3771
ref syncthing/syncthing#5737
ref syncthing/syncthing#8918


rasa (approved):
Mathes https://github.com/syncthing/syncthing/blob/e41d6b9c1e94cf503e1fb4667e355f489283b442/lib/connections/structs.go#L60

PR #975, opened 2025-11-09
new screenshot of NSSM arguments, that correspond with text description

kratochviljan:


rafaeloledo:
If you want, @tomasz1986, it's possible to merge this commit in the #955. The difference is that the #975 is adding `--no-restart`. It's changing the original text. Since i've changed other images, my image can be replaced with the one including this another flag if desired.

One thing to note is: `--no-restart` is being mentioned before the image, indeed.

image

PR #992, opened 2026-02-27
Update inotify docs

da5nsy:
Following https://forum.syncthing.net/t/how-do-i-increase-the-inotify-limit-to-get-my-filesystem-watcher-to-work/26189


da5nsy:
I'm not sure how I feel about the inclusion of the link to a SO answer. Very happy for that to be replaced if someone knows a better tool/link.

PR #999, opened 2026-03-26
docs(stignore): add pattern ordering note and common pitfalls section

hrubymar10:
Consider merging this. It took me hours to figure out why my ignore patterns weren't working, and I was only able to understand the issue after analysing the actual source code. It's definitely up to you whether you want to accept this — I just thought it could help some future people who hit the same wall. And yes, both the text in this PR and the documentation changes are AI-generated, but believe me there is a human with intent behind it who verified everything before opening this PR.

## Summary

- Add a note in the Patterns section explaining that `!/dir` auto-expands to also unignore `!/dir/**`, and that this affects pattern ordering
- Add a new "Common Pitfalls" section with a practical `.git` example showing both incorrect and correct pattern ordering
- Cross-reference from the note to the new section

## Motivation

Users commonly write `.stignore` patterns in an intuitive but incorrect order — broad whitelist first (`!/mydir`), then specific exclusions after (`/mydir/**/.git/**`). This fails silently because Syncthing uses first-match-wins and auto-expands directory unignore patterns. The current documentation doesn't warn about this ordering requirement.

This is especially common when syncing home directories or project folders while trying to exclude `.git` internals, `node_modules`, etc.

## Changes

1. **Note after first-match-wins explanation** — alerts users early that `!/dir` creates an implicit `!/dir/**` and links to the pitfalls section
2. **Common Pitfalls section** — shows the wrong ordering, explains why it fails, shows the correct ordering, and walks through how each path is matched

### DIFF
--- a/users/ignoring.rst
+++ b/users/ignoring.rst
@@ -34,6 +34,16 @@ Patterns
 The ``.stignore`` file contains a list of file or path patterns. The
 *first* pattern that matches will decide the fate of a given file.
 
+.. note::
+
+   When you write a pattern like ``!/mydir``, Syncthing automatically
+   also unignores everything inside that directory (as if you also wrote
+   ``!/mydir/**``). This auto-expansion is placed immediately after the
+   original pattern. Keep this in mind when ordering patterns — if you
+   need to ignore specific paths *inside* an unignored directory, those
+   ignore rules must appear **before** the ``!/mydir`` line. See
+   `Common Pitfalls`_ below for a detailed example.
+
 -  Regular file names match themselves, i.e. the pattern ``foo`` matches
    the files ``foo``, ``subdir/foo`` as well as any directory named
    ``foo``. Spaces are treated as regular characters, except for leading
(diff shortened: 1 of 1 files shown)


tomasz1986:
Just a comment, but `**` only actually does anything differently when it's put inside the path, e.g. `dir1/**/dir2`. Otherwise, you should just use `*` in any other place.

What I mean is that `dir/**` and `dir/*` (and even `dir/`) all do the same thing, so there is no point in using `**` instead of just `*` in this case.

calmh:
I think this writeup over-complicates and confuses, honestly, by explaining implementation details like expanded patterns and whatnot. There are two principles that need to be followed & understood:

1. The first pattern to match wins
2. Matching a directory matches anything underneath it

Point number two is kind of obvious really, it would be impossible for the ignore pattern `foo` to make sense and not also match everything under `foo/...`.

hrubymar10:
Thanks for the feedback! That makes sense - I agree the auto-expansion / implementation detail angle overcomplicates it.

I could strip it down to just a short "Common Pitfalls" example that only uses the two principles you mentioned (first match wins, directory match includes contents) without any implementation details. Would that be useful, or do you feel the existing docs cover it well enough?

Totally fine to close this if it's not needed - no big deal either way.

calmh:
I'm skeptical to showing intentionally broken examples as they tend to get picked up mentally as patterns even if it says they're incorrect. If the "first pattern wins" aspect somehow can/needs to be emphasized more strongly, I'm fine with that.

hrubymar10:
Fair enough — thanks for the feedback! Closing this.

PR #1002, opened 2026-04-14
Add VBS script method for Syncthing background launch

mevlat:
Added instructions for setting up Syncthing background startup using the VBS script and configuring the task scheduler without running the terminal.

### DIFF
--- a/users/autostart.rst
+++ b/users/autostart.rst
@@ -151,6 +151,32 @@ make selective use of them depending on your needs.
 
 .. _autostart-windows-startup:
 
+Setting up Syncthing Background Launch via VBS
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


tomasz1986:
Thanks for the contribution, but I don't really think we should add this. The reason is that VBScript is deprecated with plans to first disable and then remove it from Windows altogether (see https://techcommunity.microsoft.com/blog/windows-itpro-blog/vbscript-deprecation-timelines-and-next-steps/4148301).

mevlat:
Thanks, I didn't know about that. The reason I created this was because when I created a task in the task scheduler with the --no-console flag, the task was automatically interrupted.

tomasz1986:
As far as I'm aware, the original Task Scheduler method should still work just fine. Just in case, if you wanted to try finding out why the task failed on your end, please open a new topic on the forum about it.

mevlat:
I'm only guessing that syncthing perceives the --no-console flag as completing the task because the terminal is closing.

Edited: or I expect something else. because I expect the terminal application to disappear from the system menu at the bottom.

PR #1005, opened 2026-04-26
feat(config): block indexing

calmh:


users/config.rst
```diff
@@ -619,6 +619,17 @@ The following child elements may exist:
     devices when this is set to ``true``. See
     :doc:`/advanced/folder-send-xattrs` for more information.
 
+.. option:: folder.blockIndexing
+
+    .. versionadded:: 2.1.0
+
+    Whether to use block indexing for the folder; defaults to ``true`` for
+    send-only and receive-only folders. Block indexing makes it possible for
```
ManiacDC, line 627:
```suggestion
    send-receive and receive-only folders. Block indexing makes it possible for
```

ManiacDC (commented):
Code is send-receive

PR #1013, opened 2026-06-07
chore: clarify database resetting options

acolomb:
The `--reset-deltas` and `--reset-database` options were removed in Syncthing v2.  Fix leftover references and in the process, document the current `syncthing debug ...` subcommand group, where `reset-database` now lives.

See individual commit messages for details.  The last one I'm not quite sure about.  I resurrected the removed text from the `--reset-database` option docs, but couldn't verify that the corresponding API endpoint actually does recreate `.stfolder` markers.  Maybe this behavior was also removed in v2?  I haven't dug deep enough to be sure.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -92,7 +92,7 @@ version = release.partition('-')[0]
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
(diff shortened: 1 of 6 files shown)


calmh (approved):
noice

PR #1026, opened 2026-08-03
docs: add Zenith Hosting deploy option

odpay:
hey, i work on Zenith Hosting, a startup doing one-click hosting for open source apps. we added Syncthing to our marketplace, so this PR adds a short entry with a "Deploy with Zenith" button to the community contributions page (links to https://zenith.hosting/host/syncthing), right next to the Cloudron one.

to be clear about what it actually is: a managed, always-online syncthing device you share folders with like any other device, not a replacement for the p2p model. useful if you want a peer that stays up when your laptop is closed.

we share a cut of revenue with the projects we host. happy to explain if you're curious: hello@zenith.hosting


calmh:
Sweet, no thanks.

PR #1027, opened 2026-08-05
Add note on Global Discovery and Relaying in documentation

bburns:
When I tried SyncThing a while back, MalwareBytes threw up a warning about a bad ip address and hackers, so I abandoned it out of caution. So mentioning this upfront might be helpful - could make it shorter.


bburns (commented):
reworded the note - does that look correct?

calmh:
Why are we talking about relay nodes after starting the paragraph about discovery servers? The two are distinct and not related

bburns:
Okay, reworded it so they are described separately - does that look okay?

tomasz1986:
To be honest, not really. I don't think I've seen any reported case of antivirus software having problems with the discovery server. It is usually only relays that may trigger those warnings (as already mentioned in the FAQ).

bburns:
okay, maybe just a small note with a link to the faq entry?

Issue #1028, opened 2026-08-14
Client Key/Certificate Algorithm update


DoctorD90:
As per PR https://github.com/syncthing/syncthing/pull/10162, the paragraph in doc should be updated about the usage of the ED Curve (`ED25519`).

Maybe a note about the impossibility to use the `ED448` to due Go current limit (https://github.com/syncthing/syncthing/issues/10854#issuecomment-5280884320), would be nice. What do you think?

https://github.com/syncthing/docs/blob/56d35a660de1fb027346351f311260be9e6fe7f3/dev/device-ids.rst?plain=1#L11-L66

girlfangs:
This confused me for a bit when setting up Syncthing, the documentation said that the client keys are based on `ECDSA`, so I wanted to manually replace them with `ED25519` keys using `openssl`  (peace of mind, etc).

So while checking the format of the keys already present on the configuration folder, I saw that that `cert.pem` already was a `ED25519`  certificate.

While most users won't really care about this, "power users" that do check the cryptography will probably be concerned that `ECDSA` is used and will either try to change it (and be confused because the documentation is outdated) or they will just consider some other software for whatever they wanna do.

PR #1029, opened 2026-08-15
Update dev/device-ids to ed25519 (default since v2.0.0)

girlfangs:
Syncthing uses ed25519 since v2.0.0, but this was not updated in the docs. Solves https://github.com/syncthing/docs/issues/1028

I only updated the one important paragraph, but maybe it's better to explain what algos can actually be used and update the example output of `openssl x509` with an actual certificate generated in the latest version?


dev/device-ids.rst
```diff
@@ -14,12 +14,12 @@ Keys
 To understand device IDs we need to look at the underlying mechanisms. At first
 startup, Syncthing will create a public/private keypair.
 
-Currently this is a 384 bit ECDSA key (3072 bit RSA prior to v0.12.5,
-which is what is used as an example in this article). The keys are saved in
-the form of the private key (``key.pem``) and a self signed certificate
-(``cert.pem``). The self signing part doesn't actually add any security or
-functionality as far as Syncthing is concerned but it enables the use of the
-keys in a standard TLS exchange.
+Currently this is a 256 bit ED25519 key (384 bit ECDSA prior to v2.0.0 and
```
marbens-arch: Nit: the D should be lowercase, as "Ed" is taken from the last name ["Edwards"](https://en.wikipedia.org/wiki/Harold_Edwards_(mathematician)), and the D isn't part of an acronym.

```suggestion
Currently this is a 256 bit Ed25519 key (384 bit ECDSA prior to v2.0.0 and
```

marbens-arch (approved):

